PR #28, opened 2017-04-15
Add election explainers

### LINKED ISSUE #14: Add election explainers
Based on
https://docs.google.com/document/d/1vkhRWIjsL83wWvniECWLEn1RpSxkImorZTCUERnwn0Q/edit
for consumption in WhoCanIVoteFor and WhereDoIVote.

chris48s:
My initial thought on this was that we should be able to assemble explanation text based on facts about the election stored in the data model. e.g:

* Single/Dual-tier authority
* Country (England/Wales/Scotland)
* Voting System
* Franchise
* Local mayor/Combined Authority mayor

.. but I quickly realised that doing this well is going to involve a large number of permutations which will be difficult to debug and relies on a certain amount of data that is currently not in the model (and maybe doesn't necessarily need to be there).

So approach number two is:

* Add a table for storing explanations: `id`/`description`/`text` (markdown formatted)
* Each election has an `explanation_id` which is a nullable FK to `explanations`
* We can define any number of explanations (via `/admin`) and attach them to any applicable elections
* This allows us to leave some as `null` if we want to and makes it easy to define custom explainers for particular areas (e.g: Isles of Scilly, Doncaster, North Tyneside, others that may pop up in future)
* If there are certain elections we don't get a chance to write explainers for (e.g: by-elections), we can just leave it as `null`

Hopefully this is a suitable implementation of that plan.

Closes #14


chris48s:
bit of an odd one..
The travis build on my fork passes: https://travis-ci.org/chris48s/EveryElection/builds/222367440
..but it fails on this repo: https://travis-ci.org/DemocracyClub/EveryElection/builds/222367448

The issue is that the new dependency isn't being pulled in during the `pip install`. Note that `django-markdown-app` gets installed [here](https://travis-ci.org/chris48s/EveryElection/builds/222367440#L191) but not [here](https://travis-ci.org/DemocracyClub/EveryElection/builds/222367448#L168)

I guess there's some funny business going on with the [dependency caching in travis.yml](https://github.com/DemocracyClub/EveryElection/blob/master/.travis.yml#L7-L13). I'd imagine the build will pass if you log into travis, clear the cache and restart the build but I don't have access to do that myself on this repo :/

chris48s:
yay :)

chris48s:
rebased onto master to resolve conflict

PR #29, opened 2017-04-15
Fix footer block

### LINKED ISSUE #18: Fix footer block
* Pull in FontAwesome so the icons display
* Change github link to https://github.com/DemocracyClub/EveryElection

chris48s:
Closes #18

### DIFF
--- a/every_election/templates/base.html
+++ b/every_election/templates/base.html
@@ -80,16 +80,10 @@
       </p>
       <p>Democracy Club is a UK based Community Interest Company that builds
         the digital infrastructure needed for a 21st century democracy</p>
-      <p class="inline-list social">
-        <a href="https://www.facebook.com/democracyclub/">
-            <i class="fa fa-facebook-square" aria-hidden="true" style="color: white"></i>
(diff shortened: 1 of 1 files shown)


every_election/templates/base.html
```diff
@@ -98,30 +98,16 @@
 
 
 {% block extra_javascript %}
-{% if debug %}
```
symroe: This silly dance is needed at the moment, as we need to use `{% javascript 'scripts' %}` in debug mode, not `{% static "js/scripts.js" %}` as they're not built.

The solution is to shove the script in the [`extra_js`](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/settings/base.py#L107) list in settings.

symroe:
Left a cod comment that may be useful in general, but really though, I think we should just drop FA as a requirement and use text links – the icons add very little for a huge download!

chris48s:
Done with text links.
I did have some questions/comments about `fallback.js` and icons and stuff but somehow I think it can wait ;)

PR #31, opened 2017-04-24
Catch invalid postcodes

### LINKED ISSUE #21: Fail more gracefully on bad postcodes
e.g: https://elections.democracyclub.org.uk/api/elections/?postcode=iv251hu

This should throw an error response (probably `400 Bad Request`) rather than an unhandled exception.

This change will need some coordination so that other code which relies on EE (e.g: https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_finder/helpers.py#L132 ) can handle the error codes sensibly.

symroe:
Fixes #21


chris48s:
This probably catches most things (and achieves the main objective of stopping the `KeyError` notifications).

Couple of other suggestions:

* Try a postcode for Isle of Man or Channel Islands e.g: JE3 7DW (Jersey). Does this handle it cleanly?
* There are various other bits of handling code for other edge case mapit responses at https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/data_finder/helpers.py#L59 which _might_ be worth having a quick look at.

symroe:
On `JE3 7DW`, remember this will try our ONSPD S3 cache first. That returns

```json
{
"wgs84_lon": 0.0,
"wgs84_lat": 99.999999
}
```

A valid result, but with no areas matching. This returns

```json
{
    "count": 0,
    "next": null,
    "previous": null,
    "results": []
}
```
in our API.

On the MapIt responses, I might change this PR to raise if the response isn't 200 actually.

BRB

symroe:
Force pushed to check for a 200 status code from MapIt.

symroe:
Sorry for the odd commits - came in when resolving the migration merge conflicts

chris48s:
Migrations applied cleanly on my DB, so should be good

Did you resolve the merge conflicts via GitHub's web interface by any chance?

symroe:
I don't think so, no, I did it on my local copy.

PR #32, opened 2017-04-24
Waiting on #31: DC mapit postcode lookup

symroe:
Spotted while working on #31, we should try a couple of different MapIt installs seeing as we have them.

### DIFF
--- a/every_election/apps/elections/query_helpers.py
+++ b/every_election/apps/elections/query_helpers.py
@@ -16,7 +16,7 @@ class BasePostcodeLookup:
         raise NotImplementedError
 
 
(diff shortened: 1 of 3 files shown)


symroe:
Also, based off #31, so that need merging first.

chris48s:
Will have a look once its rebased onto master. That should sort the `multiple leaf nodes in the migration graph` failures

chris48s:
If you're in the mood for feedback, ideally I would probably define `BaseMaPitPostcodeLookup` as an abstract class and declare `mapit_base` as an abstract property.

If you're trying to do a million other things, just slap it in there and we can worry about it another day :)

symroe:
I choose the blue pill!

PR #34, opened 2017-04-27
remove 'one-click DoS our own server' feature

chris48s:
If you've got a bunch of geographies imported on your dev copy, pull this down and give it a test to make sure it actually fixes the problem. I don't have sufficient test data on my local install to notice a performance difference.


symroe:
LOL

PR #35, opened 2017-04-27
Python 3.5 and 3.6 are out

hugovk:


chris48s:
> Python 3.5 and 3.6 are out

Indeed they are, but you need to update the build matrix in tox as well for this to work.

If we have a look at the python 3.6 build you've just done at https://travis-ci.org/DemocracyClub/EveryElection/jobs/226509892 , when we run `python --version` at the start of the build we're running 3.6: https://travis-ci.org/DemocracyClub/EveryElection/jobs/226509892#L159
..but then once tox sets up the build virtual env we're back to 3.4: https://travis-ci.org/DemocracyClub/EveryElection/jobs/226509892#L199

so at the moment your build is just running the tests 3 times under python 3.4

There's an example of how to set up tox to run multiple python versions in a similar environment here: https://github.com/DemocracyClub/UK-Polling-Stations/pull/574/commits/4b53dfb820e89765881686bc79a01efefc45801b which you can work from
.. although on that project we're only testing against 3.4 and 3.5 because not all of the packages we use were updated for 3.6 compatibility last time I checked.

Cheers

hugovk:
Hmm, that last commit didn't really work. Switching to either 3.6 or 3.5 is easy enough, and both pass, so here's PRs for those instead: #36 and #37.

Interestingly, the 3.6 builds took 6m13s / 6m43, and the 3.5 builds took 2m44s / 3m18s. Looks like the big difference comes from `installdeps` of the `tox` command.

But then rerunning with cache gave more reasonable times: 3.6 at 2m18s and 3.5 at 1m47s.

PR #37, opened 2017-04-27
Python 3.5 is out

hugovk:
Replaces and closes #35 and closes #36.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 sudo: false
 language: python
 python:
-  - '3.4'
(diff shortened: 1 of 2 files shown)


chris48s:
I think it would be better to run all 3 as a matrix as you tried to do in PR #35 - we definitely want to be running the tests on python 3.4 as the server this runs on is running python 3.4 and we don't plan any major server migrations this side of GE2017 :)

hugovk:
Looks like it was just missing `tox-travis`. New PR coming soon.

Issue #45, opened 2017-06-28
Create election id from radar


chris48s:
Once something is marked as 'this is an election', there should be a 'create an id' button which:

- opens the id creator
- links the created id to this record so that once the id is created, this election disappears from the 'radar'
(MVP)

stretch goals should be:

- auto populate the 'source' field with the relevant notice of polls or ALDC link - requires implementing 'source' field (see issue #5 ) but once that's done, should be easy
- auto-populate other info if possible (e.g: authority, date)  - that is much more difficult but would be nice :)

PR #59, opened 2017-07-21
Notice of Election docs

chris48s:
Notes:
* For testing/dev purposes, I've set up a `notice-of-election-dev` bucket. Add `S3_DATA_BUCKET = 'notice-of-election-dev'` to your `local.py` so you don't upload junk to the prod bucket.
* Boto is currently expecting to find keys in a `/etc/boto.cfg` or `~/.boto` file formatted like

```
[everyelection]
aws_access_key_id = xxxx
aws_secret_access_key = xxxx
```

I'm not sure how to test it with `I_AM`/role-based access. Maybe we can have a look at this next week and work out how to deploy it.

Future work:
* Do something with the `source` field: I've added a DB field for it, but we're not actually doing anything with it yet.
* At the moment we can only attach a Notice of Election doc to an election after an ID has been created. It would be useful to be able to specify this in the wizard (non-trivial as we can create many elections at once so might need to specify many Notice of Election docs)
* Auto-populate Notice of Election field if creating from `SnoopedElection` record (deal with this under issue #45 )

I think this is is probably enough for a MVP we can iterate but let me know if you think anything else needs to be done here.

Refs #5


chris48s:
Note: this won't always archive in the nicest way as it assumes notice of election is a PDF, whereas sometimes this may be published as HTML. e.g: https://www.glasgow.gov.uk/index.aspx?articleid=21425

In YNR we deal with SOPNs that are not PDFs by converting the web page to a PDF and archiving that. It is probably worth copying that approach as an improvement, but given this represents a step forward from where we are now and will work as intended in the vast majority of cases I'm going to suggest raising that as an issue for future improvement.

symroe:
Hard to leave a line comment for this, but I think we'd be better off converting `notice_local` to a `FileField` and using an existing S3 storage class from [django-storages](http://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html). We might not want to set the `DEFAULT_FILE_STORAGE` (we might!), but we can set the storage class on the FileField directly by doing:

``` python
notice = models.FileField(upload_to='notices_of_election', storage='storages.backends.s3boto3.S3Boto3Storage')
```

This will save having to maintain the upload code ourselves, and also allow for changing the storage in future if we wanted to.

chris48s:
* Refactored to use `django-storages` as requested.

* Set `AWS_STORAGE_BUCKET_NAME = 'notice-of-election-dev'` in `local.py` to use the dev bucket.

* The `django-storages` package doesn't support configuration via a `/etc/boto.cfg` or `~/.boto` file out of the box. Only django settings or env vars. You'll have to also set

```
AWS_ACCESS_KEY_ID = 'foo'
AWS_SECRET_ACCESS_KEY = 'bar'
```

in your `local.py`. It may be possible to write something into the config file using `boto3` to pick up config from elsewhere and populate the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` vars if you don't want to deploy it like that, but maybe more effort than its worth..

* Source field is no longer added in this PR.

symroe:
Nice one :)

Issue #60, opened 2017-07-24
Link duplicate SnoopedElections to another Election or SnoopedElection


chris48s:
When we mark a `SnoopedElection` record as being a duplicate, it would be nice to be able to specify what it is a duplicate **of**. At the point we set the 'duplicate' status, we should be able to select either an existing election id or another `SnoopedElection` record.

PR #62, opened 2017-07-25
Create IDs from Radar

### LINKED ISSUE #5: Add "source" field for elections
We should track the sources of election IDs somehow.

The main thing to think about is how to add this to the ID creator form.

### LINKED ISSUE #45: Create election id from radar

chris48s:
Not ready for merge yet (see below - there are several other things that need to happen before I can finish it) but this adds a "Create an ID for this election" link to entries in the Election Radar list which opens the wizard and pre-populates with a source, Notice of Election URL and date where possible.

Notes:
* I've picked b2319d6 from PR #59 so that I've got the relevant fields in the DB to work with, but I haven't picked the other commits from that branch so that these 2 PRs can be reviewed seperately as far as possible.
* Implementing [this](https://github.com/chris48s/EveryElection/blob/issue45/every_election/apps/elections/views/id_creator.py#L261-L262) is blocked on the review/merge of PR #59, but it should only be a couple of lines that need adding here once I rebase onto it.
* Making the [Source Form](https://github.com/chris48s/EveryElection/blob/issue45/every_election/apps/elections/forms.py#L18-L22) display correctly is blocked on a fix for https://github.com/DemocracyClub/dc_base_theme/issues/15 .
* Meta-data (source, Notice of Election, `snooped_election_id`) is only attached to individual ballot ids (not group ids) by design because groups might already exist or cover multiple sources/Notice of Election docs.

Closes #5
Closes #45


chris48s:
Hmm.. just realised that EE is (like WhereDIV) on the 2016 theme, so being blocked on https://github.com/DemocracyClub/dc_base_theme/issues/15 is slightly more inconvenient than I had first envisaged. I'd like to start using this without having to do a full theme upgrade first.

I think we can probably get away with fixing that bug on a branch of the theme and pinning the [dependency](https://github.com/DemocracyClub/EveryElection/blob/master/requirements/base.txt#L12) to that commit as an interim fix pending a full theme update.

chris48s:
PR now open at https://github.com/DemocracyClub/dc_base_theme/pull/16 to fix theme issue

chris48s:
Rebased and reworked in the context of changes made under PR #59. Theme fix incorporated.
This should be finished now good for review.

symroe:
> This FK relationship was the wrong way round because one SnoopedElection can tell us about multiple election_ids.

But can't we have more than one SnoopedElection per Election too? For example, I'd expect to have one SnoopedElection object per scraper, if they're working properly.

Maybe we should just ignore that case. We could have a M2M relationship, but then we wouldn't know what the connections represented

chris48s:
So you're thinking of the case where we've picked up the same election via both the `ALDC` scraper and the `CustomSearch:NoticeOfElectionPDF` (and potentially N more scrapers in future) and we want to attach the IDs of both `SnoopedElection` records to the same election id?

We could make it a many-to-many (although it then becomes more difficult to populate). We could also tackle it like [this](https://github.com/DemocracyClub/EveryElection/issues/60) - so one scraped record becomes the 'canonical' source that we create an if from and then others telling us about the same election are duplicates. We could then link duplicates either to that `SnoopedElection` record or the id we created from it..

symroe:
Yeah, duplicates is a nicer way of doing it. In that case, merging this 👍

Issue #49, opened 2017-07-17
Allow EE to be used as a library


chris48s:

symroe:
I think "library" might be the wrong term here, at least as I see it.

The way I used EE on WCIVF was to set up a local copy of the _web service_ on each server that WCIVF used (ie, baked in to the image that spun up new instances).

This meant that I could communicate with EE over HTTP and all I needed to do was to set the `EE_BASE_URL` setting.

When we use the word "library" I start thinking about it being a reusable Django app that can be pip installed and used by the project directly in the Django/python interface.

This is a nice idea, but might be the wrong way to go:

1. using HTTP means switching from a central system at quiet times to a local system is much easier.
2. Related, setting up a dev copy of a project would be simpler, as there would be no need to sync live data from EE down to the dev copy
3. Dog fooding the public API is A Good Thing for everyone. If we use a python API then we're unlikely to have a great HTTP API (or at least, we won't care about having one).

I think the work here should be making it easier for EE to sync itself with upstream data in some way.

The considerations for that should be speed of import. For WCIVF I used a binary DB dump from postgresql. Importing this took seconds, making it quick enough to do on instance boot, meaning that we got fresh data every time an instance came up (the dumps were synced to S3 every hour, and the instance requested the latest).

chris48s:
OK, so having chewed it over a bit, lets try and work out what we're actually going to do here..

Regardless of whether we're calling EE as a python library or as a local JSON API, the actual hard bit of this problem is how we deal with syncing data and a lot of what we need to do changes the apps that consume EE and the deploy processes supporting them, rather than EE itself.

* We already have some scripts supporting the process of running a local EE instance (for local HTTP/JSON access). These should be generalised into some kind of re-usable ansible module that can be pulled into `polling_deploy` and `who_deploy` (using ansible galaxy, or a git submodule, or something).

* At the moment, we take a DB dump from EE every day on a schedule. It would be useful if we had a way to check if anything has changed in the database since the last export and only create a new dump if something has changed since the last export.

* Currently WhoCIVF doesn't do scheduled imports - it just boots and:
  * Sets itself to dirty/unhealthy
  * Runs an import
  * Sets itself to clean/healthy
  If we want to update EE, we've got to cycle the server. Is it reasonable to just do this or do we need scheduled imports?

* WhereDIV also needs to implement the clean/dirty flag

[...]

PR #73, opened 2017-08-01
Allow a 'client' EE to sync Elections app models from a 'server' EE

chris48s:
Having chewed over all sorts of complicated approaches to this I think the correct solution at this stage is to go with a really simple/dumb solution.. basically I've made an API endpoint which generates a fixture and serves it up and a corresponding management command which calls it and loads the fixture.

Using fixtures, `dumpdata` and `loaddata` in this way has several useful advantages:

1. `loaddata` handles some hairy stuff around FKs and statement order for us. There are problems around what order you run `DELETE`/`INSERT` statements in to ensure you aren't violating FK constraints which we just don't have to worry about with this approach.
2. Everything happens in a transaction - either the whole thing loads, or it fails and you're back where you started.
3. The "client" site/api stays responsive while the sync is happening, so we don;t have to wrry about clean/dirty flags in the "consumer" apps (unless our app already uses them for other reasons).

Obviously the tradeoff is that this isn't really re-usable by anyone else. You have to have the right magic numbers, but as we've noted we're not currently in a place where someone else can actually generate their own different magic numbers at the moment so lets focus on something that works for us rather than optimising for users we don't have yet.

Performance-wise, loading ~5,600 objects from a fixture was taking about 5 seconds locally. Its probably reasonable to expect we'll have just under double that by May next year, but obviously that number is going to keep growing with time. My inclination at this point is that for the medium-term future (the next few years?), we can probably get away with this process just being 'dumb' and syncing the entire elections app without having to be smart enough to check if it has changed first. The elections data represents less than 1% of the total database size, so getting to a point where we can build the orgs/geographies into an image and just sync the elections app nightly solves the big part of the problem. The rest is just diminishing returns at this stage.

We do need to ensure we rebuild the images when there are boundary changes + so on but I think we just deal with that in the same way as a new AddressBase release or whatever..

The next bits of this that need some work will be:

* Add an EE orgs/geography "layer" to the WhereDIV/WhoCIVF `packer` builds
* Find a way to "randomise" the syncing on the "client" end - I'm not particularly worried about this from the perspective of the clients staying responsive now as they seem to continue happily serving requests while the `loaddata` is happening, but `/sync` is a reasonably heavyweight endpoint and I'd rather not have 10 servers all request it at once..

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -91,6 +91,7 @@ ENV/
 
 every_election/settings/local.py
 every_election/settings/production.py
+every_election/data/
 
 # collect static output directory
 every_election/static/
--- a/every_election/apps/elections/management/commands/sync_elections.py
+++ b/every_election/apps/elections/management/commands/sync_elections.py
@@ -1,0 +1,14 @@
+import tempfile
+import urllib.request
+from django.conf import settings
+from django.core.management import call_command
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        url = settings.UPSTREAM_SYNC_URL
+        with tempfile.NamedTemporaryFile(suffix='.json') as tmp:
+            urllib.request.urlretrieve(url, tmp.name)
+            call_command('loaddata', tmp.name)
--- a/every_election/apps/elections/urls.py
+++ b/every_election/apps/elections/urls.py
@@ -1,7 +1,14 @@
 from django.conf.urls import url
 
-from .views import (ElectionTypesView, AllElectionsView,
-                    IDCreatorWizard, FORMS, CONDITION_DICT, SingleElection)
+from .views import (
+    ElectionTypesView,
+    AllElectionsView,
+    IDCreatorWizard,
+    FORMS,
+    CONDITION_DICT,
+    SingleElection,
+)
+from elections.views.sync import get_election_fixture
 
 
 id_creator_wizard = IDCreatorWizard.as_view(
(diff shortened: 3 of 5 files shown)


chris48s:
Refs #49

symroe:
Based on our conversation this morning, might it be worth changing the `get_election_fixture` view to read the dumped data from a file and then calling `dumpdata` to that file on a cron? That might get around the performance question quite quickly.

If we want to be fancy we can always call `dumpdata` after creating or updating (including soft deleting) a new election, essentially cache busting the file.

symroe:
Of course this view can run the command if the file isn't found, meaning we can bust the cache by deleting the file too.

chris48s:
Why the need for cache-busting? If I run `manage.py dumpdata > file.json`, won't it just overwrite `file.json` regardless of whether there is new content to write or not?

symroe:
Right but we only need to run that command if a) the file isn't there and b) the data is updated. If the view only runs the command when the file doesn't exist then we can clear the cache by either updating an election in the elections table, or manually by deleting the file.

chris48s:
I think it would be better to just write the file on a schedule with cron and serve it up on that endpoint. If the endpoint can kick off writing to the file there's potential for strange concurrency bugs if we're going to have multiple clients requesting that endpoint at the same time (which is the idea).

I'm not sure its really necessary to worry about making sure we don't run `dumpdata` on a schedule if we don't need to. It runs quickly enough that I was happy to have a HTTP endpoint trigger it. I don't particularly want lots of requests causing that process to run multiple times at concurrently, but running it once an hour or something isn't enough overhead to bother trying to 'optimise' it.

symroe:
Yep, I see that, fair enough!

chris48s:
Updated as requested. If the file isn't there, the 'server' will throw a 500 `FileNotFoundError` and the client will just die before it tries to load any data from it. I'll leave you to schedule `python manage.py dumpdata elections > /foo/bar/elections.json` in cron

PR #74, opened 2017-08-02
Add scraper for LibDemNewbies - resolves #72

### LINKED ISSUE #72: Add more sources to scrapers
http://libdemnewbies.org.uk/elections/forthcoming-by-elections/

mrwilson:


mrwilson:
Thanks Chris.

- I've updated the code to check for 'cause' as in the other commit.
- To get around the unique field, I've synthesised a URL using unique information in the hash since LibDemNewbies' election page is just plain text and no external/internal links

chris48s:
Cheers. We'll get it deployed next week

PR #77, opened 2017-08-11
add mailing list signup form

chris48s:
TODO:
* ~Blocked on https://github.com/DemocracyClub/Website/pull/77~

* Once deployed, we will need to generate an API key for it and then add `EMAIL_SIGNUP_API_KEY = 'f00ba2'` to `local.py`
  See https://github.com/DemocracyClub/dc_signup_form#configuration

* ~Once I've looked at implementing this on WhereDIV, I'll tag https://github.com/DemocracyClub/dc_signup_form with a `1.0` release and pin the dependency to it.~

### DIFF
--- a/every_election/settings/base.py
+++ b/every_election/settings/base.py
@@ -179,7 +179,7 @@ AWS_S3_REGION_NAME = 'eu-west-1'
 AWS_S3_FILE_OVERWRITE = True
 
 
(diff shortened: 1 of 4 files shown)


chris48s:
Good for review. There is an API key at https://democracyclub.org.uk/admin/signup_server/token/

Issue #82, opened 2017-08-21
Elections by postcode not working/elections with NULL geography


chris48s:
I am getting some unexpected results from the API. For example:

* If I call https://elections.democracyclub.org.uk/api/elections.json?postcode=SE193LG&future=1 then I would expect the response to tell me about `local.croydon.south-norwood.by.2017-09-07`
* If I call https://elections.democracyclub.org.uk/api/elections.json?postcode=NP78HH&future=1 then I would expect the response to tell me about `local.herefordshire.golden-valley-south.by.2017-09-07`
* If I call https://elections.democracyclub.org.uk/api/elections.json?postcode=LA12PQ&future=1 then I would expect the response to tell me about `local.lancaster.skerton-west.by.2017-09-07`

but in all those cases the response is

```
{
  "count": 0,
  "next": null,
  "previous": null,
  "results": []
}
```

I've synced a local copy and had a bit of a poke about. It seems that the problem is that there are a bunch of elections where the `geography_id` is `NULL`, including the ones I've specified above. I thought initially this was a problem with the id creator wizard, but having poked about there is a management command called `add_geography_to_elections` and running this locally fixes the problem. Is the only thing we need to do in order to fix this to schedule that to run with cron? (I checked the `crontab` and it isn't scheduled), or is there a reason this isn't running?

PR #83, opened 2017-08-22
populate geography when we save an election

### LINKED ISSUE #24: Assign geographies when adding a new election
See https://github.com/DemocracyClub/yournextrepresentative/issues/87

chris48s:
Slightly more than 4 lines, but not by much ;)

Once we deploy this I'll run `add_geography_to_elections` on prod and that should also sort out #82

Closes #24


symroe:
Would you mind adding a test for this case too? It should be fairly easy, and I think we need tests on anything that creates IDs, as the code is…ripe for improvement :)

PR #91, opened 2017-09-17
add full list of areas with directly elected mayor

### LINKED ISSUE #88: accurately list authorities with directly elected mayors
Tried to create an election ID for Mayor Hackney on 03/05/2018 but Hackney isn't listed as an available authority for a Mayoral election. There are also a variety of other areas missing. We should accurately list in the ID creator all areas which may have a mayoral election.
For example: http://www.londoncouncils.gov.uk/who-runs-london/london-elections/elected-mayors
also, I think Bristol has a directly elected Mayor and isn't listed, as another example.

Where can we source this from?

chris48s:
We definitely don't want to re-run the full `import_organisations` process on prod to apply this as it will fall over and break stuff in its current state. Suggest running

```
from organisations.importers import mayor_importer
mayor_importer()
```

from the django shell to apply this.

Closes #88


symroe:
Deployed and imported

Issue #70, opened 2017-07-31
Refactor IDMaker


chris48s:
The `IDMaker` class needs restructuring from 2 perspectives:

## General Code Structure:

`IDMaker` has 11 optional constructor parameters. This is a classic situation where we should be using the builder pattern. `IDMaker` should be refactored into an `ElectionBuilder` class which instantiates with the requied data and default values for everything else. It should then have `setProperty()`/`withProperty()` methods for all of the optional stuff (depending if we're going with a mutable or immutable implementation) and a `build()` method which returns an `Election` object. `ElectionBuilder` should be responsible for object creation only and decoupled from the concept of database persistence. `Election` objects should be responsible for their own DB persistence (or at least the calling code).

## Re-usability:

The logic in `IDMaker` also determines the schema for election ID generation. Our ambition is that DC election IDs should be _reproducable_ (i.e: we should define a method for generating ids, not become an issuer of ids) but at the moment our code for generating the IDs is tightly coupled both to a particular framework/ORM and the wider process of creating an election object. This makes it difficult to point to our code as a reference implementation and it is difficult for others to share.

To this end, we should also decouple the ID creation part of the process from the creation of an `Election` object (which is both coupled to a data model and includes additional meta-data not relevant to ID generation such as source, Notice of Election doc, etc). Once refactored, the `ElectionBuilder` should assemble the necessary data (such as organisation names) from the input data and our data model and hand that off to a separate object or function which only implements the logic necessary to build the ID string. That object/function is then much more suited to serve as a definition of how we create IDs.

chris48s:
Taking the objective in re-usability a bit further, what we should actually work towards here is a situation where anyone can just `pip install uk-election-ids` (or whatever) and be able to independently generate election ids in their own application which are reproducible and compatible with ids made by Every Election.

chris48s:
Big chunk of the 're-usability' stuff is now done under PR #97.

There is still refactoring work to do on `create_ids_for_each_ballot_paper()` and `IDMaker` to clean up that structure. When working on that, it is useful to consider:

1. If we need to create a bunch of ballot ids, we can do something like:

```python
>>> from uk_election_ids.election_ids import IdBuilder
>>> from datetime import date

>>> divisions = ["area1", "area2", "area3"]
>>> org_id = IdBuilder('local', date(2018, 5, 3)).with_organisation('test-org')
>>> [org_id.with_division(d).ballot_id for d in divisions]
[
[...]

Issue #48, opened 2017-07-17
Update reference definition


chris48s:
Update https://democracyclub.org.uk/projects/election-ids/reference/ to reflect current id creation rules

chris48s:
https://elections.democracyclub.org.uk/election_types is also wrong

Issue #87, opened 2017-09-05
Special cases for City of London Elections


chris48s:
Slightly edited comments from conversation on slack:

@sjorford wrote:
> According to http://democracy.cityoflondon.gov.uk/mgMemberIndex.aspx?bcr=1 "Each ward elects one Alderman and two or more Common Councilmen (collectively referred to as Members, equivalent to councillors), dependent on its population. There are currently 100 Common Councilmen and 25 Aldermen serving as elected representatives..."

@chris48s wrote:
> If I create <a href="https://elections.democracyclub.org.uk/elections/local.city-of-london.tower.by.2017-09-11/">local.city-of-london.tower.by.2017-09-11</a>, is that sufficiently unambiguous? I assume that even if there were both and Alderperson and Common Councilperson being elected for the same ward at the same time you'd essentially get have one poll with 2 ballots?
..but then I think the idea with the election ids spec is that each _ballot_ should have a distinct id..

@sjorford wrote:
> for the SP/NAW/GLA we have an intermediate level of IDs e.g. `gla.c` and `gla.a`
so do we need something like `local.city-of-london.cc.tower.by.2017-09-11` and `local.city-of-london.alder.tower.by.2017-09-11`?

I think we technically we should introduce sub-types to account for City of London oddities.

The other thing this raises is versioning the spec. For the moment it probably doesn't matter as it is only really us dogfooding the schema but once we have wider users, I think it becomes useful to be able to say that `local.city-of-london.tower.by.2017-09-11` was created under version 1.0 but would be `local.city-of-london.alder.tower.by.2017-09-11` under version 1.1 or something. Probably worth thinking about.. Even once we're accounting for all of the edge cases present in the current structure/legislation correctly there is still the possibility of electoral reform, changes to local government structure + so on..

GeoWill:
Just as a note of what I've done for the 2022 elections. 
I've created a new organisation for the aldermanic elections `city-of-london-alder' with an official identifier that is 'LDN-alder'. I then created a new divisionset and copied all the divisions over, setting the `seats_total` to 1.

```
from copy import deepcopy


# Make a new org
city_of_london = Organisation.objects.get(id=54)

city_of_london_alder = deepcopy(city_of_london)

city_of_london_alder.id = None
city_of_london_alder.official_identifier = 'LND-alder'
city_of_london_alder.slug = "city-of-london-alder"
city_of_london_alder.election_name = "City of London aldermanic election"
city_of_london_alder.save()

[...]

PR #97, opened 2017-10-25
Decouple ID Generation from DB Record creation (part 1)

### LINKED ISSUE #17: I get a "Whoops!" when trying to add an undated election
Was trying to add the Manchester Gorton by-election, everything went well until I hit 'Create ID' and then it said 'Whoops!'

I imagine it's something to do with not having the date?

chris48s:
Refs #48 and #70.

This PR pulls out the logic for assembling election ids from strings into an independent module. I am envisaging that at some point, the `uk_election_ids` app will split out from Every Election into its own repo/python module.

This does several useful things:
1. In order to work out how to correctly implement the validation rules and write all the unit tests, we've now specified the ID schema for `naw`, `sp`, `nia`, `parl`, `local`, `pcc`, `mayor`, `gla` (although a more human-readable version of this would also be handy).
2. This is one of the steps on the road to improving the structure of `IDMaker` and `create_ids_for_each_ballot_paper()`.
3. The `IdBuilder` class is now decoupled from EE, django and our database and can (in principle) be re-used by others. It also provides some nice notations/shortcuts for generating group ids and stuff. If we split this out into its own repo, this will allow others to generate reproducible ids (given appropriately slugged strings). A bit of example usage:

```python
>>> from uk_election_ids.election_ids import IdBuilder
>>> from datetime import date

>>> myid = IdBuilder('local', date(2018, 5, 3)).with_organisation('test-org').with_division('test-division')
>>> myid.ballot_id
'local.test-org.test-division.2018-05-03'
>>> myid.ids
['local.2018-05-03', 'local.test-org.2018-05-03', 'local.test-org.test-division.2018-05-03']

>>> myid = IdBuilder('gla', date(2018, 5, 3)).with_subtype('x')
ValueError: Allowed values for subtype are ('c', 'a')

>>> myid = IdBuilder('local', date(2018, 5, 3)).with_organisation('test-org')
>>> myid.election_group_id
'local.2018-05-03'
>>> myid.ballot_id
ValueError: election_type local must have a division in order to create a ballot id
```

I've also made minimal changes to the `IDMaker` class so that it is using the `IdBuilder` class, but I've not really done much to change the structure of that process at this point.

There are also a couple of tangential changes in this PR:

* Remove creation of subtype group ids
* Remove ability to create elections without a date (closes #17 as this is now irrelevant)

There are a number of bits of further work raised from this:
1. In order to make the `IdBuilder` class 'standalone' (i.e: independent of Every Election), there are certain rules/facts (e.g: `naw` has the subtypes `c` and `r`) which I've defined in the code but which also exist in our database. I'd like to avoid a situation where we have to update these in multiple places if something changes. Do you have any thoughts on the best way to deal with this?
2. I have neatly brushed over the issue of creating consistent area names with the brief aside "given appropriately slugged strings", but this is actually a huge issue encompassing 2 sub-issues:
  a) Different representations of the same area name
  b) All slugging algorithms are not the same
  We can neatly ignore these issues while we are the only users of this module, but I think these are both issues we need to address before we start encouraging others to use the code. I know writing your own slugging algorithm is generally a bad idea, but for B I am leaning towards either distributing a slugging algorithm (even if its just a copy-paste of django's) or depending on some framework agnostic third party package ( https://pypi.python.org/pypi?%3Aaction=search&term=slug&submit=search ) so that everyone is using a consistent algorithm, and I think we should do that sooner rather than later before our own IDs become hard for others to reproduce because they're coupled to a slug implementation that is part of a monolithic framework. I have less ideas about a sustainable approach to A.
3. There is still refactoring work to do on `create_ids_for_each_ballot_paper()` and `IDMaker`
4. I've started to think about versioning the id spec. At some point in the future, some assumption that we've made is going to stop being true, and we're going to have to update our id creation algorithm. At that point, in order to make reproducible ids, it becomes important to know what version of the algorithm was used to generate a particular id. I think the best solution to that would be to actually encode the spec version into the ID itself (e.g: `v1.local.2018-05-03`, `v2.3.local.2018-05-03`, etc). What are your thoughts on this?

### DIFF
--- a/every_election/apps/elections/constants.py
+++ b/every_election/apps/elections/constants.py
@@ -1,63 +1,16 @@
-ELECTION_TYPES = {
-    'parl': {
-        'name': "UK Parliament",
-        'subtypes': [],
-        'default_voting_system': 'FPTP',
-    },
-    'nia': {
-        'name': "Northern Ireland assembly",
-        'subtypes': [],
-        'default_voting_system': 'STV',
-    },
-    'naw': {
-        'name': "Welsh assembly",
-        'subtypes': [
-            {'name': 'Constituencies', 'election_subtype': 'c'},
-            {'name': 'Regions', 'election_subtype': 'r'},
-        ],
-        'default_voting_system': 'AMS',
-    },
-    'sp': {
-        'name': "Scottish parliament",
-        'subtypes': [
-            {'name': 'Constituencies', 'election_subtype': 'c'},
-            {'name': 'Regions', 'election_subtype': 'r'},
-        ],
-        'default_voting_system': 'AMS',
-    },
-    'gla': {
-        'name': "Greater London assembly",
-        'subtypes': [
-            {'name': 'Constituencies', 'election_subtype': 'c'},
-            {'name': 'Additional', 'election_subtype': 'a'},
-        ],
-        'default_voting_system': 'AMS',
-    },
-    'local': {
-        'name': "Local elections",
-        'subtypes': [],
-        'default_voting_system': 'FPTP',
-    },
-    'pcc': {
-        'name': "Police and crime commissioner",
-        'subtypes': [],
-        'default_voting_system': 'sv',
-    },
-    'mayor': {
-        'name': "Directly elected Mayor",
-        'subtypes': [],
-        'default_voting_system': 'sv',
-    },
-    'eu': {
-        'name': "European parliament (UK)",
-        'subtypes': [],
-        'default_voting_system': 'PR-CL',
-    },
-    'ref': {
-        'name': "Referendum",
-        'subtypes': [],
-        'default_voting_system': 'FPTP',
-    },
+from uk_election_ids.datapackage import ELECTION_TYPES
+
+
+# These types exist in EE but don't have defined behaviour in the IdBuilder
+ELECTION_TYPES['eu'] = {
+    'name': "European parliament (UK)",
+    'subtypes': [],
+    'default_voting_system': 'PR-CL',
+}
+ELECTION_TYPES['ref'] = {
+    'name': "Referendum",
+    'subtypes': [],
+    'default_voting_system': 'FPTP',
 }
 
 
--- a/every_election/apps/elections/forms.py
+++ b/every_election/apps/elections/forms.py
@@ -22,13 +22,6 @@ class ElectionSourceForm(forms.Form):
         label="Link to 'Notice of Election' Document")
 
 
-class ElectionDateKnownForm(forms.Form):
-    date_known = forms.ChoiceField(
-        choices=(('yes', 'Yes'), ('no', 'No')),
-        widget=forms.RadioSelect,
-        label="Do you know the date of the election?")
-
-
 class ElectionDateForm(forms.Form):
     date = dc_forms.DCDateField(help_text="The date that polls open")
 
--- a/every_election/apps/elections/templates/id_creator/id_creator_base.html
+++ b/every_election/apps/elections/templates/id_creator/id_creator_base.html
@@ -26,10 +26,7 @@
     {% block wizard_buttons %}
         <div class="form-group">
             <input type="submit" class="button" value="Submit" />
-
-            {% if wizard.steps.prev or wizard.steps.current == 'source' %}
             <p><a href="{% url "id_creator" %}?reset" class="button secondary">Start again</a></p>
-            {% endif %}
         </div>
 
     {% endblock wizard_buttons %}
--- a/every_election/apps/elections/tests/test_create_ids.py
+++ b/every_election/apps/elections/tests/test_create_ids.py
@@ -221,35 +221,6 @@ class TestCreateIds(BaseElectionCreatorMixIn, TestCase):
             subtypes=[naw_election_sub_type, ]
         )
 
-    def test_creates_tmp_id(self):
-        all_data =  self.base_data
-
-        all_data['date'] = None
-
-        all_data.update({self.make_div_id(): 'contested'})
-        expected_ids = [
-            'local.tmp-',
-            'local.test.tmp-',
-            'local.test.test-div.tmp-',
-        ]
-
-        self.create_ids(all_data)
-        assert Election.objects.count() == len(expected_ids)
-        for expected_id in expected_ids:
-            assert Election.objects.filter(
-                tmp_election_id__startswith=expected_id).exists()
-            assert Election.objects.get(
-                tmp_election_id__startswith=expected_id).election_id == None
-
-        all_data['date'] = self.date
-        self.create_ids(all_data)
-        assert Election.objects.count() == len(expected_ids)
-        for expected_id in expected_ids:
-            assert Election.objects.filter(
-                tmp_election_id__startswith=expected_id).exists()
-            assert Election.objects.get(
-                tmp_election_id__startswith=expected_id).election_id != None
-
     def test_election_with_organisation_geography(self):
         all_data = self.base_data
 
--- a/every_election/apps/elections/tests/test_create_ids.py
+++ b/every_election/apps/elections/tests/test_create_ids.py
@@ -212,7 +212,6 @@ class TestCreateIds(BaseElectionCreatorMixIn, TestCase):
 
         expected_ids = [
             'naw.'+self.date_str,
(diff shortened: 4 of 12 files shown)


chris48s:
[Every Election: New reference definition](https://trello.com/c/AIwLI4a3/927-every-election-new-reference-definition)

chris48s:
it seems I have written about versioning the spec before and then forgotten about it:
https://github.com/DemocracyClub/EveryElection/issues/87

symroe:
Quite a lot to think about / review here, so I expect I'm going to write a few comments in a fairly random order. Sorry.

## API interface

Notes on the actual interface for building IDs. 

### Validate that strings passes in will produce a valid ID:

For example, catch these cases:
```python
IdBuilder('local', date(2018, 5, 3)).with_organisation('tes!!!....t-org').with_division('test-d.i.v.is.ion')
>>> 'local.tes!!!....t-org.test-d.i.v.is.ion.2018-05-03'
```

I don't think we've defined everything that *can't* be in a section, but I think `.` should be validated for at the very least. Maybe anything that can't be passed through `slugify` too.

### Date strings
Not a major one, but I'd like to be able to do this:

```python
IdBuilder('local', "2018-03-03")
```

### Division lists

Do we want something like:

```python
def with_divisions(self, division_list):
    for division in division_list:
        self.with_division(division)
```

??

Or do you think a single division per object is the best way to go? I know we talked about this, can't remember what we decided was best. 

### Validating more

I know how to bundle the data is complex, but do you have ideas as to if/how we might validate that an organisation is able to hold an election of a type? Like Stroud District Council _can't_ be valid for type `mayor` (at the moment). Should that be in scope for this builder?

One thing I thought about was the builder knowing about a type of object being passed in. That is, `with_organisation()` could take a string _or_ an object. If the object contained a method (`valid_types` for example) then it could validate against them. This would work nicely if we wanted to pass in django models later on.

chris48s:
## Intro

> Quite a lot to think about / review here

Yep. I envisaged that this PR would be the start of a conversation rather than the end of one.

> I'm going to write a few comments in a fairly random order

Then allow me to respond to them in an equally spurious order :D


## Dates

> Not a major one, but I'd like to be able to do this: `IdBuilder('local', "2018-03-03")`

Yes - added in d657dfb.


## Creating Multiple IDs

> Do we want something like:
> [snip]
> Or do you think a single division per object is the best way to go?

I think that a single `IdBuilder` object should build at most one `ballot_id`. I think if you can pass it either a string or a list of divisions/orgs/subtypes, it complicates the implementation because we need to think about:
* Do we need 2 separate properties for `.ballot_id` and `.ballot_ids`?
* What does `.ballot_id` do if we passed in a list of divisions - should it throw an exception?
* If we only passed a string division, does `.ballot_id` return a string or a singleton list?

I'd propose keeping the `IdBuilder` itself focussed in this respect. If we want to create a list of division ids, we can do that easily enough like this:

```python
>>> from uk_election_ids.election_ids import IdBuilder
>>> from datetime import date

>>> divisions = ["area1", "area2", "area3"]
>>> org_id = IdBuilder('local', date(2018, 5, 3)).with_organisation('test-org')
>>> [org_id.with_division(d).ballot_id for d in divisions]
[
    'local.test-org.area1.2018-05-03',
    'local.test-org.area2.2018-05-03',
    'local.test-org.area3.2018-05-03'
]
```

..and then, we can still access the group ids, for example:

```python
>>> org_id.election_group_id
'local.2018-05-03'
```

Of course, there's nothing to prevent us from also adding some convenience functions to the `uk_election_ids` package to make that easier, but I'd suggest they sit outside of the `IdBuilder` class itself.


## Validation/Slugging

> ### Validate that strings passes in will produce a valid ID

Yeah, so I'm going to go back to that small phrase that is doing a lot of work in this PR: "given appropriately slugged strings". At the moment, `IdBuilder` expects the strings to have been slugged already. However, I think that it is valid to say slugging is in-scope for the `IdBuilder` class, given it is fundamentally concerned with building an id string. i.e: we could specify that in a 'finished' implementation the behaviour should be:

```python
org_id = IdBuilder('local', date(2018, 5, 3)).with_organisation('test-org')
>>> org_id.organisation_group_id
'local.test-org.2018-05-03'

org_id = IdBuilder('local', date(2018, 5, 3)).with_organisation('Test Org')
>>> org_id.organisation_group_id
'local.test-org.2018-05-03'
```

although obviously that's definitely not what it does at the moment. Rather than re-pasting a chunk from the PR description about slugging, see point 2 in the description here.

Your point about dots (etc) is interesting because obviously part of the job of a slugging algorithm is to remove and replace unwanted characters so I reckon the correct approach is that rather than 'validating' the input, this job should be handed off to a slugging algorithm which can just give us a valid string for our purposes given some input. However the fact that there are certain characters which are fundamentally invalid (which we might want special handling for) does maybe bolster the case for including our own slugging algorithm as part of the `uk_election_ids` package, rather than using something 'off-the-shelf'?

e.g: we're going to want to slug "St. Helen's" to `st-helens` not `st--helens` or `st-helen-s`.


## Django Models

> One thing I thought about was the builder knowing about a type of object being passed in. That is, `with_organisation()` could take a string *or* an object. If the object contained a method (`valid_types` for example) then it could validate against them. This would work nicely if we wanted to pass in django models later on.

In general I'm against the `IdBuilder` object knowing anything about Django or Django models. One of the objectives here is that the `uk_election_ids` app should be something that can be used in any python project, so I wouldn't want to couple it explicitly to Django models.

If we particularly wanted an implementation of `IdBuilder` in EE that could accept a Django model, I'd be inclined to do it by extending or wrapping `IdBuilder` specifically for EE, maybe something like:

```python
from uk_election_ids.election_ids import IdBuilder

class EveryElectionIdBuilder(IdBuilder):

    # ...

    def with_organisation(organisation):
        # additional validation or Django model specific stuff
        return self.with_organisation(organisation.slug)

    def with_division(division):
        # additional validation or Django model specific stuff
        return self.with_division(division.slug)

    # ...
```

Alternatively, if `with_organisation()` could (optionally) accept some more generic `Organisation` object, can you expand further on what the `valid_types` property would return/be used for?

At the moment, our [Organisation model](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/organisations/models.py#L5-L38) doesn't have this property.


## Other

Did you have any thoughts on points 1 or 4 in the description:
* How do we avoid encoding these rules in multiple places?
* Versioning the spec (#87 gives us a nice concrete example of how we might want to change the spec in the near future)

symroe:
> Alternatively, if with_organisation() could (optionally) accept some more generic Organisation object, can you expand further on what the valid_types property would return/be used for?
> At the moment, our Organisation model doesn't have this property.

Not called that, but is has a feature that would allow us to do this by doing `org.election_types.all()`.

Maybe django models was the wrong word for it, but accepting a python object as well as a string might be useful in general. No reason why that object needs to be a subclass of django's `models.Model` though.

But yeah, maybe this is too complex. Happy with your sub-class suggestion at the moment.

symroe:
> Did you have any thoughts on points 1 or 4 in the description:

1. Not sure about this. I think there's value in them being in the database still (for example, we're using them to produce the pages about each organisation. We could find a way to store this data in the `uk_election_ids` package, along with all other data taken from registers on the list of organisations – so a sort of data package. Then the EE _website_ could import from it as well as this package?

4. I'm not really keen on versioning being in the ID itself (seems to be mixing concepts), but I can't think of another way of doing it. We could just ignore it and provide a map between new and old IDs (and a redirect) that we've created? Really not sure how best to do it.

polling-bot-4000:
pointless post so that this thread goes back into Chris' Github Notifications and doesn't get forgotten

chris48s:
Based on chat this morning:

* 17360a1 builds the validation rules and populates the DB from some data defined in the `uk_election_ids` app to keep things DRY (this was actually a lot easier than I expected it to be as it was kind of half done).
* f2cce78 resolves the issues around slugging/making sure the strings we build the ids from contain valid characters.

The other stuff is useful to consider as we build this out further but out of scope for this PR. I've already raised an issue about versioning. I'll keep the other stuff in mind as I work out the next steps on this path..

Could do with another quick once over before it gets merged, but hopefully its good to go..

symroe:
Nice one :D

PR #108, opened 2017-11-22
Refactor IDMaker

chris48s:
Refs #70
Continued from PR #97

The is the next step in the task of refactoring the Election creation process. In the previous PR I split out the process of creating an Election ID string into the `IdBuilder` class. This PR splits the remains of `IDMaker` into `ElectionBuilder` (which we use to create an `Election` object and has a consistent interface with `IdBuilder`) and moves the database persistance code over to the `Election` model object itself. This also allows us to pass around and review an `Election` object which we can then persist if we're happy with it, rather than passing around and reviewing a proxy object which may persist an `Election`, which I think is conceptually nicer.

Most of the action is in 2760de6 - sorry this commit is a bit large but it was quite difficult to do this in multiple smaller stages :(

5f70c22 adds the validation checks we talked about in https://github.com/DemocracyClub/EveryElection/pull/97#issuecomment-342475799 . Conceptually I think this layer is the correct place for this validation to live. In the end, I structured it using composition rather than inheritance (i.e: an `ElectionBuilder` *has a* `IdBuilder` as opposed to `ElectionBuilder` *is a special* `IdBuilder`), which I think is more correct conceptually but it gets us to roughly the same outcome we discussed.

As with PR #97, I've tried to constrain the scope of this a bit so I've tried not to change `create_ids_for_each_ballot_paper()` too much in this PR. The final step of this piece of work will be to tidy that function up a bit. I think broadly the scope of it (as an interface between the form wizard and the `ElectionBuilder`) is ok, but there might now be scope to write some of it a bit more nicely.

Also tests - lots of tests

### DIFF
--- a/every_election/apps/elections/models.py
+++ b/every_election/apps/elections/models.py
@@ -3,7 +3,7 @@ import urllib.request
 
 from datetime import date, timedelta
 
-from django.db import models
+from django.db import models, transaction
 from django.core.files import File
 from django.core.urlresolvers import reverse
 from django.core.exceptions import ObjectDoesNotExist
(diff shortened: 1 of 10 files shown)


symroe:
Lovely!

Issue #41, opened 2017-05-10
Can't add Ettrick, Roxburgh and Berwickshire by-election


chris48s:
I am trying to add Scottish Parliament by-election for Ettrick, Roxburgh and Berwickshire ( https://www.scotborders.gov.uk/info/20057/elections_and_voting/199/elections ) but if I add an election for Scottish parliament, when I get to the stage where I pick an organisation there are no organisations to choose from. I can only mark everything as contested. This means I can't add a by-election:

![screenshot at 2017-05-10 11 44 09](https://cloud.githubusercontent.com/assets/6025893/25894993/ff431fa0-3575-11e7-8a07-3bdca39f6d8e.png)

Assume something has gone wonky with the geography import?

PR #110, opened 2017-12-07
Add division_election_sub_type when importing divisions

### LINKED ISSUE #103: Fix NAW geographies
We can't currently create an ID for Welsh Assembly elections, but we expect we will need to soon:
http://www.bbc.co.uk/news/uk-wales-politics-41904161

symroe:
Fixes #103

### DIFF
--- a/every_election/apps/organisations/management/commands/import_organisation_divisions.py
+++ b/every_election/apps/organisations/management/commands/import_organisation_divisions.py
@@ -66,7 +66,9 @@ class Command(BaseCommand):
         for generation_id, generation in self.mapit_generations.items():
             generation['uri'] = "{}/{}".format(url, generation_id)
 
-    def create_single_division(self, division_set, organisation, mapit_dict):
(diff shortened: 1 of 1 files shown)


symroe:
To add to this a little more:

The problem was that the divisions weren't appearing in the form wizard for either NAW subtypes.

This was because [the form filters by division subtype](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/elections/forms.py#L42-L45) but so sub-type was being added at import time.

So the wizard knew to show a subtype for because the election type has subtypes, but none of the divisions were being returned when they were being filtered.

This change adds the subtypes at import time. This also fixes the `sp` and `gla` types, with the (assumed) problem that GLA elections don't ever get `Additional` areas added. This is a different bug though!

chris48s:
Yeah - I suspected that whatever was causing this might be the same thing as #41, so this should close both, right?

chris48s:
Closes #41

Issue #102, opened 2017-11-06
Sort out various boundary problems


chris48s:
Oh the horror, _the horror_! Can open - worms everywhere :(

![image](https://user-images.githubusercontent.com/6025893/32452009-42df125a-c310-11e7-97a4-8549a46ac383.png)

There are a variety of boundary problems we need to sort out, of various levels of severity.

Before I created the election ids for the All-up elections in May 2018, I checked that there hadn't been any completed boundary reviews completed since ~April this year. However, I have missed 2 important categories of changes that we should have taken into account before creating the IDs:

* There are some areas which had boundary changes before the local elections in 2017 which I assumed had been imported, but actually haven't. Example:

    Redbridge:
    * https://elections.democracyclub.org.uk/organisations/RDB/ (boundaries we hold, outdated)
    * http://www.lgbce.org.uk/current-reviews/greater-london/redbridge (boundary review 28th April 2017)
    * https://elections.democracyclub.org.uk/elections/local.redbridge.2018-05-03/ (election ids created against wrong boundaries)

    I guess, thinking about it, that @symroe only imported the reviewed boundaries that were actually relevant to elections happening in May 2017, not everything that had been reviewed since the last boundary line release. I hadn't realised this though :(

* There are some areas which currently have a boundary review in progress which is expected to be completed in time for the boundaries to apply to the elections in May 2018 but are still under review and where I've already created election ids. Example:

    Leeds:
    * https://elections.democracyclub.org.uk/organisations/LDS/ (boundaries we hold, current but due to be superseded)
    * http://www.lgbce.org.uk/current-reviews/yorkshire-and-the-humber/west-yorkshire/leeds (boundary review, still in progress)
    * https://elections.democracyclub.org.uk/elections/local.leeds.2018-05-03/ (election ids created against wrong boundaries)

Both of these cases are problematic because we've already created IDs against the wrong boundaries.

In the first case (e.g: Redbridge), I assume that the correct boundaries are now available in published sources so I assume that to account for these we can probably import the latest BoundaryLine or something rather than faffing with the Boundary Commission stuff (or is it better to establish the workflow that boundaries always come from the Boundary Commission first?). Then we'll need to delete the Election objects we've already created, re-create them and deal with the fact that they've already propagated to YNR, somehow. It was probably reasonable (in principle) to create the IDs, but we should have updated the boundaries first.

[...]

PR #114, opened 2017-12-08
Division end dates

symroe:
This PR puts in place some of the things we need to address some parts of #102.

The main change:

1. `ElectionBuilder` will raise an exception if we don't have a division set the election date
2. The ID Creator wizard will show a message explaining this and won't allow creating election IDs for non-current division sets (this previously caused a 500)
3. General fixups around using dates properly etc.

Note that this doesn't add any end dates at all, so even after merging this PR it will still be possible to create incorrect IDs. 

End dates should be the subject of a different PR.


every_election/apps/elections/utils.py
```diff
@@ -5,6 +5,10 @@
 from uk_election_ids.election_ids import IdBuilder
 
 
+class NoDivisionsForDateError(ValueError):
+    pass
+
```
chris48s: I did think about doing this sort of thing myself. From a stylistic perspective, I decided not to define a custom exception for each error we've defined here. i.e: I've just had everything raise a generic `ValueError` rather than defining:

```python
class SubtypeInvalidForElectionError(ValueError):
    pass

class DivisionNotChildOfOrganisationError(ValueError):
    pass
```

..and so on. This is partly because the naming becomes very clumsy and partly because I think unless it is specifically necessary to write

```python
try:
    # do a thing
except NoDivisionsForDateError:
    # handle this error one way
except DivisionNotChildOfOrganisationError:
    # handle this error differently
```

in the calling code, it is fine to just throw a more generic exception rather than define lots of custom ones. Do you think there's a strong case for handling `NoDivisionsForDateError` differently here?

every_election/apps/elections/utils.py
```diff
@@ -77,6 +83,15 @@ def with_division(self, division):
                 (division, self.organisation)
             )
 
+        divisionset = division.divisionset
+
+        if divisionset.start_date > self.date:
```
chris48s: Should this check also be `if divisionset.start_date and divisionset.start_date > self.date`?  It looks like [start_date could be `null`](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/organisations/models.py#L44)

every_election/apps/elections/forms.py
```diff
@@ -85,10 +86,34 @@ def __init__(self, *args, **kwargs):
...
                 organisation=organisation,
                 start_date__lte=election_date
+                ).filter(
+                    models.Q(end_date__gte=election_date)
+                    | models.Q(end_date=None)
                 ).order_by('-start_date').first()
+            if not div_set:
+                # There is no active division set for this organisation
+                # on this date
+                no_divs_field = forms.CharField(
```
chris48s: Is a fake field the best way to deliver this error? For example a [message](https://docs.djangoproject.com/en/1.10/ref/contrib/messages/) might be a slightly more usual way to deliver this error (there are styles for `success`, `error`, etc in the `v0.2` theme - we use one for the mailing list signup form).

Is this just a workaround because the wizard thing falls over if you give it a form with no fields or something? If so, meh.. its not the end of the world

every_election/apps/elections/forms.py
```diff
@@ -85,10 +86,34 @@ def __init__(self, *args, **kwargs):
...
+                    ),
+                    required=False,
+                    label="""
+                        There is no active divisions for this organisation.
+                        This is normally because we know a boundary change
+                        is about to happen but it's not final yet.
+                        Please try again in future, or contact us if you think
+                        it's a mistake.
+                    """
+                )
```
chris48s: screenshot at 2017-12-08 14 20 11

Can we hide the 'mark all seats as contested' button in this situation. Also I can still press submit and it creates an election group id (but not an organisation id). That doesn't seem right. I reckon the only option here be 'start again', or if we do allow them to proceed it should create an organisation id as well as an election group id.

every_election/apps/elections/forms.py
symroe: I think there is another way of doing this, but consider the case of making more than one organisation election ID at a time: the page can get very long and a message would get lost at the top of it. We'd also need to have a valid form for the other IDs.

every_election/apps/elections/utils.py
symroe: Hmm I guess this is matter of style. I prefer custom exceptions for making testing and exception messages a little clearer. Also, note that because  `NoDivisionsForDateError` is a sub-class of `ValueError` then your exception handler can just catch `ValueError` and include `NoDivisionsForDateError` exceptions. 

Maybe we could have a generic `ElectionBuilderException` and other sub-classes as needed, all inheriting from `ValueError`?

every_election/apps/elections/forms.py
chris48s: Oh yes - good point. If you want to create ids for multiple orgs then you want to show pickers where possible and then show this error 'inline'.

every_election/apps/elections/utils.py
chris48s: Either way, lets pick one and be consistent. At the moment every error throws `ValueError` except this one specific case that throws `NoDivisionsForDateError` which is just a bit odd.

symroe: Cool, I've removed this in favor of `ValueError` for now, and we can think about more exceptions in future.

every_election/apps/elections/utils.py
symroe: You're right it can be `null`, although I think that should be changed. I've made #117 for this, and pushed a fixup to check if it's `None`

every_election/apps/elections/forms.py
chris48s: Thinking about in the context of your other comments, this it is less straightforward if you are creating ids for multiple orgs at once. If you're creating ids for 10 councils and 9 of them have valid divisions while one doesn't, you would probably want to allow the user to continue in that case which complicates this...

symroe: Looking in to this button more, there are a lot of things wrong with it, to the point where I'm not sure it's ever going to be useful. I might punt this to fixing under #116

PR #134, opened 2017-12-14
Fix id creation for gla.c election subtype

### LINKED ISSUE #112: ValueError creating election object for GLA Constituency
```
from elections.utils import ElectionBuilder
x = ElectionBuilder('gla', '2017-01-01')
x.with_subtype('c')
ValueError: 'c' is not a valid subtype for election type 'Greater London assembly'
```

oh noes!!

chris48s:
Closes #112


symroe:
Nice

PR #135, opened 2017-12-15
management command to attach end_dates to OrganisationDivisionSets

### LINKED ISSUE #117: Make DivisionSet start_date NOT NULL
As @chris48s raised in #114, `start_date` can currently be `null`. 

I suspect this shouldn't be the case, for the same reasons as laid out in #113.

chris48s:
Closes #117 
Refs #102

See comment at the top of the file for docs/usage. Next thing to do is assemble a CSV of area codes and end dates we can import with it...


every_election/apps/organisations/management/commands/update_end_dates.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+
+    def parse_csv(self, csv):
+        header = next(csv)
+        if self.EXPECTED_COLS != header:
+            raise ValueError(
+                "Unexpected header. Found %s expected %s" %\
+                (str(header), str(self.EXPECTED_COLS))
+            )
+        return [self.Record(*row) for row in csv]
```
symroe: Any reason for doing this rather than using `DictReader`?  Not that we care too much about memory, but I think `DictReader` will work with the stream rather than iterating over it all to memory. Also, it removes the need to worry about the named tuple. Only asking because I tend to use `DictReader` but I might be missing something.

every_election/apps/organisations/models.py
```diff
@@ -41,7 +41,7 @@ def format_geography_link(self):
 
 class OrganisationDivisionSet(models.Model):
     organisation = models.ForeignKey(Organisation, related_name='divisionset')
-    start_date = models.DateField(null=True)
+    start_date = models.DateField(null=False)
```
symroe: Does this (or the `unique_together` change) break anything with live data? I'd hope not…

chris48s: I tested by applying the migration on a local DB which contains a dump exported from prod a few days ago

symroe: Phew!

every_election/apps/organisations/management/commands/update_end_dates.py
chris48s: > Any reason for doing this rather than using `DictReader`?

Kind of, and kind of not.. it was a conscious decision to read into a list of named tuples and then implement `prepare_data()` as a pure function which accepts a list of named tuples and returns a different list of named tuples. In general I've been thinking recently about using immutable data structures more where possible as it eliminates certain classes of error relating to mutable state.

You are correct to point out that if we were reading in thousands and thousands of rows of data, the mutable state approach (storing as a dict and then having `prepare_data()` modify them in place) would be a much more efficient solution, although performance is probably not a serious consideration in this particular case given the number of records we are likely to deal with..

Its very much a stylistic thing though - happy to change it to a `DictReader` if that is more consistent with the rest of the codebase.

symroe: Nope, that seems like a good reason. Very happy to go with this :)

PR #145, opened 2017-12-30
Prevent overlapping start and end dates

### LINKED ISSUE #136: Prevent overlapping start and end dates
Where we've got multiple org/div sets for an org, the start/end dates shouldn't be able to overlap

i.e: it shouldn't be possible for the start date of the next generation to be before the end date of the previous generation

It would be helpful to enforce this as a DB constraint. It appears that it is possible to do this in postgres:
https://stackoverflow.com/questions/10616099/postgres-date-overlapping-constraint
but we won't be able to do it via Django's ORM: we'll have to add it as a custom migration.

Also we'll need to identify any cases where this error already exists and fix them before we can safely apply this migration.

When we do #15 we should apply the same rule there too.

chris48s:
This isn't yet safe to merge and deploy. The migrations can not be applied to the prod database yet because there are 3 integrity issues which we need to fix first:

62: West Sussex - 2 division sets with same `end_date`
499: House of Commons - 2 division sets with `NULL` `end_date`
256: Hastings Borough Council - overlapping start and end dates

Once we've fixed those issues, it will be possible to apply this to prevent problems like this from happening in future.

Closes #136


chris48s:
Note to self: once #162 is merged, rebase this on to it. The migrations will conflict because there is now a migration `0029` on the same model in this PR and #162. You'll need to rename the migration files and update the `dependencies` properties to resolve it.

chris48s:
rebased

PR #147, opened 2017-12-30
Fix 'Mark all seats as contested' buttons

### LINKED ISSUE #116: Duplicate 'mark all seats contested' button for elections with subtype
This isn't deployed to live yet, but using the latest code on master, I get this

screenshot at 2017-12-08 14 49 07

trying to create an election with subtypes. Must have missed this in the review of #110 :(

chris48s:
This PR replaces the multiple 'Mark all seats as contested' buttons attached to each header with a single 'Mark all seats as contested' button for the whole page. This fixes the problems in https://github.com/DemocracyClub/EveryElection/pull/114#discussion_r155789081 and #116 . In principle its a bit less flexible but in practice, I don't think the old buttons actually worked as intended anyway.

The only situation where this is still a bit odd is where we try to create an election and there are >0 orgs but 0 divisions available, but in general this is an improvement and fixes several problems.

I've intentionally done this by defining a django `SelectAllButtonField` form field rather than doing this in pure JS so that if we really want to put work into a more fine-grained approach, we can choose to conditionally create or not create `SelectAllButtonField`s based on conditions we encounter in the constructor of `ElectionOrganisationDivisionForm`. Note this would also require some additional fiddling to allow us to pick up which radio buttons relate to a particular organisation via javascript on the front-end.

Closes #116

### DIFF
--- a/every_election/apps/elections/templates/id_creator/election_organisation_division.html
+++ b/every_election/apps/elections/templates/id_creator/election_organisation_division.html
@@ -17,16 +17,17 @@
 <script>
 fallback.ready(function() {
 
-    $("<button type=button class=button>Mark all seats as contested (can't be undone)</button>")
-    .insertAfter(".inline_radios .form-group h2").on('click', function() {
-        $('input[value=seats_contested]').prop("checked", true);
-        $('input[value=seats_contested]').parent().addClass("selected");
-        $('input[value=no_seats]').parent().removeClass("selected");
-        $('input[value=by_election]').parent().removeClass("selected");
-    });
-});
+    $("form").prepend(
+      $("<button type='button' class='button'>Mark all seats as contested (can't be undone)</button>")
+        .on('click', function() {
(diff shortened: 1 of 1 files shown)


symroe:
I think I'd like the button creation to be in JS: this is because without JS the button wont work at all, so it should keep to the principles of progressive enhancement.

The real fix here, I think, would be to put each division list in it's own fieldset and then attach the button to each fieldset.

I don't think fieldsets are properly supported in django, so we'd have to do something like this https://gist.github.com/FZambia/5393983

chris48s:
OK no probs. I've modified the PR so the functionality works as described in the initial post (single button for the entire page) but I've created the button in javascript rather than as a django form element.

We can worry about one button per area as another issue. Fieldsets might be one way. If they are a pain, maybe another way to do it might be to add a css class to each radio group and its associated heading (which identifies the parent organistaion) that we can use as a selector?

PR #150, opened 2018-01-05
Create Election records for subtype groups

chris48s:
This PR adds in election subtypes as a proper heirarchy layer. In principle, this allows an election to have any combination of:

election type -> subtype -> organisation -> division

even though there is currently no election type we model that actually has all 4 of those things. These now operate in a strict hierarchy.

The upshot of this is we can:

* Deploy this
* Delete `naw.2018-02-06`, `naw.c.2018-02-06` and `naw.c.alyn-and-deeside.by.2018-02-06` from prod
* Re-create the IDs for `naw.c.alyn-and-deeside.by.2018-02-06` using this code (lets hold off on doing that for the moment though until we know if we also need to change the election system for `naw.c`)
* Modify the YNR import process to request elections from EE and do 2 passes over it:
  * One which makes a `Post` object for every election where the `group_type` is `NULL`
  * A second one which makes an `Election` object for the (first) parent election object of each `Post` + create a `PostExtraElection` join between them.
* That should then allow us to import everything cleanly (including this Welsh Assembly election) because all the meta-data we need should always be on the group id. We also shouldn't neeed to do anything else to the EE API to enable YNR to consume it.

We still need to account for mayors and PCCs.. not sure how best to do that. This seems like progress anyway.

refs https://github.com/DemocracyClub/yournextrepresentative/issues/320

### DIFF
--- a/every_election/apps/api/serializers.py
+++ b/every_election/apps/api/serializers.py
@@ -143,7 +143,7 @@ class ElectionSerializer(serializers.HyperlinkedModelSerializer):
         return obj.get_current
 
     def get_voting_system(self, obj):
-        if obj.group_type == 'organisation' or not obj.group_type:
+        if obj.group_type == 'organisation' or obj.group_type == 'subtype' or not obj.group_type:
             return VotingSystemSerializer(obj.voting_system).data
         return None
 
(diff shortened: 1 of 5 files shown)


symroe:
👍

PR #153, opened 2018-01-09
set voting system to FPTP in AMS Constituency elections

chris48s:
This isn't really the best way of doing this, but at the moment we can only store `voting_system` as a property of `ElectionType`. There is maybe a further bit of work here to allow us to store this information in the DB in a more flexible way rather than hard-coding rules in here but it requires some DB earth works :/


chris48s:
[Alyn & Deeside by-election on YNR](https://trello.com/c/qmrzDl8J/1089-alyn-deeside-by-election-on-ynr)

PR #156, opened 2018-01-11
Add Election ID Documentation

### LINKED ISSUE #48: Update reference definition

chris48s:
Notes:
* This is a useful starting point. It probably needs refinement, but it is a step up from nothing.
* I've ~fixed 'Election Types'~ made 'Election Types' less broken in this PR, but I'm not sure if it is necessarily a useful thing to have alongside this as a lot of the content is duplicated. Maybe it should just be replaced by this?
* I'm struggling to properly articulate when the organisation layer is relevant. An organisation is 'an administrative body which can hold an election', but sometimes that organisation is implicit in the election type and sometimes it needs to be explicit in the identifier but I can't quite explain the rule for when each of those scenarios is true.
* ~I have deliberately not talked about how you know what area names to use because this is still a bit.. fuzzy.~
* ~I've also left out a description of the slugging algortihm. I'm not quite sure how to address it.~
* I need to submit another PR to https://github.com/DemocracyClub/Website/ removing https://github.com/DemocracyClub/Website/blob/master/democracy_club/templates/election-ids/reference.html and linking to this.

Closes #48

### DIFF
--- a/every_election/apps/elections/models.py
+++ b/every_election/apps/elections/models.py
@@ -15,9 +15,6 @@ from .managers import ElectionManager
 
 
 class ElectionType(models.Model):
-    """
-    As defined at https://democracyclub.org.uk/projects/election-ids/reference/
-    """
 
     name = models.CharField(blank=True, max_length=100)
     election_type = models.CharField(blank=True, max_length=100, unique=True)
@@ -29,9 +26,6 @@ class ElectionType(models.Model):
 
 
 class ElectionSubType(models.Model):
-    """
-    As defined at https://democracyclub.org.uk/projects/election-ids/reference/
-    """
 
     name = models.CharField(blank=True, max_length=100)
     election_type = models.ForeignKey('ElectionType', related_name="subtype")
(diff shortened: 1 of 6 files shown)


chris48s:
[Every Election: New reference definition](https://trello.com/c/AIwLI4a3/927-every-election-new-reference-definition)

chris48s:
4a5e396 adds some more detail on a couple of points. Can we see if we can come up with a sentence on Monday to try and explain when the organisation layer is/isn't relevant. I think both of us are fairly keen to get some better docs for this live 📚

symroe:
> An administrative body which can hold an election.

"When am election type can only contain a single organisation the organisation part of the ID is excluded. This is to prevent `parl.parl` where the type and the organisation are both `parl`."

Is that the only case, actually?

chris48s:
`naw`, `sp`, `gla` are the same, but that makese sense - cheers

PR #157, opened 2018-01-14
improve naming for group ids

### LINKED ISSUE #155: Better "election_title" for subtype elections
On https://elections.democracyclub.org.uk/api/elections/naw.c.2018-02-06/ the "election_title" is `(Constituencies)`.

This should be…nicer. I'd say something like "Welsh assembly (Constituencies)" would do it.

chris48s:
Closes #155 

To deploy this we'll need to re-run `python manage.py add_election_types` to update the titles on the election types table.

### DIFF
--- a/every_election/apps/api/tests/test_api_election_endpoint.py
+++ b/every_election/apps/api/tests/test_api_election_endpoint.py
@@ -128,7 +128,7 @@ class TestElectionAPIQueries(APITestCase):
                         "territory_code": "ENG"
                     },
                     "election_type": {
-                        "name": "Local elections",
+                        "name": "Local Elections",
                         "election_type": "local"
                     },
                     "explanation": null,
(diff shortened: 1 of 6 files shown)


symroe:
Code LGTM but @JoeMitchell might have some opinions about the title casing?

chris48s:
You can't really assign a context here because it's data, not prose. If I take that data and use it like this

```<h1>{{ election.name }}</h1>```

then the context is different than if I take that data and use it like

```the {{ election.name }} took place on {{ election.poll_open_date }}```

but it is still the same data point.

We may surface this in both ways:

1. Here's an example of these being used as titles in YNR:
screenshot at 2018-01-15 10 46 07

2. Here's an example of using in mid-sentence in WhoCIVF:
screenshot at 2018-01-15 10 46 20

JoeMitchell:
Looks good to me!

chris48s:
@JoeMitchell - proposed new election titles:

Local elections
European Parliament (UK) elections
Northern Ireland Assembly elections
Welsh Assembly elections
Scottish Parliament elections
Greater London Assembly elections
Police and Crime Commissioner elections
Mayoral elections

JoeMitchell:
All good, except I'd go for the official name of NAW

>National Assembly for Wales elections

And then the missing one I spose is

>UK Parliament elections

PR #162, opened 2018-01-22
 Improve script for importing DivisionSets

### LINKED ISSUE #47: Ensure unicode names are imported correctly
e.g. Eilean a' Cheò

chris48s:
Sorry - this does several things in one commit, but it was difficult to avoid :(

1. Instead of setting the end date on the previous DivisionSet based on the start date of the new one, allow the start date of the new DivisionSet to be inferred from the end date of the old one (+ perform necessary validation)
2. Build all our objects and then persist them in a transaction. This reduces the chance of a bad data point leaving our database in an inconsistent state.
3. Assume CSV encoding is `utf-8` if importing from google docs. Google docs doesn't serve up an encoding header so requests assumes `ISO-8859-1` (because that's what the spec says). This was giving me some encoding errors. Manually setting to `utf-8` fixes.

Hopefully it is sufficiently well commented that it is not too difficult to review.

I didn't actually intend to fix issue #47 but it just sort of.. happened :D
Closes #47

@symroe - once you've reviewed this, we can run `python manage.py import_divisionsets_from_csv "https://docs.google.com/spreadsheet/ccc?key=18X4wYGGmJNbL9xwN-ALXPQYDAed4PN1M9xC0YiKZ-TM&output=csv&gid=1833724937"` - that will get us to a stage where we can create ids for all the elections happening this May (all the reviews relating to May 2018 now have a made ECO), although we will still need to sort out the DivisionGeographies as the next job.

Also, I would like to add some tests for this. Do you reckon we should just get this merged so we can import the names and worry about tests later, or shall I add tests to this PR first?


chris48s:
Oh yeah, the diff for the PR is not terribly useful because I've renamed the file in the first commit, so you'll want to just look at c55b715 to review

chris48s:
have added tests to this now

every_election/apps/organisations/management/commands/import_divisionsets_from_csv.py
```diff
@@ -99,24 +146,34 @@ def create_div_from_line(self, div_set, identifier, line):
         if not seats_total:
             seats_total = 1
 
-        div, _ = OrganisationDivision.objects.update_or_create(
```
symroe: Sorry if I'm missing something – will this stop the script from being idempotent?

every_election/apps/organisations/tests/test_import_divisionsets_from_csv.py
```diff
@@ -0,0 +1,156 @@
...
+        cmd.get_data = lambda x: [self.base_record]
+        with self.assertRaises(Exception):
+            cmd.handle(**{'url': 'foo.bar/baz'})
+
+    def test_valid(self):
+        # all data is valid - should import cleanly
+        cmd = Command()
+        cmd.get_data = lambda x: self.valid_test_data
+        cmd.get_division_type_from_registers = lambda x: 'DIW'
+        cmd.handle(**{'url': 'foo.bar/baz'})
```
symroe: As per my other comment – if this line is duplicated then the test fails.

Is this expected? Happy to be convinced it's ok to be like this however the old command, for all its problems, worked find if called twice, so this is a new 'feature'.

(Error is `Exception: End date for previous DivisionSets 119: (2017-05-04 to now) is NULL`)

chris48s:
Discussed this on a call, but for the benefit of future Chris + Sym looking back on this, quick bullet point summary:

* Previously we were specifying the start date for the new DivisionSet and then using that to infer an end date for the previous DivisionSet
* Now we're moving to a model where we don't specify a start date for the new DivisionSet and infer it based on the end date of the most recent DivisionSet (which we should have already set using `manage.py update_end_dates`
* If we import a new DivisionSet and try to import it again, it (correctly) fails because the end date of the most recent DivisionSet (the one we just imported) is now `NULL`
* This does change the behaviour of this command, but it is by design.
* This does also mean we need one CSV per import rather than a single 'rolling' CSV, but it makes more sense in the long-term.
* All of that is non-obvious. I do plan to properly document all this as I follow the process through a few times

PR #171, opened 2018-01-30
never try to call DC mapit

### LINKED ISSUE #169: API call with bad postcode throws 500 error
e.g: https://elections.democracyclub.org.uk/api/elections.json?postcode=INVALID&future=1

We're still attempting to fall back to DC mapit (there was always going to be one!)

https://github.com/DemocracyClub/EveryElection/blob/bea4be1d414fef2486fb9b8f0f79361c84ecdc6c/every_election/apps/elections/query_helpers.py#L58-L59

..which is now gone.

In general, we need to get rid of the ONSPD static Json gubbins and move EE to using `uk_geo_utils` but as a interim we should at least stop trying to contact a server that doesn't exist anymore.

chris48s:
Closes #169


symroe:
:wave: bye!

PR #176, opened 2018-02-05
upgrade to Django 1.11

### LINKED ISSUE #149: Upgrade to Django 1.11
Django 1.10 is now unsupported and no longer receives security patches.

There's at least one blocker to upgrading this repo which is we need to fix https://github.com/DemocracyClub/dc_base_theme/issues/12 first

I suspect there are relatively few other deprecations/compatibility issues to worry about when upgrading this repo but it needs some testing..

chris48s:
This is not ready for merge yet - need to get https://github.com/DemocracyClub/dc_base_theme/pull/24 in first so I can upgrade the theme dependency - should be good to go after that.

Closes #149

Side issue while I am looking at dependencies: given most of our dependencies on this project are un-pinned we're occasionally unwittingly upgrading them, so I think I'm going to work through them, pin a load of stuff and enable pyupbot on this repo but I will deal with that as a different piece of work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,9 @@ script:
 after_success:
   - coveralls
 before_cache:
-    - rm -f .tox/py34-django110/log/*.log
-    - rm -f .tox/py35-django110/log/*.log
-    - rm -f .tox/py36-django110/log/*.log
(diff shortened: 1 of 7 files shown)


chris48s:
This should be good to go now..

PR #178, opened 2018-02-06
don't return null if official_identifier key is set

chris48s:


chris48s:
this is a one line fix for an obvious bug and I need it tomorrow, so I'm just going to merge it..

PR #184, opened 2018-02-14
explicitly require css compressor

chris48s:


chris48s:
I'm just going to merge this so I can crack on with the deploy work

PR #195, opened 2018-02-22
Be trailing-slash agnostic

### LINKED ISSUE #187: Be trailing-slash agnostic
This is probably (hopefully) an easy fix

https://elections.democracyclub.org.uk/reference_definition works but
https://elections.democracyclub.org.uk/reference_definition/ 404s

https://elections.democracyclub.org.uk/elections works but
https://elections.democracyclub.org.uk/elections/ 404s

etc

We should be set up so that both version work, like this:

https://wheredoivote.co.uk/privacy
https://wheredoivote.co.uk/privacy/

chris48s:
Closes #187 - cheers for the tip on `show_urls` - that is very useful

I'll add this same check to WhereDIV as well once I can tidy up the mess on my working copy well enough to work on other things.. Now I've started looking at it I realise that there are a few urls on there with the same issue and it is really easy add a new route and not notice this.

I'm not going to do anything to WhoCIVF for the moment because the majority of the routes don't have a trailing slash and I'm slightly worried about introducing bugs by editing all the routes just to satisfy my trailing-slash-based OCD :D


symroe:
Longer term I think it's worth pulling this in to a package with some handy generic tests in (I keep thinking about that video you showed me where they had a load of simple tests for stuff like this).

FWIF I think it's fairly easy to do without requiring `django-extensions` too:

```python
from django.conf import settings
from django.urls import RegexURLPattern, RegexURLResolver

urlconf = __import__(getattr(settings, 'ROOT_URLCONF'), {}, {}, [''])

def list_patterns(patterns):
    
    for pattern in patterns:
        if isinstance(pattern, (RegexURLResolver)):
            list_patterns(pattern.url_patterns)
        else:
            if not pattern.regex.pattern.endswith(('^$', '/$')):
                print(pattern.regex.pattern)

list_patterns(urlconf.urlpatterns)
```

This clearly needs some work, but you get the idea. It would be nice for tests to fail if anything matched (we'd need a way of excluding some patterns form being tested, like anything ending in a file extension (`.xml` for the site map) or the static and media paths.

chris48s:
Yeah - it does seem like it would be handy if we could just `pip install` this..

> we'd need a way of excluding some patterns form being tested

Yeah I did this on https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/pollingstations/tests/test_urls.py#L8-L13 but we could probably assemble a reasonably sensible generic whitelist

PR #199, opened 2018-03-01
prep work to enable pyup

chris48s:


chris48s:
I'm just going to go rogue and merge this because it doesn't change anything - it is just pinning stuff with no version so pyup doesn't go mental and submit 50 pull requests when I enable it..

PR #210, opened 2018-03-02
Import posts per ward

symroe:
Add a management command for importing posts per ward from a CSV file.

This command has opinions about `seats_total`. I think that's a good thing, as we should have seats total for everywhere. 

There are options though:

1. If you want to set the seats total from the CSV file then add `--trust-csv-seats-total`
2. If you want to ignore seats total add `--skip-unknown`

### DIFF
--- a/every_election/apps/core/mixins.py
+++ b/every_election/apps/core/mixins.py
@@ -1,0 +1,72 @@
+import csv
+import re
+
+import requests
+from django.conf import settings
+
+from storage.s3wrapper import S3Wrapper
+
+
+class ReadFromCSVMixin(object):
+    """
+    A generic mixin for for a reading from a CSV file in a Django
+    management command.
+
+    The CSV file can be a local file, a URL or on an S3 bucket.
+
+    `self.S3_BUCKET_NAME` needs to be set when using S3.
+    """
+    ENCODING = 'utf-8'
+    DELIMITER = ','
+    S3_BUCKET_NAME = settings.LGBCE_BUCKET
+
+    def add_arguments(self, parser):
+        group = parser.add_mutually_exclusive_group(required=True)
+        group.add_argument(
+            '-f',
+            '--file',
+            action='store',
+            help='Path to import e.g: /foo/bar/baz.csv',
+        )
+        group.add_argument(
+            '-u',
+            '--url',
+            action='store',
+            help='URL to import e.g: http://foo.bar/baz.csv',
+        )
+        group.add_argument(
+            '-s',
+            '--s3',
+            action='store',
+            help='S3 key to import e.g: foo/bar/baz.csv'
+        )
+
+    def read_local_csv(self, filename):
+        f = open(filename, 'rt', encoding=self.ENCODING)
+        reader = csv.DictReader(f, delimiter=self.DELIMITER)
+        return list(reader)
+
+    def read_csv_from_url(self, url):
+        r = requests.get(url)
+
+        # if CSV came from google docs
+        # manually set the encoding
+        gdocs_pattern = r'(.)+docs\.google(.)+\/ccc(.)+'
+        if re.match(gdocs_pattern, url):
+            r.encoding = self.ENCODING
+
+        csv_reader = csv.DictReader(r.text.splitlines())
+        return list(csv_reader)
+
+    def read_csv_from_s3(self, filepath):
+        s3 = S3Wrapper(self.S3_BUCKET_NAME)
+        f = s3.get_file(filepath)
+        return self.read_local_csv(f.name)
+
+    def load_csv_data(self, options):
+        if options['file']:
+            return self.read_local_csv(options['file'])
+        if options['url']:
+            return self.read_csv_from_url(options['url'])
+        if options['s3']:
+            return self.read_csv_from_s3(options['s3'])
--- a/every_election/apps/elections/management/commands/attach_posts_per_election_from_csv.py
+++ b/every_election/apps/elections/management/commands/attach_posts_per_election_from_csv.py
@@ -1,0 +1,122 @@
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from core.mixins import ReadFromCSVMixin
+from elections.models import Election
+
+
+class Command(ReadFromCSVMixin, BaseCommand):
+
+    SEATS_CONTESTED_FIELD = 'posts up'
+
+    help = """
+    Given a CSV file, path or url import the number of contested seats for that
+    election.
+
+    Expected CSV format is, e.g:
+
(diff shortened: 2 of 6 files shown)


symroe:
[Attach Posts-per-ward to EE](https://trello.com/c/xMDHSBX1/1049-attach-posts-per-ward-to-ee)

every_election/apps/core/mixins.py
```diff
@@ -0,0 +1,70 @@
+from storage.s3wrapper import S3Wrapper
+import csv
+import re
+import requests
+
+from django.conf import settings
+
+
+class ReadFromCSVMixin(object):
```
chris48s: :+1: This is a really useful abstraction. Do you think you could also convert [update_end_dates](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/organisations/management/commands/update_end_dates.py) to use it as well? The boilerplate is very similar.

I know it is a little bit different, but I reckon (untested) you can probably just pass [parse_csv()](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/organisations/management/commands/update_end_dates.py#L70-L77) a `DictReader` object instead of a `reader` object and it would work the same or need a one line modification. If it turns out to be a pain, ignore it and I'll refactor it as a seperate piece of work to avoid holding this up.

every_election/apps/elections/management/commands/attach_posts_per_election_from_csv.py
```diff
@@ -0,0 +1,113 @@
+from django.core.management.base import BaseCommand
+from core.mixins import ReadFromCSVMixin
+
+from elections.models import Election
+
+
+class Command(ReadFromCSVMixin, BaseCommand):
+
+    SEATS_CONTESTED_FIELD = 'posts up 2018'
```
chris48s: Why 2018? I don't see any reason this can't be re-used at another time.

every_election/apps/elections/management/commands/attach_posts_per_election_from_csv.py
```diff
@@ -0,0 +1,113 @@
...
+    election.
+
+    Optionally update the seats total for that election if
+    --trust-csv-seats-total specified.
+    """
+
+    def add_arguments(self, parser):
+        super(Command, self).add_arguments(parser)
+        parser.add_argument(
+            '--trust-csv-seats-total',
```
chris48s: I think the `trust-csv-seats-total` flag should have a name like `replace-seats-total` or `overwrite-seats-total` which makes it clearer that using this flag will overwrite existing database content (even if it is non-`NULL`).

every_election/apps/elections/management/commands/attach_posts_per_election_from_csv.py
```diff
@@ -0,0 +1,113 @@
...
+        if election.division.seats_total:
+            db_seats_total = election.division.seats_total
+
+        if line['seats_total']:
+            csv_seats_total = int(line['seats_total'] or 0)
+
+        if db_seats_total:
+            # Warn if the CSV has a different seats total
+            if csv_seats_total != db_seats_total:
+                print(
```
chris48s: Using `self.stdout.write` rather than `print` makes it easier to surpress/capture stdout under test.

every_election/apps/elections/management/commands/attach_posts_per_election_from_csv.py
```diff
@@ -0,0 +1,113 @@
...
+                print("Taking seats total from CSV for {}".format(
+                    election.election_id,
+                ))
+                seats_total = csv_seats_total
+            else:
+                raise ValueError("Seats total not known for {}".format(
+                    election.election_id
+                ))
+
+        if not seats_total >= seats_contested:
```
chris48s: Bit of a nitpick: This is definitely a useful check, but it doesn't belong in the `get_seats_total()` method. `get_seats_total()` should just well.. get `seats_total` and then this check should be performed using the result of that fetch. I was looking for this check elsewhere.

every_election/apps/elections/management/commands/attach_posts_per_election_from_csv.py
```diff
@@ -0,0 +1,113 @@
...
+                    if options['skip_unknown']:
+                        print(e.args[0])
+                    else:
+                        raise
+
+                election.seats_contested = seats_contested
+                if seats_total and trust_csv:
+                    election.seats_total = seats_total
+
+                election.save()
```
chris48s: I think it would be safer if this command batched up all the database I/O and performed it in a transaction at the end so that there is less risk of leaving the DB in an inconsistent state. `import_divisionsets_from_csv` has an example of this.

every_election/apps/core/mixins.py
symroe: Was a little more complex than passing the DictReader list directly, but not too much.

every_election/apps/elections/management/commands/attach_posts_per_election_from_csv.py
symroe: It's what's in the CSV :) But agreed it doesn't need to be. Have pushed a fixup, but we need to update the Google Doc before we can import from it now.

every_election/apps/elections/management/commands/attach_posts_per_election_from_csv.py
```diff
@@ -0,0 +1,123 @@
...
+
+    Expected CSV format is, e.g:
+
+    ORG Name,ORG GSS,REG,geography_curie,ward name,seats_total,has_election,posts up,created,id,problem/note
+    Eastleigh,E07000086,EAT,EAT:bishopstoke,Bishopstoke,3,yes,3,yes,local.eastleigh.bishopstoke.2018-05-03,
+    Eastleigh,E07000086,EAT,EAT:botley,Botley,2,yes,2,yes,local.eastleigh.botley.2018-05-03,
+
+
+    Optionally update the seats total for that election if
+    --trust-csv-seats-total specified.
```
chris48s: Guessing this should now read `--replace-seats-total`?

every_election/apps/elections/management/commands/attach_posts_per_election_from_csv.py
```diff
@@ -0,0 +1,123 @@
...
+            help="Replace seats total in the database with the value in the CSV file"  # noqa
+        )
+        parser.add_argument(
+            '--skip-unknown',
+            action='store_true',
+            help="Skip elections with unknown seats total"
+        )
+
+    def get_seats_total(
+            self, election, line, trust_csv=False):
```
chris48s: now that you've removed a long param (think it was `seats_contested`) its probably easier to read this if the whole declaration is on one line

PR #223, opened 2018-03-09
Store additional meta-data object against election objects (for voter id pilots)

chris48s:
Sometimes we will want to store some kind of additional meta-data about an election. The current use-case for this is the voter id pilots but inevitably other things will happen. Rather than try to predict a data model for this, lets just take the very broad view that sometimes **a thing** will happen and we want to be able to attach info about it to an election object.

This PR just gives us the ability to define a big ol blob of json and attach it to one or more election objects via /admin.

There is some additional work to do around:
a) working out how we should use this object in general and
b) what content we should put in it WRT the upcoming voter id trails

but I'm going to keep this PR unopinionated on those topics and just focus on the mechanics.


chris48s:
A bit more thought on how we go about actually using this (doesn't affect the code itself)..

Sometimes **a thing** will happen and we want to be able to attach info about it to an election object.

Real life examples:

* May 2018 Voter ID Pilots
* City of London elections: Non-standard polling station opening times

Fictional examples:

* You can vote at any polling station
* Online voting is available in your area
* Late night polling station opening hours
..whatevs

I'm fairly clear that we don't want the meta-data object to have different fields every time. In principle it should be able to support multiple notifications for a single election, so what I don't want is for one election object to have a key:

```
...
"metadata": {
    "title": "Bromley Voter ID Pilot",
    "summary": "You must bring photo ID to the polling station",
    "detail": "To vote at a polling station in London Borough Bromley on 3 May 2018 you must take either one of the following:\n* a passport issued by the United Kingdom, a Commonwealth country or a member state of the European Union\n* a photocard driving licence (including a provisional licence) issued in the United Kingdom or by a Crown Dependency, or by a member State of the European Union\nbla bla bla bla bla",
    "url": "http://www.bromley.gov.uk/info/200033/elections_and_voting/1177/voter_id_pilot"
}
...
```

and then another election to have an object

```
...
"metadata": {
    "title": "Extended polling station opening hours are available in your area",
    "start_time": "2018-05-03T06:00:00.000Z",
    "end_time": "2018-05-03T23:59:59.000Z",
    "detail": "Roll out of the pub, grab some chips and cast your vote: what could go wrong?"
}
...
```

as that is going to make any code which parses it very brittle. What we want to do is:

```
"metadata": {
    "voter_id_trial": {
        "title": "Bromley Voter ID Pilot",
        "summary": "You must bring photo ID to the polling station",
        "detail": "To vote at a polling station in London Borough Bromley on 3 May 2018 you must take either one of the following:\n* a passport issued by the United Kingdom, a Commonwealth country or a member state of the European Union\n* a photocard driving licence (including a provisional licence) issued in the United Kingdom or by a Crown Dependency, or by a member State of the European Union\nbla bla bla bla bla",
        "url": "http://www.bromley.gov.uk/info/200033/elections_and_voting/1177/voter_id_pilot"
    },
    "late_night_polling_stations": {
        "title": "Extended polling station opening hours are available in your area",
        "start_time": "2018-05-03T06:00:00.000Z",
        "end_time": "2018-05-03T23:59:59.000Z",
        "detail": "Roll out of the pub, grab some chips and cast your vote: what could go wrong?"
    }
}
```

and then make sure that any object under `voter_id_trial` always has the keys `title`, `summary`, `detail`, `url` + so on. That way:

* We can combine them if we want
* If your code finds an object called `voter_id_trial` under `metadata` it knows how to deal with it.. if its a different(ly structured) object, it has a different key. This makes it easier for our applications and third parties consuming our code.

The thing I'm not sure on is what the top-level keys should be.. we could just give them arbitrary names, like I've done above. Do you have any better ideas on this, @symroe ?

symroe:
I reckon this is absolutely good enough for our current sue case, and it's really hard to have opinions about future uses until they happen. Basically, I don't feel able to review how suitable this will be in future…

A few thoughts:

I get that the JSON field gives us most flexibility and that's useful right now, but in your comment you said that clients could expect `title`, `summary`, `detail`, `url`. I think it's reasonable to make them nullable keys that always exist, because ultimately they're very flexible concepts that would be useful for a tonne of use cases.

Given that, is it worth putting them on the `MetaData` model as actual DB fields, and then having a JSON field called `extra` or something. The `extra` field could be added to the API output. This would save accidentally editing the JSON in such a way as to rename a required field (a space in the `url` key for example), but also allow for the extendibility we currently have in this PR.

I think the trade off between having a slightly less flexible model (in that we will have standard fields) is worth it for predictability – in most cases I expect we'll be able to bend our use case to these fields.

Would like your thoughts on this.

As for the name of the key – `metadata` feels ok, but might not indicate that it's something we want API consumers to pay attention to. Maybe `additional_info` or `special_cases` or something along those lines?

Like I say, I'm sort of shooting in the dark here as I'm not sure of the use cases in future. If you disagree with my comments then I'm happy for this to go in as it's the quickest most flexible option that doesn't tie us in very much, and we could always migrate in to something else if we needed to.

chris48s:
If you make the meta-data model a Record(title, summary, detail, url), do you propose that `metadata` (or `additional_info` or `special_cases` or whatever you want to call it) should be an array or an object?

symroe:
I agree with you that it should be an array – so it would need to be a many to many between Election and MetaData. Like I say, I'm happy to ignore this for now and think about it when we have more use cases.

chris48s:
Don't mind doing it as an Record/many-to-many.

If we're not going to do that, we should still think about how to key sub-objects inside the `metadata` object. If we do one then change to the other, it does introduce BC-breaks for consumers down the line. That said, by the time we have multiple examples it is probably a few years down from now..

symroe:
Happy to have a chat about that – or we can figure it out later as we don't have other examples. Let me know how you'd like to progress…

chris48s:
For the benfit of future Chris and Sym:

Lets do this for now.

We'll present data in the form:

```json
"metadata": {
    "id1": {
        "key": "value"
    },
    "id2": {
        "key": "value"
    }
}
```

If at a later date (having spotted some patterns) we want to switch to a more structured format with a M2M relation, instead of moving to 

```json
"metadata": [
    {
        "id": "id1",
        "structured": "data",
        "extra": {}
    },
    {
        "id": "id2",
        "structured": "data",
        "extra": {}
    }
]
```

we will transform at serialization time, so we can still present it like

```json
"metadata": {
    "id1": {
        "structured": "data",
        "extra": {}
    },
    "id2": {
        "structured": "data",
        "extra": {}
    }
}
```

that will allow us to

* maintain backwards compatibility and
* it is easier to work with because we can say `if key in object` or `if object.key` rather than having to iterate an array of N objects to see if a particular id is in there

PR #260, opened 2018-04-16
Add filter for elections with metadata

symroe:
Adding this so that it's easy to get a list of elections with a metadata object.

This is mainly for WhoCanIVoteFor's use, so that it doesn't need to iterate over every election to get the ones with metadata in.

Of course it could just look for the key in the response from the postcode lookup, but the current architecture just uses that response to get a list of IDs that are then fed directly in to a queryset. returning the metadata alongside that would be confusing at the moment, or require some refactoring.

Rather than doing that, it's easier to store the metadata on the election object at import time, and this filter speeds that up.

### DIFF
--- a/every_election/apps/api/tests/test_api_election_endpoint.py
+++ b/every_election/apps/api/tests/test_api_election_endpoint.py
@@ -7,6 +7,7 @@ from rest_framework.test import APITestCase
 from elections.tests.factories import ElectionFactory
 from organisations.tests.factories import (
     OrganisationFactory, OrganisationDivisionFactory)
+from elections.models import MetaData
 
 
 class TestElectionAPIQueries(APITestCase):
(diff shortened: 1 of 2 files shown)


chris48s:
Before I merge this, you have remembered that WhoCIVF imports elections from YNR, not EE - right?

https://github.com/DemocracyClub/WhoCanIVoteFor/blob/bb658e3f7f705fe81265149d1a50e33ba04c3fec/wcivf/apps/elections/management/commands/import_elections.py

symroe:
:D Yep!

I have this patch locally that I'll PR later:
```diff --git a/wcivf/apps/elections/management/commands/import_elections.py b/wcivf/apps/elections/management/commands/import_elections.py
index 469c3cb..1e952cf 100644
--- a/wcivf/apps/elections/management/commands/import_elections.py
+++ b/wcivf/apps/elections/management/commands/import_elections.py
@@ -15,9 +15,27 @@ class Command(BaseCommand):
             self.add_elections(results)
             next_page = results.get('next')
 
+        # Now import any metadata from EE
+        self.import_metadata()
+
     def add_elections(self, results):
         for election in results['results']:
             election_obj, created = Election.objects.update_or_create_from_ynr(
                 election)
             if created:
                 print("Added new election: {0}".format(election['name']))
+
+    def import_metadata(self):
+        req = requests.get(
+            settings.EE_BASE + '/api/elections/?current=true&metadata=1')
+
+        metadata_for_ids = []
+
+        for election in req.json()['results']:
+            try:
+                election_obj = Election.objects.get(
+                    slug=election['election_id'])
+                election_obj.metadata = election['metadata']
+                election_obj.save()
+            except:
+                pass
```

chris48s:
right, yep that makes sense :+1:

PR #281, opened 2018-05-08
Election GeoJSON endpoint and some style tweaks

russss:
Adds a `/api/election/<id>/geo` endpoint for elections.

I've also added links to the API from the election and organisation pages, and improved the styling on the election page:

image

### DIFF
--- a/every_election/apps/api/serializers.py
+++ b/every_election/apps/api/serializers.py
@@ -96,6 +96,7 @@ class ElectionTypeSerializer(serializers.HyperlinkedModelSerializer):
         depth = 1
 
 
+
 class ElectionSubTypeSerializer(serializers.ModelSerializer):
     class Meta:
         model = ElectionSubType
@@ -176,3 +177,27 @@ class ElectionSerializer(serializers.HyperlinkedModelSerializer):
         )
         depth = 1
 
+
+class ElectionGeoSerializer(GeoFeatureModelSerializer):
+    geography_model = GeometrySerializerMethodField()
+
+    def get_geography_model(self, obj):
+        if obj.geography is None:
+            return None
+        return obj.geography.geography.simplify(0.0009)
+
+    class Meta:
+        model = Election
+        extra_kwargs = {
+            'url': {
+                'view_name': 'election-geo',
+                'lookup_field': 'pk'
+            }
+        }
+
+        geo_field = 'geography_model'
+
+        fields = (
+            'election_id',
+            'election_title'
+        )
--- a/every_election/apps/elections/templates/elections/election_detail.html
+++ b/every_election/apps/elections/templates/elections/election_detail.html
@@ -4,37 +4,49 @@
 {% block content %}
 <section class="columns large-6 large-centered">
 
-    <div class="card postcode_card">
-
-        <h1 style="font-size:1.6em">{{ object.election_id }}</h1>
-        <h2 style="font-size:1.4em">{{ object.election_title }}</h2>
-        <h2 style="font-size:1.4em">{{ object.election_type }}</h2>
+    <div class="card postcode_card election">
+        <h1>{% if object.election_title %}
+                {{ object.election_title }}
+            {% else %}
+                {{ object.election_id }}
+        {% endif %}</h1>
         {% if object.election_subtype %}
         <h3>{{ object.election_subtype }}</h3>
         {% endif %}
-
-        <p>Election date: <strong>{{ object.poll_open_date }}</strong></p>
-        <p>Election type: <code>{{ object.election_type.election_type }}</code></p>
+        <dl>
+            <dt>ID</dt> <dd>{{ object.election_id }}</dd>
+            <dt>Type</dt> <dd>{{ object.election_type }}</dd>
+            <dt>Date</dt> <dd>{{ object.poll_open_date }}</dd>
         {% if object.organisation %}
-        <p>Organisation: <a href="{{ object.organisation.get_absolute_url }}">
-            {{ object.organisation }}</a>.</p>
+            <dt>Organisation</dt><dd><a href="{{ object.organisation.get_absolute_url }}">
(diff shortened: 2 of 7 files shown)


chris48s:
[Review 2x EE PRs](https://trello.com/c/ilf6KXJh/1347-review-2x-ee-prs)

every_election/apps/api/serializers.py
```diff
@@ -176,3 +177,27 @@ class Meta:
...
+                'lookup_field': 'pk'
+            }
+        }
+
+        geo_field = 'geography_model'
+
+        fields = (
+            'election_id',
+            'election_title'
+        )
```
chris48s: Could the output of the `/elections/{id}/geo` endpoint serialize the full details of the election object in the `properties` of the feature object?

For example if I call
https://elections.democracyclub.org.uk/api/organisations/DNC/ and
https://elections.democracyclub.org.uk/api/organisations/DNC/geo/
the properties object on the `/geo` response has the same keys as the non-geo object, but the properties of `/elections/{id}/geo` and `/elections/{id}` are quite different.

The way I've done this on the polling stations project without loads of duplication is I've defined all the common stuff in a base `Serializer` class and then declared a `DataSerializer` and a `GeoSerializer` which inherit from it. See https://github.com/DemocracyClub/UK-Polling-Stations/blob/master/polling_stations/apps/api/pollingstations.py#L15-L66 for an example. We could use the same pattern here.

every_election/apps/api/serializers.py
```diff
@@ -176,3 +177,27 @@ class Meta:
...
         depth = 1
 
+
+class ElectionGeoSerializer(GeoFeatureModelSerializer):
+    geography_model = GeometrySerializerMethodField()
+
+    def get_geography_model(self, obj):
+        if obj.geography is None:
+            return None
+        return obj.geography.geography.simplify(0.0009)
```
chris48s: Hmm, so the comment I was going to leave here was:

> Simplification is useful for display purposes (e.g: making a coverage map), but not for working out if a point is inside/outside a polygon. Can we serve the unsimplified boundary by default and add a param to configure this e.g: `?simplify_tolerance=0.0009`.

..but then I've realised you've copied this from the `/organisations/{id}/geo` endpoint where we are simplifying by default. @symroe - what was your thinking on that? My inclination is we shouldn't simplify by default (on both endpoints)

every_election/apps/elections/templates/elections/election_detail.html
```diff
@@ -53,20 +65,34 @@ <h3>{{ object.election_subtype }}</h3>
...
+        <ul>
+            <li><a href="{% url 'api:election-detail' election_id=object.election_id %}">
+                    Election detail (JSON)
+                </a>
+            </li>
+            {% if object.geography %}
+            <li><a href="{% url 'api:election-geo' election_id=object.election_id %}">
+                    Election boundary (GeoJSON)
+                </a>
+            </li>
```
chris48s: Nitpick: this block should be indented one level.

every_election/apps/elections/templates/elections/election_detail.html
```diff
@@ -4,37 +4,49 @@
...
+        </dl>
         {% if object.group_type == 'organisation' %}
-        <p>View <a href="https://whocanivotefor.co.uk/elections/{{ object.election_id }}/{{ object.election_type|slugify }}">
-            this election on WhoCanIVoteFor</a>.</p>
-        <p>View <a href="https://candidates.democracyclub.org.uk/election/{{ object.election_id }}/constituencies">
-            this election on candidates.democracyclub.org.uk</a>.</p>
+            <p>View this election on:</p>
+            <ul>
+                <li><a href="https://whocanivotefor.co.uk/elections/{{ object.election_id }}/{{ object.election_type|slugify }}">WhoCanIVoteFor</a></li>
+                <li><a href="https://candidates.democracyclub.org.uk/election/{{ object.election_id }}/constituencies">candidates.democracyclub.org.uk</a></li>
```
chris48s: nice! :+1:

As a future improvement, I'd like to do this on ballot ids as well, but we need to get our house in shape on c.DC and WhereDIV re: actually using our own ids first before we can do it! :(

chris48s:
I really like all the additional info you've added in this PR, but I reckon things look a bit cluttered in the case where we don't have a notice of election doc yet:

screenshot at 2018-05-15 16 29 24

Maybe some additional horizontal spacing around the form would help? Perhaps this should actually be more than one card? If you've got a clear idea for improvement, feel free to push it. If not, I'm inclined to go ahead with this anyway as this improves what we have - we could look at improving that layout later.

chris48s:
I've intentionally not followed up on this for a few weeks because we've been doing a bunch of restructuring work in #315 which affects this. Now that I've merged that, there aren't any merge conflicts to resolve as such, but I tried rebasing 6ae2e1a onto the tip of `master` and there are now a bunch of tests failing:

```
every_election/apps/api/tests/test_api_election_endpoint.py FFF.FFFFF
every_election/apps/api/tests/test_api_organisation_endpoint.py .F.F.F
```

Sorry for slightly pulling the rug out from under you on this.. given the API/data model changes this needs to account for are relatively fresh in my mind, I don't mind 'adopting' this PR, accounting for those changes and finishing off these last few bits that need doing to take this over the line.

If you're mad keen to work on it, let me know. If not I'll aim to pick this up as there are several useful changes here that would be useful to get finalised before `master` diverges too much more.

Cheers

russss:
> Sorry for slightly pulling the rug out from under you on this.. given the API/data model changes this needs to account for are relatively fresh in my mind, I don't mind 'adopting' this PR, accounting for those changes and finishing off these last few bits that need doing to take this over the line.

Go for it. I'm a bit busy at the moment anyway and this PR has been nagging me so it would be good to see it merged.

chris48s:
now moved to #346 - cheers for your work on this

Issue #283, opened 2018-05-09
Offset in some local borough geodata


russss:
After getting my boundary export code working, I've unfortunately noticed that a significant number of boundaries in EE are exhibiting what appears to be a constant offset of a few hundred metres.

![image](https://user-images.githubusercontent.com/45057/39808855-f8fa1cea-5377-11e8-9346-ca77a295953a.png)

The offset seems to mostly affect entire local authorities, and I've only verified it in the elections for 2018-05-03. A limited selection of affected areas are:

* Cherwell
* Rochford
* Watford Borough
* Winchester
* Welwyn Hatfield
* Merton **Abbey ward only**
* Sheffield
* Craven **Penyghent, Settle and Ribblebanks, and Hellifield and Long Preston wards only**
* Knowsley

This strikes me as rather a large offset to be caused by differences in CRS transformation.

chris48s:
It has taken quite a while to get this sorted, but I've now _mostly_ fixed this.

There are still a few further issues around this topic:

* Some Northern Ireland boundaries still have issues, but these can't be corrected by importing from BoundaryLine because OS only covers GB. This spawns is a separate bucket of work around consuming open data from OSNI, which I've raised as #364
* There are 12 local gov wards which have had 'technical' boundary changes (i.e: where the division boundaries were changed but there was no Electoral Change Order). Some of these may still have issues and I haven't re-imported them. I need to do a bit of extra work to get those sorted (and work out how to deal with it in future). I've raised #278 to deal with those.
* The new County Electoral Divisions which were introduced prior to the 2016 local elections are still using boundaries imported from LGBCE's draft recommendations. We haven't back-ported the boundaries from BoundaryLine on to them because BoundaryLine doesn't (yet) have GSS codes for CEDs and we aren't able to assign them a reliable/stable identifier. I've raised #360 as a placeholder for this. I'll review it next when the October BL release appears. None of these should be _offset_ but some of them will probably not tessellate exactly in a layer with CED polygons sourced from BoundaryLine.

Other than the cases identified above which are now covered under more specific issues, this problem should be fixed (including for historic boundaries where we hold them) so I am going to close the issue. If you spot anything else, feel free to re-open (hopefully not! :crossed_fingers: ) Now that most of these issues are sorted, I've now gone back and re-generated all the election TopoJSON layers for all the historic elections back to 2015 and whacked them all in the S3 bucket. If you want to use those to check any known errors that previously existed, go for it.

PR #284, opened 2018-05-09
Election boundary export

russss:
This adds a management command to statically export election boundary JSON files for each election date, intended for use in web maps and similar applications:

`python ./manage.py export_boundaries --from 2018-01-01`

This produces three files for each election date (in `every_election/static/exports` by default), for example:

* `local.2018-05-03.json` (GeoJSON, 44MB)
* `local.2018-05-03-topo.json` (TopoJSON, 24MB)
* `local.2018-05-03-topo-simplified.json` (TopoJSON with simplified geometry, 5.8MB)

(Note that the TopoJSON file sizes here are affected by the offset issue in #283, because boundaries between neighbouring areas can't be shared. The simplified topojson should be less than 2MB without those issues, which makes it reasonably acceptable to load in to web maps.)

The [topojson](https://github.com/topojson/topojson) node.js command line tools are required to be installed - although there are some ports of the topojson logic to Python, they all appear to be quite old and unsupported, and I decided it was probably better to shell out to the maintained tools.

The `id` attribute of the features is set to the election ID, and name metadata is provided in the `properties` field.

This probably wants to be deployed so that the JSON files are served directly by the web server rather than through Django, so I haven't added any code to do that. (They'll also need CORS headers if we want to allow maps on other sites to use them.)

It would be good to make these files discoverable though the web interface (and possibly API) as well, but I think that's a job for another pull request.

### DIFF
--- a/every_election/apps/api/management/commands/export_boundaries.py
+++ b/every_election/apps/api/management/commands/export_boundaries.py
@@ -31,12 +31,10 @@ class Command(BaseCommand):
         output_dir = os.path.join(settings.BASE_DIR, 'static', 'exports')
         parser.add_argument('--from', dest='from',
                             help='Export elections from this date',
-                            type=parse_date,
-                            default=datetime.now().date())
+                            type=parse_date)
         parser.add_argument('--to', dest='to',
                             help='Export elections until this date',
-                            type=parse_date,
-                            default=(datetime.now() + timedelta(days=180)).date())
+                            type=parse_date)
         parser.add_argument('--output', dest='output',
                             help='Output directory (default every_election/static/exports)',
                             default=output_dir)
@@ -47,9 +45,16 @@ class Command(BaseCommand):
         except FileExistsError:
             pass
 
-        elections = Election.objects.all().filter(group_type='election')\
-            .filter(poll_open_date__gte=options['from'])\
-            .filter(poll_open_date__lte=options['to'])
+        if not (options['from'] or options['to']):
+            elections = Election.objects.future().filter(group_type='election')
+        else:
+            elections = Election.objects.all().filter(group_type='election')
+
+            if options['from']:
+                elections = elections.filter(poll_open_date__gte=options['from'])
+
+            if options['to']:
+                elections = elections.filter(poll_open_date__lte=options['to'])
 
         for election in elections:
             self.stdout.write("Exporting elections for group %s" % election)
@@ -77,31 +82,32 @@ class Command(BaseCommand):
     def export_election(self, parent):
         " Return GeoJSON containing all leaf elections below this parent "
         features = []
-        elections = self.get_leaf_elections(parent)
+        elections = self.get_ballots(parent)
         for election in elections:
-            if not election.geography:
+            if election.geography:
+                gj = geojson.loads(election.geography.geography.json)
+
+                # Round coordinates to 6 decimal places (~10cm) precision to reduce
+                # output size. This is probably as good as the source data accuracy.
+                gj['coordinates'] = set_precision(gj['coordinates'], 6)
+
+            else:
                 self.stderr.write("Election %s has no geography" % election)
-                continue
-
-            gj = geojson.loads(election.geography.geography.json)
-
-            # Round coordinates to 6 decimal places (~10cm) precision to reduce
-            # output size. This is probably as good as the source data accuracy.
-            gj['coordinates'] = set_precision(gj['coordinates'], 6)
+                gj = None
 
             feat = geojson.Feature(geometry=gj,
                                    id=election.election_id,
                                    properties={
                                         'name': election.election_title,
-                                        'division': election.division.name,
+                                        'division': election.division.name if election.division else None,
                                         'organisation': election.organisation.official_name
                                    })
             features.append(feat)
         return geojson.FeatureCollection(features,
                                          election_group=parent.election_id)
 
-    def get_leaf_elections(self, group):
-        " Return the leaf-level elections for a group. "
+    def get_ballots(self, group):
+        " Return the leaf-level ballots for a group of elections. "
         to_visit = [group]
         leaf_nodes = []
         while len(to_visit) > 0:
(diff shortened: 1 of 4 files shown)


chris48s:
[Review 2x EE PRs](https://trello.com/c/ilf6KXJh/1347-review-2x-ee-prs)

every_election/apps/api/management/commands/export_boundaries.py
```diff
@@ -0,0 +1,115 @@
...
+                                   properties={
+                                        'name': election.election_title,
+                                        'division': election.division.name,
+                                        'organisation': election.organisation.official_name
+                                   })
+            features.append(feat)
+        return geojson.FeatureCollection(features,
+                                         election_group=parent.election_id)
+
+    def get_leaf_elections(self, group):
```
chris48s: I think we've settled on calling a 'leaf election' a 'ballot'. Could you change this to `get_ballots()`?

every_election/apps/api/management/commands/export_boundaries.py
```diff
@@ -0,0 +1,115 @@
...
+        subprocess.check_call(['toposimplify', '-S', '0.2', '-F', '-o', dest, source])
+
+    def export_election(self, parent):
+        " Return GeoJSON containing all leaf elections below this parent "
+        features = []
+        elections = self.get_leaf_elections(parent)
+        for election in elections:
+            if not election.geography:
+                self.stderr.write("Election %s has no geography" % election)
+                continue
```
chris48s: In the case where one or more ballots have a `null` geography, this will skip the offending division and continue rendering an incomplete layer for the election group. I can see reasons for going both ways, but unless you've got a strong case for working with an incomplete layer I think on balance if we hit this condition, I would prefer to abort rendering the whole geo/topojson layer for this election group (but still move on to the next election group).

Also, given this is likely to run on a schedule and we usually configure our scheduled tasks to only output on error, if one or more election groups trigger this condition, I think it would probably be better for this command to (eventually) exit with a non-zero status code (even if some election groups render successfully). This will sometimes trigger some unwanted notifications because we do sometimes have to create an election id for an area with a `null` geography. However, in most cases if this path is taken, it represents an error we want a notification about rather than a condition that we should skip silently.

every_election/apps/api/management/commands/export_boundaries.py
```diff
@@ -0,0 +1,115 @@
...
+
+            gj = geojson.loads(election.geography.geography.json)
+
+            # Round coordinates to 6 decimal places (~10cm) precision to reduce
```
chris48s: :+1: for [RFC 7946 recommendation](https://tools.ietf.org/html/rfc7946#section-11.2)

every_election/apps/api/management/commands/export_boundaries.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def handle(self, *args, **options):
+        try:
+            os.mkdir(options['output'])
+        except FileExistsError:
+            pass
+
+        elections = Election.objects.all().filter(group_type='election')\
+            .filter(poll_open_date__gte=options['from'])\
+            .filter(poll_open_date__lte=options['to'])
```
chris48s: I expect the way we will usually want to run this will be to schedule it to run against all future elections. Can we make `--from` and `--to` optional and make the default behaviour here just to call `.future()`: https://github.com/DemocracyClub/EveryElection/blob/63de48d4b25aab39053fbaad6115bc6241d686eb/every_election/apps/elections/managers.py#L34-L35

every_election/apps/api/management/commands/export_boundaries.py
```diff
@@ -0,0 +1,115 @@
...
+
+            # Round coordinates to 6 decimal places (~10cm) precision to reduce
+            # output size. This is probably as good as the source data accuracy.
+            gj['coordinates'] = set_precision(gj['coordinates'], 6)
+
+            feat = geojson.Feature(geometry=gj,
+                                   id=election.election_id,
+                                   properties={
+                                        'name': election.election_title,
+                                        'division': election.division.name,
```
chris48s: This line of code throws `AttributeError: 'NoneType' object has no attribute 'name'` on Mayor and Police & Crime Commissioner elections because they don't use divisions.

every_election/apps/api/management/commands/export_boundaries.py
symroe: If one use case is election results, wouldn't `current` be a better queryset here?

chris48s: I'm not envisaging we delete existing files when we sync to S3. If the election has already happened, you would hope we've been able to generate a valid layer by that point in time.

every_election/apps/api/management/commands/export_boundaries.py
russss: My thought here is that if we're missing the geometry for a single election (out of, potentially, hundreds), it's probably still worthwhile to render the rest. Although in this case maybe a better solution would be to put it in the geojson with a null geometry.

Happy to add an exit code (although maybe it would be better to check for missing geometries with a separate process?)

symroe: I don't have strong feelings either way, but I think it's more honest to present as much as we can (or know) to the user – that is, we know of an election but not a geography. Omitting the election might give the impression we don't know about it, meaning the user would have to check for themselves what's going on.

chris48s: :+1: Lets go with that then: always include all the ballots ids and assign the object a `null` geometry property if we don't have one (yet)

russss:
OK, I think the above commit should fix all the suggested changes.

chris48s:
Thanks for your additional work on this :)

I've merged this, but given there is a bit of additional faffing I need to do in order to get it properly deployed (scheduled and writing data somewhere useful), there will be some delay before its actually available in production.

For my own reference, I will need to:

* Install `npm` and run `npm install` in ansible scripts
* Set up a S3 bucket and temp location
* Schedule:
  * `export_boundaries`
  * Sync files to S3
  * Clean up temp files
* Run `export_boundaries` for past elections and sync to S3 (maybe its best to sort out the issues with non-tesselating boundaries in #283 first!)
* Keep an eye on resource usage

I'll keep you informed on that. Thanks again

PR #303, opened 2018-05-18
allow seats_total and seats_contested to be edited via /admin

### LINKED ISSUE #276: Allow seats_total and seats_contested to be edited via /admin

chris48s:
closes #276 

Note we need `blank=True` otherwise we can't edit elections where we don't want to set this (e.g: it makes no sense to set this on `local.2018-05-03`). A more sophisticated approach would be to only make this editable for certain election types, but given the audience for `/admin` is basically 2 people I think we can probably trust ourselves.


symroe:
Nice one!

Issue #15, opened 2017-02-28
Add start and end dates to Organisations


symroe:
All good things must end, including government organisations :'(

We should make sure that we don't offer old organisations on the list of organisations in the ID creator when we know the election date, and validate that we don't have elections for organisations that have ended.

chris48s:
This would cover the situation where (for example) several district councils merge together into a new larger unitary authority (i.e: the actual _organisation_ changes). We need to handle boundary changes that don't change the organisation itself separately.

chris48s:
There's 4 rough jobs here:

1. Add the cols to the database
2. Decide how we're going to populate them moving forwards
3. Retrospectively attach dates to areas that already exist in our DB
4. Use those dates to actually do something useful in the id creator wizard

chris48s:
Closed by #315

Issue #98, opened 2017-10-27
Move some fields to the Geography model


chris48s:
This is kind of gnarly, but it is better to think about it now rather than having to deal with it in a rush when there is a boundary change we can't cope with.

## Problem 1:
We've keyed our organisations using gov.uk register IDs, which means that the organisation identifier stays stable across boundary changes. This is useful, but our current data model incorrectly:

* Associates a GSS code with an organisation record and
* Only allows us to attach one boundary to an organisation.

To do this correctly, a single organisation record should be able to have N boundaries attached to it and the GSS code should be associated with the boundary not the organisation (because that is what it actually describes). While an organisation should be able to have N boundaries associated with it, it should only be allowable for one of those to be active/relevant at any one time (i.e: they shouldn't be able to have overlapping dates).

## Problem 2:
Additionally, divisions (wards, constituencies, whatevz) are subject to boundary changes and there are 2 types of boundary changes we care about:

* Boundary changes that occur because of an Electoral Change Order (if this happens, the areas affected have a by-election)
* Boundary changes that occur because of practical considerations (if this happens, the areas affected *don't* have a by-election)

GSS codes don't care about the reason for a boundary change - if its a different polygon, it gets a new GSS code, but we want to think about this differently: In the case where a boundary change occurs due to an ECO, we want to treat this as the old ward/constituency/whatevz being destroyed and a new one being created. In the case where the boundary change is practical, we want to treat this as a single conceptual division what has had N boundaries at different points in time.

Currently divisions are keyed by GSS code (or a proxy id if it doesn't exist yet, but assume that we should be back-filling this). Again, I think this is incorrect for our use case because every time the boundary changes the code changes, but we only want to consider these divisions to be new/different if the changes happen as a result of an ECO.

Start/end dates are attached to the division *set*.

## Solutions

I'm fairly clear on the direction we need to go with this (I'm also going to pull in issue #15 at this point because I think it is easier to address that at the same time here).

* Add start/end dates to the Organisations table
* Move GSS code off the Divisions/Organisations tables and on to the Geographies table
* Also give Geographies a start/end date

This will allow us to record that either different orgs/div have existed at different points in time, or that a single org/div has had N geographic boundaries associated with it at different points in time.

We also need to think about how to correctly enforce the constraints on overlapping start/end dates. I'm not sure there is a neat way to do that on the DB layer - it might need to happen at the application level?

[...]

Issue #290, opened 2018-05-15
format_geography_link should not link to mapit if we don't have a GSS code


chris48s:
(or possibly at all)

chris48s:
Fixed in #315

Issue #183, opened 2018-02-12
How do we deal with versioning content from registers


chris48s:
At the moment we don't, but we need to.. In order for an identifier to be reproducible, we need to know the version of the register record it is derived from. Registers are versioned at a record level: https://local-authority-eng.register.gov.uk/record/LND/entries
At the moment we don't even have a mechanism to store this and update it over time, but we need to not only do that but also define how to derive the relevant register revision to use for an election as part of the identifier spec.

PR #315, opened 2018-05-31
Revise data model for Organisations

chris48s:
Done so far:
* Allow Organisations to have a `start_date` and `end_date`
* Use Organisation `start_date` and `end_date` in wizard/builder
* Delete broken code

High-level tasks left to do (some in this PR, some maybe not):
* Allow multiple `Organisation` records to have the same identifier as long as start/end date are different: Org becomes unique on (id, start_date)
* Move GSS code off `Organisation` model on to Geography (use correct boundaries based on election date)
* Set sensible start/end dates on existing Organisations (we can use registers for local auths, but there are other orgs too - tooling to import this?)
* Set additional DB constraints e.g: divset's `start_date` can't be before related org's `start_date` (we prob need to have sensible start dates set on everything before we can actually enforce these though)
* Tools to maintain Orgs going forward (probably make them editable though `/admin`)
* Use these features to account for recent changes in local gov

Refs https://github.com/DemocracyClub/EveryElection/issues/15
Refs https://github.com/DemocracyClub/EveryElection/issues/98


chris48s:
Right.. spent the day working on adding the `OrganisationName` feature we discussed. Its not finished, but I've pushed what I've done so far and written up my notes on what is left to do below. I'm aware there are still failing tests. @symroe - maybe have a look through the diff/remaining work. Hopefully this is sufficiently fleshed out you can see where I'm trying to go with it.. Also if you can spot anything I'm doing/planning here which could be done in a less cumbersome way, I'm all ears. This approach does seem to be adding a lot of complexity to support a very uncommon edge case, but I'm not sure allowing multiple `Organisation`records with the same GDS register code is necessarily better. Might be a bit of an "out of the frying pan and into another almost identical frying pan" situation :( We can catch up next week..

In order to finish off the work on adding an `OrganisationName` model, I will nedd to:

* Fix API serializer code (prob need 2 different serializer classes). I think the way to go on this is:
  * If we're serializing an `Organisation` in isolation, just pick the latest names
  * If we're serializing an `Organisation` in an election object, pick the names that are applicable to the `poll_open_date`
  * Present it as an flattened `Organisation` object for backwards-compatibility with YNR
* Can we ensure an `Organisation` object has at least one related `OrganisationName` object, or at least streamline object creation so it is harder to create an `Organisation` which has no related `OrganisationName`s?
* Sort out Organisation ordering in wizard. I've set the default to `official_identifier` because the default ordering has to be an actual DB field on that table, but now we need to sort them by the name value corresponding to the selected `poll_open_date` somewhere in the python code before we render the page.
* I've replaced all the occurrences of `.slug`, `.official_name`, `.common_name`, `.election_name` but there are probably also places where er are using `.name` and this is not a sensible thing to do anymore (e.g: Organisation list in wizard). This is harder to grep for because loads of stuff has a `.name` property.
* Fix factory test code
* Write additional test cases for `ElectionBuilder` ensuring it spits out correct slugs/throws correct errors in cases where there is not a one-to-one mapping between `Organisation` and `OrganisationName`

These next 2 sort of conflict.. we probably can't exactly do both of them at the same time. Need to find a tradeoff:
* Apply constraints/validation:
  * Should not be possible for an `Organisation` to have names with overlapping start/end dates
  * Shouldn't be possible for `OrganisationName` to have a `start_date` before related `Organisation`'s or `end_date` after related `Organisation`'s
* In 99% of cases there is a one-to-one mapping between `Organisation` and `OrganisationName`. Maybe it would be better to allow `OrganisationName.start_date` to be `NULL`, only bother to set dates on names where there is >1 and just assume that `Organisation.start_date` and `Organisation.end_date` apply to `OrganisationName` unless otherwise specified. 936df46 helps with this but I think more work needed.

also, ideally:
* Make object creation less verbose in unit test code (can we add helpers or make more use of existing factories)
(maybe not in this PR though)

chris48s:
Right, fc1caa6 --> 704b6db is now 'chunk 2' of this (unless I rebase. I'll try to avoid doing that to keep this coherent). I think it makes sense to review that as a block.

This basically makes Organisation an entity with a composite key `('organisation_type', 'official_identifier', 'start_date')` and updates everything else around it that was assuming `official_identifier` was unique to account for that.

Meanwhile I'll crack on with moving GSS off the `Organisation` model (which again builds on what I've done so far in this PR) and push that up as a third chunk.

chris48s:
Right, fc1caa6 --> 2f8dc49 is now 'chunk 2' of this (unless I rebase. I'll try to avoid doing that **again** to keep this coherent). I think it makes sense to review that as a block.

This basically makes Organisation an entity with a composite key `('organisation_type', 'official_identifier', 'start_date')` and updates everything else around it that was assuming `official_identifier` was unique to account for that.

Meanwhile I'll crack on with moving GSS off the `Organisation` model (which again builds on what I've done so far in this PR) and push that up as a third chunk.

chris48s:
Right, this seems like a good place to take a step back from this.. having got to here I think this closes:

#15 - One of the main things we're doing here.
#56 - Does moving the GSS to the geography and the mods to `format_geography_link()` (x2) solve this?
#57 - I've deleted it all, so.. meh.
#98 - Again, One of the main things we're doing here. Not exactly as described, but the problems identified in that issue are solved by this work.
#182 - Making `official_identifier` not unique fixes this problem/makes it irrelevant.
#290 - Mods to `format_geography_link()` close this.

Deploy notes and remaining TODOs:

* We need a way to import/create/modify Organisation records (and the related objects: `ElectedRole`, `OrganisationGeography`, etc). I think this looks like a combination of enabling stuff in `/admin` and management commands. I don't think we can import from registers anymore because we break the assumption that there is a 1-to-1 mapping of register records to records in our Organisations table (also there are Organisations which are not in the registers), but maybe this is also a good time to think about https://github.com/DemocracyClub/EveryElection/issues/183 . We need to do this before we can do anything useful with this, but we could deploy without it and it should be a seperate PR, just for the sake of sanity.

* I've created a migration which assigns every organisation a safe default `start_date`, but given that `start_date` is going to become part of the URL schema and API, I don't think we should go live with that because it means we're going to change the URLs for everything as we populate the start dates. Also we're probably going to have `DivisionSets` in the DB which start before the start date of the parent org which will cause problems. Before we deploy this we need to make some attempt to fill in valid start dates for all the organisations (for the local auths we can use the registers. For everything else, we need to do some research). This involves 2 jobs:
  * Work out what the start dates are (source data/research problem)
  * Devise a method to import/attach them (tooling issue)

This is a blocker to deployment IMO.

* Now that we are exposing the organisation type in the URL schema, we should review the descriptions we're using here. For example, at the moment we have `police_area` (with an underscore) but `local-authority` (with a dash). This isn't a huge issue when we're only using them internally, but once these become part of the URL schema, it becomes more important that these are consistent and harder to change them, so lets review them before we go live with this and make sure we are definitely happy with them before they become hard to change. Again this is a blocker to deployment, but we can have a quick chat and quickly solve this.

* There are a number of non-reversible data migrations in this PR. I've tested them against a recent DB dump, but just before deploying I need to grab another DB dump and test the migrations against that again.

comic_book_guy

TODO:
* boil ocean
* kitchen sink

(sorry)

every_election/apps/organisations/models.py
```diff
@@ -75,6 +75,14 @@ def format_geography_link(self):
         )
 
 
+class OrganisationGeography(models.Model):
+    organisation = models.ForeignKey(Organisation, related_name='geographies')
+    start_date = models.DateField(null=True)
+    end_date = models.DateField(null=True)
+    gss = models.CharField(blank=True, max_length=20)
+    legislation_url = models.CharField(blank=True, max_length=500, null=True)
```
symroe: Why is `legislation_url` on the geography?

every_election/apps/organisations/models.py
```diff
@@ -83,6 +83,19 @@ class OrganisationGeography(models.Model):
     legislation_url = models.CharField(blank=True, max_length=500, null=True)
     geography = models.MultiPolygonField()
 
+    class Meta:
+        ordering = ('-start_date',)
+        get_latest_by = 'start_date'
+        unique_together = ('organisation', 'start_date')
+        unique_together = ('organisation', 'end_date')
```
symroe: This is wrong in the same way as the other instance. Should be a tuple of tuples.

every_election/apps/organisations/models.py
```diff
@@ -74,6 +74,21 @@ def format_geography_link(self):
...
+            return None
+        elif len(self.geographies.all()) == 1:
+            return self.geographies.all()[0]
+        else:
+            geogs = self.geographies.filter(
+                models.Q(start_date__lte=date) | models.Q(start_date=None)
+            ).filter(
+                models.Q(end_date__gte=date) | models.Q(end_date=None)
+            )
+            if len(geogs) != 1:
```
symroe: Same `.exists` comment here as before.

every_election/apps/organisations/models.py
```diff
@@ -83,6 +108,16 @@ def get_geography(self, date):
         elif len(self.geographies.all()) == 1:
             return self.geographies.all()[0]
         else:
+            if date < self.start_date:
+                raise ValueError(
+                    'date %s is before organisation start_date (%s)' %\
```
symroe: I bet you one womble that this will happen legitimately one day.

every_election/apps/organisations/views.py
```diff
@@ -17,14 +19,31 @@ def get_context_data(self, **kwargs):
...
+            organisation_type=kwargs['organisation_type']
+        ).filter(
+            official_identifier=kwargs['official_identifier']
+        ).filter(
+            start_date__lte=kwargs['start_date']
+        ).filter(
+            models.Q(end_date__gte=kwargs['start_date'])
+            | models.Q(end_date=None)
+        )
+        if len(orgs) != 1:
```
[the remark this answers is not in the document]
chris48s: They check 2 different things - `orgs.exists()` would returns true if `len(orgs) == 2`, but its useful that you've encouraged me to re-read this code (and the other similar block) and think about it a bit more.. See f3bbc98 .

every_election/apps/organisations/models.py
chris48s: I added it so that we can attach the relevant legislation e.g: https://www.legislation.gov.uk/ssi/2017/430/note/made if there is a boundary change.

In many cases though, we also want to to be able to attach a legislative order to the organisation itself. We don't currently have this, but it makes sense to add it there as well in this PR so I have added it in 818e490 .

every_election/apps/organisations/models.py
chris48s: I am sensible enough not to take that bet, but I reckon going with this assumption is sensible until we have an example that breaks it so we can understand why :)

every_election/apps/organisations/models.py
```diff
@@ -30,7 +30,9 @@ def name(self):
         return self.official_name or self.common_name or self.official_identifier
 
     class Meta:
-        ordering = ('official_name',)
+        ordering = ('official_name', '-start_date')
+        unique_together = ('official_identifier', 'organisation_type', 'start_date')
+        unique_together = ('official_identifier', 'organisation_type', 'end_date')
```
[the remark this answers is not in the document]
chris48s: yep - good shout (x2). Fixed in a6aa392

every_election/apps/api/urls.py
```diff
@@ -5,12 +5,31 @@
...
-router.register(r'elections', ElectionViewSet)
-router.register(r'election_types', ElectionTypeViewSet)
-router.register(r'election_subtypes', ElectionSubTypeViewSet)
-router.register(r'organisations', OrganisationViewSet)
+
+class OrganisationRouter(routers.DefaultRouter):
+
+    include_root_view = False
+
+    def get_lookup_regex(self, viewset, lookup_prefix=''):
```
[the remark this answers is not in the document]
chris48s: Mmngh probably, sort of.. I've slightly forgotten all of the variations of this I ended up trying before deciding this was the least bad, but I'm not 100% happy with it.

The way I've done this, DRF spits out the following routes:

```
^api/organisations/$ [name='organisation-list']
^api/organisations\.(?P<format>[a-z0-9]+)/?$ [name='organisation-list']
^api/organisations/(?P<pk>[^/.]+)/$ [name='organisation-detail']
^api/organisations/(?P<pk>[^/.]+)\.(?P<format>[a-z0-9]+)/?$ [name='organisation-detail']
^api/organisations/(?P<pk>[^/.]+)/geo/$ [name='organisation-geo']
^api/organisations/(?P<pk>[^/.]+)/geo\.(?P<format>[a-z0-9]+)/?$ [name='organisation-geo']
^api/organisations/(?P<organisation_type>[-\w]+)/(?P<official_identifier>[-\w]+)/(?P<date>\d{4}-\d{2}-\d{2})/$ [name='organisation-detail']
^api/organisations/(?P<organisation_type>[-\w]+)/(?P<official_identifier>[-\w]+)/(?P<date>\d{4}-\d{2}-\d{2})\.(?P<format>[a-z0-9]+)/?$ [name='organisation-detail']
^api/organisations/(?P<organisation_type>[-\w]+)/(?P<official_identifier>[-\w]+)/(?P<date>\d{4}-\d{2}-\d{2})/geo/$ [name='organisation-geo']
^api/organisations/(?P<organisation_type>[-\w]+)/(?P<official_identifier>[-\w]+)/(?P<date>\d{4}-\d{2}-\d{2})/geo\.(?P<format>[a-z0-9]+)/?$ [name='organisation-geo']
^api/organisations/(?P<organisation_type>[-\w]+)/(?P<official_identifier>[-\w]+)/$
^api/organisations/(?P<organisation_type>[-\w]+)/(?P<official_identifier>[-\w]+)\.(?P<format>[a-z0-9]+)/?$ 
```

This means it _supports_ (variations of) `/organisations/{organisation_type}/{official_identifier}/{date}` and `/organisations/{organisation_type}/{official_identifier}` but all the rest of the DRF gubbins is still there, so it will link to `/organisations/` (organisation-list) from the API root page and it considers `/organisations/{pk}` (with a numeric id) to be the canonical URL for the resource, and all the rest of it. I expect the way I've done this may break some stuff like `HyperlinkedModelSerializer`s, or it will work but always link to `/organisations/{pk}` (although we don't currently have any links in the API - its all just serialized flat).

I don't think we can convince DRF that `/organisations/{organisation_type}/{official_identifier}/{date}` is the canonical URL to using out-of-the-box functionality, but if we want to link models using URLs other than the ones DRF wants to give us, it is possible to bash square pegs into a round rest framework e.g:

https://github.com/DemocracyClub/UK-Polling-Stations/blob/6d71c8800c0c87785a1fa3ba876da4fd230c5ff8/polling_stations/apps/api/pollingstations.py#L17-L30

I'm very open to suggestions for how we can do this better though. There may be important things I am missing. Now I've got as far as writing this, I think I am going to spend a bit more time on it again and see if I can come up with a better approach..

every_election/apps/api/views.py
```diff
@@ -81,15 +81,21 @@ class OrganisationViewSet(viewsets.ReadOnlyModelViewSet):
     @detail_route(url_path='geo')
     def geo(self, request, **kwargs):
         kwargs.pop('format', None)
-        org = Organisation.objects.all().get(**kwargs)
```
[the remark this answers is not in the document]
chris48s: When you're doing a review you can go back and delete comments before submitting a review if you later realise the comment is not relevant

chris48s:
Right.. Been looking at the DRF stuff. Has definitely been beneficial to revisit it in the context of your comments and not having not worked on it for ~a week and semi-forgotten it. I've spotted some improvements that can be made (which I haven't pushed yet). Quick (maybe not actually quick) question...

First one is on URL schema:

I share your preference that if `/organisation/{org-type}/{id}/{date}` is a thing, then subsets of that should also work, so ideally we want:

```
/organisation/{org-type}              returns [array]
/organisation/{org-type}/{id}         returns [array]
/organisation/{org-type}/{id}/{date}  returns {object}
```

There's a second conflicting "requirement" though. If you remember, we also agreed that because at the moment the links to organisations are like

https://elections.democracyclub.org.uk/organisations/HAS/
https://elections.democracyclub.org.uk/organisations/cambridgeshire-and-peterborough/
https://elections.democracyclub.org.uk/api/organisations/HAS/
https://elections.democracyclub.org.uk/api/organisations/cambridgeshire-and-peterborough/

we would attempt to be backwards-compatible where possible. So on the front-end I've set it up like this:

https://github.com/chris48s/EveryElection/blob/f3bbc98f68249a6a7a1684190fd548052656aad6/every_election/apps/organisations/urls.py#L17-L34

https://github.com/chris48s/EveryElection/blob/f3bbc98f68249a6a7a1684190fd548052656aad6/every_election/apps/organisations/views.py#L34-L49

so what we've basically got there is:

```
/organisation/{id}                    might return an {object}, or just fall over
/organisation/{org-type}/{id}         returns [array]
/organisation/{org-type}/{id}/{date}  returns {object}
```

I reckon the API and the front-end should work the same way: we should do the same thing in both places. Now I'm looking at this again, in retrospect, I think its actually better to break backwards-compatibility here, go with the first option in both places and create a more consistent URL schema moving forwards (accepting that we're going to break links). I also think we probably shouldn't try to be clever and support `/organisation/{id|org-type}` and magically work it out. I reckon that is one of those design decisions that we'll probably regret down the line..

Opinions?

symroe:
Yep, I agree that we should break URLs now and now support legacy URLs in future. This project is basically still in alpha, and it's only public because it's better to be public than private.

It feels like once this PR is in (plus maybe a few more bits of work) we can call it "beta" and start worrying about breaking things, but also start publicising it a little more.

Issue #293, opened 2018-05-16
Report elections with no geometry


chris48s:
As noted in https://github.com/DemocracyClub/EveryElection/pull/284#discussion_r188529699
Sometimes we may need to create these, but often/usually if they exist this indicates an error we should fix or a piece of work we need to keep sight of.

What's the best way to deliver this info? Scheduled management command? Notifications in `/admin` ?

chris48s:
Turns out there is already a handy `list_divisions_without_geography` command knocking about which we can use as a base. Maybe mod this so it exits with not-zero if it finds anything and run it daily..

chris48s:
I'm going to close this as https://github.com/DemocracyClub/EveryElection/pull/365 makes this information visible via /admin

PR #343, opened 2018-06-27
Manage Organisations via /admin

chris48s:
This PR allows us to create and manage the meta-data to handle changes to organisations, organisation names, boundary changes, etc affecting organisations via /admin. Hopefully it is a bit easier to review than the last one. There is slightly less code! :D

The one missing piece of this puzzle is that we also need a way to attach the actual boundary to an `OrganisationGeography` record. Having fiddled with things a bit, I have decided that the easiest way to manage this is to:

* manage all the meta-data (setting start/end dates etc) via /admin
* allow an `OrganisationGeography` record to be initially created with a NULL geography
* do #293 so that if we leave any without boundaries, we'll get some kind of notification
* create a management command to import a single boundary from BoundaryLine or similar (I'll look at this next, but will probs do it as a standalone PR)

I've added quite a lot of lines of code here and no tests, but it is mostly 'declarative' stuff and is not end-user facing so I'm inclined not to spend a lot of time trying to unit test it. If there is anything I've added here which you think would benefit from having tests, let me know in review.

TODO: Since we're going to do this infrequently and it the process is _slightly_ fiddly (e.g: if you create a new `Organisation` but forget to set up any `ElectedRole`s for it, you won't be able to create any elections), I'll add some docs to the wiki as well.

### DIFF
--- a/every_election/apps/organisations/management/commands/list_divisions_without_geography.py
+++ b/every_election/apps/organisations/management/commands/list_divisions_without_geography.py
@@ -10,28 +10,36 @@ from organisations.models import (
 
 class Command(BaseCommand):
 
+    def output_org(self, org):
+        self.stdout.write("\t {} {}, {}: {} ({})".format(
+            org.territory_code,
+            org.official_identifier,
+            org.start_date.isoformat(),
+            org,
+            org.official_identifier)
+        )
 
     def handle(self, **options):
         seen = set()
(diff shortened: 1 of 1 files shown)


symroe:
Great stuff! I reckon we might iterate the admin interfaces as we use them over time, but this is a great improvement :)

Issue #100, opened 2017-11-06
Better processes/tooling around boundary reviews


chris48s:
Refs "do boundaries properly"

1. We need a better record of when we last imported/modified boundaries from Boundary Commission reviews
2. We need a better record of when we last checked http://www.lgbce.org.uk/current-reviews (not necessarily the same as when we last imported - sometimes we'll look at this and decide we don't need to do anything)
3. We need a better way to identify when there is new content we need to review at http://www.lgbce.org.uk/current-reviews (scraper/boundary bot)

chris48s:
in progress, see https://github.com/DemocracyClub/boundary-bot/issues

chris48s:
Docs: Once I've cleared the backlog, as I work through the process of dealing with ECOs on a rolling basis, produce some documentation for how we handle this moving forwards.

chris48s:
Also to think about under this issue: how do we make our state reproducible?

First questions to answer there are probably 'how did we get to where we started from' and 'could we get there again if we wanted to?'

chris48s:
Also, do we want to retrospectively attach GSS codes to replace temp identifiers once boundaries imported through this process appear in Boundary Line. If so how?

Would doing so increase "reproducibility"?

If we _don't_ do this do we start to hit issues with non-unique identifiers in future. e.g: if we've curried `ABC:central` in one generation and there's also a "central" in the next generation, is it a problem to have `ABC:central` in 2 generations if we haven't backported the previous one to be `X01000001` yet?

chris48s:
>Also, do we want to retrospectively attach GSS codes to replace temp identifiers once boundaries imported through this process appear in Boundary Line

Answer: Yes we should. Not doing so makes our data harder to re-use because it is difficult to attach it to published boundaries once coded versions are available.

Also, we should consider providing some mechanism for users to get the boundaries via our API - at least in the cases where we are using the pre-Boundary Line releases as there is no other convenient source for them.

chris48s:
Interesting edge case to deal with when you look at back-porting codes onto boundaries which we received pre-publication:

Marple South/Marple South and High Lane - name changed but not the boundary at local level: http://democracy.stockport.gov.uk/mgConvert2PDF.aspx?ID=128900&ISATT=1#search=%22marple%20south%22 but no boundary change.

Code E05000795 needs to exist with 2 different names. Think that is fine as long as there are 2 `DivisionSet`s, but needs checking.

chris48s:
[...]

Issue #56, opened 2017-07-20
Handle official IDs better


symroe:
We have a `gss` field on `Organisation`, but in a lot of cases this isn't actually a GSS code (going forward we won't have GSS codes for anywhere at the point of import).

We should rename the field to something, but also support different ID schemes for one object.

chris48s:
maybe sensible to look at this with #98 ?

chris48s:
Does moving the GSS code on to the geography and the changes in https://github.com/DemocracyClub/EveryElection/commit/1997287f6a0a8933a341e3ca1de2ff27e10443cd in PR #315 solve this, or is there further work to do in order to close this?

PR #353, opened 2018-07-04
Attach codes from BoundaryLine to divisions imported from LGBCE

chris48s:
One of the things we need to do as part of our cycle of boundary maintenance is retrospectively attach codes to boundaries we've imported from LGBCE and assigned temporary identifiers. This PR adds a script which attempts to match up divisions in our database with temp codes to GSS codes or unit_ids in BoundaryLine and defines some of the structure that will house the next bits of BoundaryLine tooling.

Notes:

* I think it makes sense to group the BoundaryLine commands and helpers into a sub-app within organisations `organisations.boundaryline`. I might also move the existing code for dealing with boundary reviews into a `organisations.boundary_reviews` or `organisations.lgbce` sub-app, but I'll do that as a separate job rather than cluttering this PR with it.
* I've bacially taken [the simplest approach possible](https://github.com/DemocracyClub/EveryElection/commit/2670e30cf43795d9a82d67142be2d4f40c28d77c) to making sure we store official ids and can still refer to our temp ids. It would be nice to have a `DivisionCode` entity allowing us to store N codes against any `Division` object, but I think going off on that tangent now is going to be too much of a distraction at this point. Maybe that dovetails nicely with....
* I've decided to use `unit_ids` for CEDs because the main thing I'm concerned about at this stage is being able to tie our data to BoundaryLine but there may also be a secondary case for using http://geoportal.statistics.gov.uk/datasets?q=county%20electoral%20division%20names%20and%20codes&sort=name to also attach a GSS code to help others wanting to tie our data to other datasets. I'm not going to worry about this for the moment though. Ideally these codes will magically appear in BoundaryLine in future though, which would help.
* I haven't written any tests for this yet. I think I'm going to need to create some relatively realistic subset-of-BoundaryLine-fixture to run tests against, so I might wait until I've got more of this stuff written before I do that.
* I have realised that I've created migrations `0047`, `0048` etc in both this PR and #343 so whichever of these gets merged first I'm going to need to rebase and re-order the migrations. Oops.

Next Steps:

* Running this process will successfully attach codes to divisions in 99% of cases, but there will always be some handful of areas left over where we need to manually work out what to do (for various reasons, all of which are horrible). I think the most convenient way to deal with that 'mop-up' is to add some tooling to `/admin` which makes it possible to:
  * Easily identify Division records which don't have a suitable identifier
  * Manually edit Division records

  This is blocked on merge of #343 

* Once we've attached codes to `Division` objects, we also want a command (or some commands) to help us attach boundaries from BoundaryLine to `Division` objects. That will build on the work in this PR.

Closes #328
Refs #56
Refs #100


every_election/apps/core/mixins.py
```diff
@@ -41,13 +39,46 @@ def add_arguments(self, parser):
...
-    def read_local_csv(self, filename):
+    def read_from_local(self, filename):
+        return open(filename, 'rt')
+
+    def read_from_url(self, url):
+        tmp = tempfile.NamedTemporaryFile()
+        r = requests.get(url, stream=True)
+        r.raise_for_status()
+        for chunk in r.iter_content(chunk_size=1024):
+            if chunk:
```
symroe: Can you explain what this `if` statement is for?

every_election/apps/organisations/boundaryline/boundaryline.py
```diff
@@ -0,0 +1,54 @@
+from django.contrib.gis.gdal import DataSource, OGRGeometry
+from django.utils.text import slugify
+
+
+class BoundaryLine:
```
symroe: Naming bikesheding I know, but `import boundaryline.boundaryline.BoundaryLine` seems…a bit much?

Might be worth calling this file something related to what it does / is for?

Obviously this isn't super important, but it would be nice to reduce the duplication a little :)

every_election/apps/organisations/boundaryline/boundaryline.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __init__(self, filename):
+        ds = DataSource(filename)
+        if len(ds) != 1:
+            raise ValueError("Expected 1 layer, found %i" % (len(ds)))
+        self.layer = ds[0]
+
+    def normalize_name(self, name):
```
symroe: This feels like something that needs some documentation and tests :) Reading this, I don't really understand what it's for. Can you add a docstring at least?

If it's only used for matching, can you call it something like `normalize_name_for_matching` (or whatever) as on first reading this looked like you were changing names that we use.

every_election/apps/organisations/boundaryline/boundaryline.py
```diff
@@ -0,0 +1,54 @@
...
+            if self.normalize_name(feature.get('name')) == division_name:
+                match = self.get_code_from_feature(feature)
+                matches = matches + 1
+            if matches > 1:
+                # ...but we also need to be a little bit careful
+                print('Found >1 matches for division {div}'.format(
+                    div=div.official_identifier))
+                return None
+
+        return match
```
symroe: For YNR's ResultsBot 🤖 I used a [console based picker](https://github.com/DemocracyClub/yournextrepresentative/blob/master/ynr/apps/resultsbot/matchers/mappings.py#L7) that saved the mapping (in a file that could be committed or saved somewhere). It might be worth considering something like this in future, to save the work of manually matching on each environment we run this.

symroe:
> I haven't written any tests for this yet. I think I'm going to need to create some relatively realistic subset-of-BoundaryLine-fixture to run tests against, so I might wait until I've got more of this stuff written before I do that.

Fair enough – do you see that as part of this PR? Dropping 5% coverage isn't amazing, so we'll have to remember to come back and improve tests as a dedicated PR if you think it's best to wait.

every_election/apps/core/mixins.py
chris48s: Apparently we needed this to filter out keep-alives, but tbh I just cargo-culted it from the docs: http://docs.python-requests.org/en/latest/user/advanced/#streaming-requests

However, re-reading this with fresh eyes, I realise that `requests` provides a poor abstraction over this and we can reduce this whole block to a single line by using the standard library, so I've just [done that instead](https://github.com/DemocracyClub/EveryElection/pull/353/commits/af1962129d038abb8bf573810feb18ed5b666717) :D

every_election/apps/organisations/boundaryline/boundaryline.py
chris48s: Yep from `organisations.boundaryline.boundaryline import BoundaryLine` is a bit much. An option to give it a cleaner import signature would be to import the class in the app's `__init__.py` so the import can reduce to `organisations.boundaryline import BoundaryLine`

every_election/apps/organisations/boundaryline/boundaryline.py
chris48s: Clarified in d52fa81 . I've split out and added some tests for some bits that doesn't require a BoundaryLine 'fixture' to test it.

every_election/apps/organisations/boundaryline/boundaryline.py
chris48s: I think the level of 'investigation' needed into each case where a match can't be found makes an interactive management command a bad choice in this case. I think I want it to spit out a list so I can go and puzzle over them rather than stop for each one and ask me for a decision now.

In terms of replicating the work in multiple environments..

The way I'm envisaging this process is:

* Run this
* It updates the DB in 99% of cases
* Work out what has happened in the other 1% of cases, work out how to fix them + fill in the gaps in `/admin`
* Now we've got codes against everything, move on to importing boundaries

but another equally valid process would be:

* Run this
* It doesn't change the DB, but spits out a file with [temp id --> proper id] mappings in 99% of cases and placeholders in the other 1%
* Work out what has happened in the other 1% of cases, work out how to fix them + fill in the placeholders in the file
* Import the file to update the DB
* Now we've got codes against everything, move on to importing boundaries

On the one hand: you do have a good point that if we can generate a mapping locally, we overcome the issue that unzipping BoundaryLine and reading it into memory isn't going to go well on a `t2.micro`. If we split this into 2 processes (one to generate a [temp id --> proper id] mapping file and a second to import it), we could generate the file in one place and run the more lightweight import on a micro instance. Also we can run a slow thing less times.

On the other hand: after doing this, the next thing we need to do (although I haven't wirrten the code for it yet) will read BoundaryLine in and import boundaries from it, so we're still going to need a beefier instance to do that. As such, we're still going to need to run some subset of this suite of commands from a higher spec instance so I'm not sure it gives us much gain from that perspective. Running a slow thing fewer times is still advantageous though.

Using either method we will be able to tie our temp ids to the offficial ids we attach later.

Running this process over 2 years' worth of boundary review backlog on my machine takes about 6 minutes.

This is a process we would expect to perform once per year as we move forwards, usually processing one year's worth of backlog.

Does that additional info on the tradeoffs involved here clarify your thinking on this?

chris48s:
> do you see that as part of this PR?
> we'll have to remember to come back and improve tests as a dedicated PR

My intention is definitely to write tests for all this. I see adding tests for this code happening in a follow-up PR which adds additional management commands to the `boundaryline` app. I think I've got a higher chance of testing the right things and spending less time on it if I do it that way.

My preference for the next sequence of events would be that we merge this PR so we can get the data model changes/migrations into `master` and then I can do the work on `/admin` which depends on those (although depending on your options about [this](https://github.com/DemocracyClub/EveryElection/pull/353#discussion_r200171698), maybe that becomes less important?) and further work on BoundaryLine management commands as separate PRs (probably one for `/admin` stuff and another one or two for further BoundaryLine stuff).

There are other options though:

* I can write tests for this next instead of moving on to the scripts to import boundaries from BoundaryLine.
* I could split this into 2 PRs: one which makes data model changes which can be merged ahead of this, rebase this onto it, continue adding BoundaryLine stuff to this PR and block the merge pending tests. Then I can do the `/admin` stuff separately but based on top of the migrations which get merged separately.
* We could do this as another 'long running branch' like #315 where we review it in stages. The advantage we have here (which was not true for #315) is that if we merge this, it doesn't really change anything user-facing. If I deploy this today, the impact is an basically that an additional management command (which doesn't have tests yet) exists.

Let me know which you prefer.

chris48s:
Rather than attempt an elaborate rebase manoeuvre I'm just going to merge this and the next chapter will end up making some further changes to this.

PR #354, opened 2018-07-11
import_boundaries command

chris48s:
Think this is good for review now. Sorry this is another slightly messy one :(

Hopefully the commit messages document this reasonably, but at a high level..

* Add a source field to boundaries so we know where they came from and pre-populate with sensible values. This will be useful as we work through re-importing boundaries that are shifted, simplified or otherwise dodgy as we'll know which ones we've updated and which ones we haven't.
* Add a command to allow us to to import boundaries from BoundaryLine onto `Divisions` and `Organisations`. This is very 'dumb' and just relies on us feeding it a list of codes - I've not made any attempt to write code to work out what should or shouldn't go into that list of codes in this PR.
* Various bits of refactoring and moving stuff about
* A few fixes and tweaks to `backport_codes` command added in #353
* Add tests for `backport_codes` and `import_boundaries` commands

There are going to be some edge cases on this which I haven't properly considered here, but I think rather than trying to solve them all in the abstract, I want to actually start using this and see what we hit.

I think the next steps are:

* Submit another PR making minimal modifications to `/admin` allowing us to mop up codes that `backport_codes` doesn't pick up
* Get this deployed (possibly look at some of the staging deploy stuff first so this step doesn't become "make prod servers fall over" ;) )
* Start using what I've done so far to _actually solve some problems_ in the real world
* As I'm doing that:
    * Work out what edge cases we're going to hit that I haven't already thought of (+ how to account for them)
    * Work out what checks/reports/etc we want moving forwards to stay on top of this (and help with building up the lists of codes to feed into `import_boundaries` )


every_election/apps/organisations/boundaryline/boundaryline.py
```diff
@@ -15,6 +16,28 @@ def __init__(self, filename):
             raise ValueError("Expected 1 layer, found %i" % (len(ds)))
         self.layer = ds[0]
 
+    def get_feature_by_field(self, fieldname, code):
+        matches = 0
+        match = None
```
symroe: Of course this works, and it's a minor point, but wouldn't this be neater:
```python
matches = []
for […]:
    matches.append()
if matches:
    …
if len(matches) == 1:
    …
raise MultipleObjectsReturned
```

every_election/apps/organisations/boundaryline/management/commands/boundaryline_import_single_boundary.py
```diff
@@ -36,7 +36,10 @@ def get_record(self, identifier):
         code_type, code = split_code(identifier)
 
         if code_type == 'unit_id' and re.match(r'^\d+$', code):
-            return OrganisationDivision.objects.all().get(official_identifier=identifier)
+            # FIXME before 2021
+            error = "Importing boundaries from BoundaryLine against CEDs is " +\
+            "not yet available because unit_id is not a stable identifier"
```
symroe:
```python
send_mail('Please add GSS codes for {} to BoundaryLine'.format(identifier), ['feedback@http://ordnancesurvey.co.uk'],)
```

PR #375, opened 2018-08-03
upgrade to pytest 3.7

chris48s:
doing this should cause the build to fail

### DIFF
--- a/requirements/testing.txt
+++ b/requirements/testing.txt
@@ -3,7 +3,7 @@
 mock
 model-mommy
 tox
-pytest<=3.6.4
+pytest>=3.7.0
 pytest-django
 pytest-pep8
 pytest-flakes


chris48s:
This `RuntimeError: Model class every_election.apps.organisations.admin.division_problem.DivisionProblem doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.` error reproduces in `pytest==3.7.0` and `pytest==3.7.1` but didn't occur in any previous versions.

Tried a few things to sort it:

Setting an `app_label` on the models where they are ends up with some form of `RuntimeError: Conflicting 'divisionproblem' models in application 'organisations'` error due to a mix of relative and absolute import styles. Changing them all to absolute fixes that but then causes `django.contrib.admin.sites.AlreadyRegistered` because we effectively define a circular import in `organisations/admin/__init__.py`. (this feels like `nose` all over again)

Adding an `admin` sub-app to `INSTALLED_APPS` just causes the application to fail bootstrapping with `django.core.exceptions.AppRegistryNotReady: Apps aren't loaded yet.`. I didn't really expect this to work.

I think this probably all goes away if I move all the proxy models to `organisations/models.py` but I'd prefer to keep the admin-only model code separated from the app model code. All the tutorials I've read on defining proxy models for /admin seem to recommend defining them along with the /admin code, like I have done.

For now I'm just going to leave us pinned to the older version of `pytest` as I'm not sure what to try next.. Dunno if you've got any bright ideas on this @symroe  ?

chris48s:
Also possibly of relevance here:

https://github.com/pytest-dev/pytest/blob/master/CHANGELOG.rst#pytest-370-2018-07-30

although I have no idea which of those changes might be connected to this :(

symroe:
On a phone so no strong ideas now, but make sure there are no `__init__.py` files in the test directories. They shouldn't be packages.

chris48s:
moved to #474

PR #381, opened 2018-08-20
Staging deploy improvements

chris48s:
This PR does a couple of things relating to staging deploys:

* Use [django-basicauth](https://github.com/hirokiky/django-basicauth) to enable basic auth

    This plugin uses a middleware to negotiate basic auth with the browser without having to involve the web server. This is helpful for our use case because we can configure it in django instead of nginx. Config will look like:

```
BASICAUTH_DISABLE = False
BASICAUTH_REALM = 'Staging'
BASICAUTH_USERS = {
    'staging': 'staging'
}
```

I'll set it conditionally on AWS tags.

* Show a banner in the base template on staging based on the `SERVER_ENVIRONMENT` setting (see https://github.com/DemocracyClub/ee_deploy/pull/14 )

### DIFF
--- a/every_election/templates/base.html
+++ b/every_election/templates/base.html
@@ -14,9 +14,11 @@
 
 {% block top_banner %}
 
-{# <div class="callout warning"> #}
-{#   <p class="text-center">This is an alpha site, so please don't expect anything to work just yet! <a href="https://democracyclub.org.uk/contact/">Get in touch</a> if you'd like to know more.</p> #}
-{# </div> #}
+{% if SERVER_ENVIRONMENT == 'test' or SERVER_ENVIRONMENT == 'staging' %}
+<div class="callout warning">
+  <p class="text-center">This is a staging site.</p>
+</div>
+{% endif %}
 
 <nav class="top-bar">
     <div class="columns large-8 large-centered">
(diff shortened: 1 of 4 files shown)


every_election/templates/base.html
```diff
@@ -14,9 +14,11 @@
 
 {% block top_banner %}
 
-{# <div class="callout warning"> #}
-{#   <p class="text-center">This is an alpha site, so please don't expect anything to work just yet! <a href="https://democracyclub.org.uk/contact/">Get in touch</a> if you'd like to know more.</p> #}
-{# </div> #}
+{% if SERVER_ENVIRONMENT == 'test' or SERVER_ENVIRONMENT == 'staging' %}
+<div class="callout warning">
+  <p class="text-center">This is a staging site.</p>
```
symroe: For YNR I added a couple of things in the banner to help with debugging. See https://github.com/DemocracyClub/YNR_deploy/blob/master/webapp_settings/banner/staging.html for more. This requires the banner to be an HTML file that's optionally included. https://github.com/DemocracyClub/yournextrepresentative/blob/8eb364f23b92204dcb657d1586cb201bcb27f47e/ynr/templates/generic_base.html#L42

This is fine for now as it adds the basic scaffolding, but it might be worth moving to that model, so we know what version a staging image is of, etc

PR #399, opened 2018-09-14
check memory before importing ONSPD

### LINKED ISSUE #341: Check we have enough memory in import_onspd_remote
If we run `import_onspd_remote` on a server with low memory, it is likely to fail.

Can we add some check at the start of the script to warn if this is likely to happen and reduce the chance of failing in an unsafe fashion.

chris48s:
closes #341


chris48s:
Going to crack on and merge this one too. Would be useful to check this on staging instances while I'm working on deploy repo stuff. Easier to ask forgiveness than permission :see_no_evil:

Issue #391, opened 2018-08-30
log errors to sentry


chris48s:
It would be helpful if we set up raven and a sentry DSN for Every Election instead of sending the errors to email

chris48s:
done in https://github.com/DemocracyClub/ee_deploy/commit/5fdb0da76d652624e0df7101ad4a0753756bb5fa

PR #400, opened 2018-09-14
log a warning if we geocode from mapit

### LINKED ISSUE #339: Notify if we fall back to mapit
It should be relatively rare that we are failing to geocode from ONSPD and falling back to mapit, but we don't really have much visibility on this. Somewhere here:

https://github.com/DemocracyClub/EveryElection/blob/438621ebaeb8a2895d9f55d6b93f65fd8e64b524/every_election/apps/elections/query_helpers.py#L33-L47

we should log a notification that sends an email or sends something to Sentry so we can understand if/when this is happening.

chris48s:
closes #339
refs  #391

This doesn't do much other than give us log event to hook into, but if I set up sentry in prod, I'll be able create a sentry log handler and do something like

```py
'elections.query_helpers': {
    'level': 'WARNING',
    'handlers': ['sentry'],
    'propagate': False,
}
```

so we pipe that somewhere visible. (see https://docs.sentry.io/clients/python/integrations/django/#integration-with-logging )

I'll do that in the deploy scripts repo though as it is a production deployment concern as opposed to an application concern.

### DIFF
--- a/every_election/apps/elections/query_helpers.py
+++ b/every_election/apps/elections/query_helpers.py
@@ -1,10 +1,15 @@
 import abc
+import logging
 import requests
 
 from django.contrib.gis.geos import Point
 from django.core.exceptions import ObjectDoesNotExist
 
 from uk_geo_utils.geocoders import OnspdGeocoder
+
+
+logger = logging.getLogger(__name__)
+
 
 class PostcodeError(Exception):
     pass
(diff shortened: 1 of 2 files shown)


chris48s:
merging this so I can get working on deploy repo bits

chris48s:
see https://github.com/DemocracyClub/ee_deploy/commit/5fdb0da76d652624e0df7101ad4a0753756bb5fa

Issue #197, opened 2018-03-01
Set a boundary on UK parliament


chris48s:
Currently this org has a null geography. Either find a good quality composite boundary or make one (e.g: the union of all the constituencies).

chris48s:
First need to do this for NI see #411

chris48s:
You can complete this with the NI outline from https://github.com/DemocracyClub/EveryElection/pull/524 which you used for EU region

PR #411, opened 2018-09-26
Import Northern Ireland divisions and geographies

### LINKED ISSUE #364: Re-import Northern Ireland geographies
Some of the NI geographies have the same issue described in #283

screenshot at 2018-07-23 11 13 08

We've got several issues with Northern Ireland boundaries:

* All NI council (organisation) boundaries are offset
* NI Assembly (organisation) boundary is offset
* We don't hold any boundaries at all for NI council wards (none of the NI councils have an associated `DivisionSet`)

It looks like the boundaries for NI Westminster Constiruencies and NI Assembly constituencies are OK, but I'd like to do a clean import for them anyway.

Not quite sure where all of this came from, but none of this is in BoundaryLine because OS only covers GB so we're also going to need tools to import [open data from OSNI](http://osni-spatial-ni.opendata.arcgis.com/datasets?q=Boundaries&sort_by=updated_at) to fix these issues.

Because it requires a completely different set of tooling, I'm going to consider this a separate issue from the current BoundaryLine work.

We will need to fix these issues as the NI council and ward boundaries will be needed for the May 2019 NI local elections. As things stand we won't be able to create the election ids because we haven't even got ward names.

chris48s:
closes #364

Notes:
* We don't need to deal with wards at all in NI, so I haven't bothered. To model elections we only care about District Electoral Areas.
* We'll import those from https://docs.google.com/spreadsheet/ccc?key=18X4wYGGmJNbL9xwN-ALXPQYDAed4PN1M9xC0YiKZ-TM#gid=652550336 for now and worry about it again next time they change. I haven't attempted to do anything in this PR to deal with change moving forwards.
* I've been able to neatly sidestep [this issue](https://github.com/mysociety/mapit/blob/e46726d47d1a18c5be4999aa372d41101c38c91a/mapit_gb/management/commands/mapit_UK_import_osni.py#L50-L61) by just requesting the data from ArcGIS in GeoJSON format, which means ArcGIS deals with the SRID transform server-side and serves up srid=4326 in all cases (this is not the case if you use the shapefiles). Not sure if this option wasn't available ~2 years ago or there was some other tradeoff at work, but that's an issue we can conveniently outsource to ESRI here :)
* I started off intending to write a generic `osni_import_boundaries` with roughly the same interface as `boundaryline_import_boundaries` but realised that:
    * We don't really have a need to import a single boundary from OSNI because if stuff changes the whole country changes at once, which means a lot of the stuff we're dealing with in `boundaryline_import_boundaries` would be pointless to replicate.
    * Each of the 4 cases we need to deal with is subtly different *sigh*
    ..so I ended up just writing 4 bespoke scripts. As with the script for importing divisions, they're not very generalised (and I've been able to neatly ignore a bunch of nuance around start/end dates for Orgs/DivisionSets because we only have one of each entity so far). We'll probably need to address that in future, but this allows us to kick that can down the road for 4 or 5 years.
* Fascinating fact: Northern Ireland Assembly constituencies and Westminster constituencies have the exact same names and boundaries but each area has 5 representatives in the Assembly as opposed to 1 in Westminster (maybe you already knew this - I didn't). Because of this, OSNI don't publish 2 different datasets and ONS don't assign them distinct GSS codes (I assume this is because for _statistical_ purposes, the difference is moot - its the same name/polygon). Anyway, in mapit their data model doesn't allow 2 different areas to have the same GSS code, so they worked around it e.g:
https://mapit.mysociety.org/area/66124.html
https://mapit.mysociety.org/area/16987.html
Becuase we seeded our DB from mapit, we've also inherited that design decision. For example, our Belfast East NIA constituency has the (pseudo-)"GSS code" "osni_oid:NIE-9". I chewed this over a bit and for the moment I've just gone with that, but simultaneously our data model doesn't contain the assumption that a GSS code is unique (although this property is intended for use when we want a code to exist in multiple DivisionSets). In principle, we _could_ give them both the same GSS code. Simultaneously that doesn't mean its _actually a good idea_. Do you have any thoughts on the tradeoffs involved there?
* The other thing I was hoping to do in this PR was create a corrected composite boundary for Northern Ireland Assembly from the union of all the constituencies. I had a bash at that, but ended up with this:

    screenshot at 2018-09-26 16-16-55

    which is not quite as clean as I'd like. I think for now I'm going to park that one, delete the (offset) boundary we already have and come back to it under #197

Process for importing in production is going to be:

```
python manage.py ni_import_district_electoral_areas -u "https://docs.google.com/spreadsheet/ccc?key=18X4wYGGmJNbL9xwN-ALXPQYDAed4PN1M9xC0YiKZ-TM&output=csv&gid=652550336"
python manage.py osni_import_lgd
python manage.py osni_import_lge
python manage.py osni_import_nia
python manage.py osni_import_wmc
```

..and then manually delete NIA `OrganisationGeography` record via `/admin`


symroe:
Set up locally and all seems to work nicely. Not much else to say – good work!

Also `retry` looks handy, I've not seen that before!

chris48s:
Cool - thanks. I'll deploy and run it in production tomorrow :+1:

PR #418, opened 2018-10-02
Election status part 1 - soft delete

chris48s:
High-level tasks in this PR are:

* Add a 'deleted' status for elections
* Set all the elections that already exist to 'approved'
* Hide elections that aren't approved in most contexts
* Add a flag to include deleted elections in the API (so YNR can use it)

Functionality-wise, this just deals with the case where we want to (soft-)delete an election which has been added in error, but it also lays down how I'm going to approach the rest of the work around moderation so would be useful to get some feedback at this point. I quite like the approach of using 2 managers with different base query sets - it made changing the `/api` and `/admin` behaviour pretty trivial :)

I've taken on board Joe's suggestion that we should treat internal and external events separately and I'm going to treat "election is cancelled" flag as meta-data we store against an election (like the date, or a notice of election document. 'status' (in terms of events we expose to the outside world) can be treated as a value derived from meta-data which is independent of the moderation process. I think this will be a better way to handle it.


every_election/apps/elections/models.py
```diff
@@ -96,7 +96,9 @@ class Election(SuggestedByPublicMixin, models.Model):
     snooped_election = models.ForeignKey('election_snooper.SnoopedElection',
         null=True, blank=True, on_delete=models.SET_NULL)
 
-    objects = ElectionManager.as_manager()
+
+    public_objects = PublicElectionsManager()
+    private_objects = PrivateElectionsManager()
```
chris48s: Do you think `public_objects` and `private_objects` is sensible here. Would it be better to just expose `public_objects` as `objects`?

every_election/apps/api/views.py
```diff
@@ -27,20 +27,24 @@ class APICoordsException(APIException):
...
     def geo(self, request, election_id=None, format=None):
-        election = Election.objects.get(election_id=election_id)
+        election = self.get_queryset().get(election_id=election_id)
         return Response(ElectionGeoSerializer(election,
                         context={'request': request}).data)
 
     def get_queryset(self):
-        queryset = Election.objects.all()
+        queryset = Election.public_objects.all()
+        if self.request.query_params.get('deleted', None):
```
chris48s: I've added this filter here but I've not exposed a deleted flag or the moderation status directly via the API. Do you think we should do this? (bearing in mind we'll also have a separate 'status' based on real-world events)

every_election/apps/elections/managers.py
```diff
@@ -36,3 +36,47 @@ def current(self):
 
     def future(self):
         return self.filter(poll_open_date__gte=datetime.today())
+
+
+class BaseElectionManager(models.Manager):
```
symroe: Do you think using [`QuerySet.as_manager()`](https://docs.djangoproject.com/en/2.1/ref/models/querysets/#django.db.models.query.QuerySet.as_manager) here?

every_election/apps/suggested_content/models.py
```diff
@@ -1,20 +1,22 @@
...
+    rejected = Status("Rejected", "Rejected by a moderator")
+    approved = Status("Approved", "Approved by a moderator")
+    deleted = Status("Deleted", "Deleted (because it was added in error)")
 
 class SuggestedByPublicMixin(models.Model):
     suggested_status = models.CharField(
         max_length=255,
-        choices=[(x.name, x.value) for x in SuggestedStatuses],
-        default=SuggestedStatuses.suggested.value.lower(),
+        choices=[(x.name, x.value.short_title) for x in SuggestedStatuses],
```
symroe: I get that this is just adding to the existing statuses, but do you think we'll end up using a model for this, with a M2M between a set of statuses the election, with a through model that adds any extra info we want to store?

Is it worth moving to that model soon?

every_election/apps/elections/models.py
symroe: Oh sorry, I thought that a model had `objects` by default.

Ace, in that case I think being explicit it better, as you have here.

every_election/apps/api/views.py
symroe: Yeah, I reckon we should add it to the API

every_election/apps/suggested_content/models.py
symroe: Yeah, and notice documents, notes as to background etc. 

Happy to leave that sort of thing to a different PR of course, but it might be worth thinking about any impact it might have, or if it's even a good idea.

chris48s: In keeping with the objective of separating data that has meaning in the outside world from data that is part of our internal moderation process, I'd want to keep the notice (of election or cancellation) separated from the moderation events, but I think it probably does make sense to move to this model now. I suspect we'll end up doing it in the near future.

I can see a couple of variations on this approach. I'll look into it today.

Issue #180, opened 2018-02-08
Bulk id creation


chris48s:
User story: As a creator of (many) election ids, I would like to not have to do this

![gifrecord_2018-02-08_105807](https://user-images.githubusercontent.com/6025893/35969696-9bf0c016-0cbf-11e8-8786-b2eeb9e7c2bc.gif)

for days at a time when there are 1/3 up local elections.

PR #419, opened 2018-10-05
Convert moderation status to M2M

chris48s:
eurgh.. not quite ready for review yet :( Still a couple of bugs to work out :beetle: . Mainly putting this here so I can get a nice coverage report. Will pick up on Monday..


symroe:
This all broadly seems good. I don't want to get in to premature optimisation here, but I had a few thoughts while reviewing. Feel free to ignore or punt these for future:

1. Relying on a `post_save` signal might cause problems. Not just because they don't run in some cases, but because it can slow things down in unpredictable ways. I'm seeing this with YNR at the moment, where some views rack up 100s of queries because signals are firing off all over the place. I get they they perform a useful function (and that YNR creates more things that EE and is more complex in other ways) but it might be worth exploring how to deal with models that don't have a status created, and punting creation to later on (create when called, or something). Of course, this might in turn cause more problems…just something to think about. 

2. Do we care about performance enough to think about denormalizing the status, or some other method? Adding a join to every query might be a problem for us, especially when we know the majority case query (thinking about a large data user wanting to convert postcode to elections). Related, is it worth maintaining the `ModerationStatus` objects as objects in memory rather than the database?

chris48s:
> Relying on a post_save signal might cause problems. Not just because they don't run in some cases

Yep. This is a valid concern. Its not a pattern that I'm a massive fan of either and I try to avoid it where possible. My original approach to this was to create the related record directly inside the `Election.save()` method. This is how we automatically create group objects when we save a ballot id.

The reason I ended up switching it to a `post_save` hook is exactly because its useful that we can disable them in certain situations. In this case, it was problematic to always create a status record on save when creating test objects. Converting to a `post_save` hook solved this problem because it allowed me to decorate `ElectionFactory` with `@factory.django.mute_signals(signals.post_save)` meaning we can always have control to set up the right test cases. This seemed like a useful tradeoff here.

Looking at the docs, `post_save()` occurs inside the `@transaction.atomic` which applies to `save()`, so we're not loosing anything there. What are the other situations where hooks don't (or might not) run? It would be useful to understand that.

> but it might be worth exploring how to deal with models that don't have a status created

I did consider creating the record with no related status, but it added complexity to the code for no real payoff IMO.

> it can slow things down in unpredictable ways. I'm seeing this with YNR at the moment, where some views rack up 100s of queries because signals are firing off all over the place

In terms of write performance, I'm not deeply concerned about it. There are several reasons for this:

* In the 'off-season', we mostly create ~one record at a time. Even if that spawns 50 additional writes (I _think_ its now spawning between 2 and 9 depending on whether the parent groups already exist, but I'll confirm this.. see later), I don't see that as a big problem.
* When we do create hundreds (or thousands) of IDs at once, I see the main point of friction as data entry tedium (see https://github.com/DemocracyClub/EveryElection/issues/180 ) not DB write performance, but we can look at it when we think about improvements to bulk creation. If we get to addressing bulk creation and want to look at inserting in a more efficient way, the fact that we can turn off hooks is probably an _advantage_ in that situation too because we could disable them in a bulk import management command and do them more efficiently.
* Creating lots of IDs is best viewed as a batch process. I view it as more analogous to processing boundary updates rather than an interactive process.
* The cardinality of the amount of writes we need to do is fundamentally an order of magnitude smaller than YNR.
* YNR was running ~70 DB queries to create a single person and we only actually noticed it as a performance issue once we had 3 SOPN parties happening at once. EE is _way_ under that threshold and doesn't lend itself to that style of usage.

> Do we care about performance enough to think about denormalizing the status, or some other method? Adding a join to every query might be a problem for us

In terms of read performance, that is something that we do need to consider in principle because it potentially limits our ability to scale. Whereas the amount of writes we need to do in EE is bounded by the number of elections we need to create, EE API usage can scale with more users. From a practical perspective, we basically deal with that by cloning it into WhereDIV/WCIVF but if (when?) we start encouraging more usage of the core EE service from outside users we do ned to consider it.

From a practical perspective, I don't think I'm worried about introducing a PK `JOIN` between 2 tables with <100,000 rows in them here (or at least not enough to worry about optimising it yet). I avoid extra `JOIN`s like the plague in WDIV because there we're dealing with tables that have tens of millions of rows, or we're doing spatial `JOIN`s, but that's not the pattern we're dealing with here.

---

All of that said, all of these problems get worse one extra little query at a time, so it seems like a sensible thing to do here is introduce some assertions about DB performance/IO. This will give us a sort of `[number of straws:camel's back strength]` ratio we can monitor :) When I go back and look at those inline comments (both good points), I'll have a look at adding some `assertNumQueries()` tests to make sure we know what the performance looks like and ensure we monitor it moving forwards.

symroe:
> Looking at the docs, post_save() occurs inside the @transaction.atomic which applies to save(), so we're not loosing anything there. What are the other situations where hooks don't (or might not) run? It would be useful to understand that.

`bulk_create()` or anything that creates a model without calling `save()` (or, if we override the `save()` method and forget to call `super()` on it) would all be situations where the signal isn't called.

I can see us using `bulk_create` in the case you talk about with creating a lot of election IDs, but in that single case we can bulk create the status objects after (or do some other manual process for creating them).

I'm not very worried about this, especially given the points you make 👍

chris48s:
Right, yeah. I remember being bitten by this with `bulk_create` before. That applies even if we move it inside `save()` though: https://github.com/DemocracyClub/UK-Polling-Stations/commit/3aef3d3d0354c86faaf71a4dabd83a9cc06b5d15

Its probably necessary evil here, but in general I'm on board with you that its a pattern we should not make extensive use of if we can avoid it.

symroe:
Let's keep an eye on it. We addresses some of this thin-ice-style data model problems in YNR by running a SQL query on a cron that would error if something wasn't right. 

No real need to do that yet, but we always have that sort of option open to us.

chris48s:
Cheers for having a look. I was actually slightly premature in pushing earlier. Have added the performance assertions now :)

Interestingly, the `save()` is making a lot more queries than I predicted but looking at the logs of which queries are performed, because we're doing 3 nested transactions there are a lot of `SAVEPOINT` and `RELEASE SAVEPOINT` statements, each of which is a DB hit.

There is probably some scope to improve the API performance a bit with some `prefetch()` or `selected_related()` wizardry, but for now I reckon putting a number there so know if it changes is a good first move.. useful to know where we are with it anyway.

chris48s:
note to self: squash f8e6c1c and ce9dfb2 into 8921e4e before merge

symroe:
👍 Yep, checks look good – 19 is a lot, but I don't think it matters for the use profile of this project

PR #426, opened 2018-10-19
Election status part 2 - cancelled elections

chris48s:
First bash at allowing us to deal with cancelled elections:

* Data model changes to allow us to store that an election is cancelled, attach a document or notice and link 2 election records when one replaces another
* Basic front-end work to allow an admin to set the 'cancelled' flag and 'replaces' field in `/admin` (front-end for notice document is TODO)
* API changes so that client apps can pick up cancelled elections (this can be safely deployed without making changes to API consumers)
* Front-end changes for showing cancelled elections on the EE website:

screenshot at 2018-10-19 11-35-18

Next steps:

* Changes to consume this data in YNR, WCIVF, WDIV
* Front-end changes to enable a wider group of authenticated users to use some of the features added here and in the last few PRs


every_election/apps/elections/models.py
```diff
@@ -271,6 +287,16 @@ def get_organisation_geography(self):
...
 
+    def clean(self):
+        if self.group_type and self.cancelled:
+            raise ValidationError(
+                "Can't set a group to cancelled. Only a ballot can be cancelled"
+            )
+        if not self.cancelled and self.cancellation_notice:
+            raise ValidationError(
+                "Only a cancelled election can have a cancellation notice"
+            )
```
chris48s: Having added a couple of checks in here, it occurs to me that we've probably got various bits of validation in the wizard form class, `ElectionBuilder` class and `/admin`. When we add more front-end for authenticated users that problem will probably get even worse. I reckon it probably makes sense to try and collate more of that validation logic here in the model class, but I think I'll leave that to another PR to try and keep the scope of this one sensible.

every_election/apps/elections/models.py
```diff
@@ -149,12 +149,28 @@ def get_children(self, manager):
...
+        null=True, blank=True, related_name="_replaced_by")
+
+
+    @property
+    def replaced_by(self):
+        if len(self._replaced_by.all()) == 0:
+            return None
+        if len(self._replaced_by.all()) == 1:
+            return self._replaced_by.all()[0]
+        raise AttributeError('Election should only have one replacement')
```
symroe: If we make an election before an ECO, and the ECO merges two wards then we might want a one to many here. The good news is a) we're unlikely to do this as we prevent elections being made before we have division sets (although we might make a mistake) and b) the data model allows this, so we can always edit is assertion.

My question is: is it worth presenting this as a list in the API anyway? Or do we think it's such a rare case that we can just ignore it?

chris48s: In that case, I'd day its not a "replacement". If election A is cancelled and election B also exists and uses a different boundary, they're just.. 2 different elections.

symroe: Yeah, you're right. It actually falls in to the internal "deleted" category (in that it's a class of error caused by us, not the state)

PR #429, opened 2018-10-19
force 🔌pluggy 0.7.1 after installing pytest

chris48s:
The man from del travis - he say...


chris48s:
:tada: 

this is partly a symptom of being pinned to an old version if pytest. I should probably revisit #375

PR #432, opened 2018-10-23
remove some deprecated calls

chris48s:
This fixes the following deprecation warnings:

* `DeprecationWarning: detail_route is deprecated and will be removed in 3.10 in favor of action, which accepts a detail bool. Use @action(detail=True) instead`
* `RemovedInDjango20Warning: Specifying a namespace in django.conf.urls.include() without providing an app_name is deprecated. Set the app_name attribute in the included module, or pass a 2-tuple containing the list of patterns and app_name instead.`
* `DeprecationWarning: Please use assertRaisesRegex instead.`
* `RemovedInDjango20Warning: set_srid() is deprecated, use the srid property instead.`
* `DeprecationWarning: invalid escape sequence \.`

### DIFF
--- a/every_election/apps/api/views.py
+++ b/every_election/apps/api/views.py
@@ -30,7 +30,7 @@ class ElectionViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = Election.public_objects.all()
     serializer_class = ElectionSerializer
     lookup_field = 'election_id'
-    lookup_value_regex = "(?!\.json$)[^/]+"
+    lookup_value_regex = r"(?!\.json$)[^/]+"
     filter_fields = ('group_type', 'poll_open_date')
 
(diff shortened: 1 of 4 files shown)


every_election/apps/api/views.py
```diff
@@ -30,7 +30,7 @@ class ElectionViewSet(viewsets.ReadOnlyModelViewSet):
     queryset = Election.public_objects.all()
     serializer_class = ElectionSerializer
     lookup_field = 'election_id'
-    lookup_value_regex = "(?!\.json$)[^/]+"
+    lookup_value_regex = "(?!.json$)[^/]+"
```
symroe: I think this is also fixed with `r""` strings? Is that a batter pattern to use?

PR #451, opened 2018-12-05
Front-end for election moderation

chris48s:
Rough overview:

* Add front-end bumf for doing user-ey stuff: log in/log out/reset password/etc
* Add a moderators group
* Create elections as 'suggested' by default + send us a notification (so we know it needs moderation)
* Store user and notes on `ModerationHistory` events
* Automatically approve elections if they are suggested by moderators
* Front-end for approving/rejecting suggested elections
* Require user to be moderator to:
  * Approve/reject elections suggested by untrusted users
  * Set status of possible elections found by election radar scrapers
  * Upload notice of elections docs

Once we've got the review done on this there is also some work on comms to our (small) group of regular users about the change of process and creating accounts for them, etc.


every_election/templates/registration/login.html
```diff
@@ -0,0 +1,16 @@
...
+{% block content %}
+<section class="columns large-6 large-centered">
+  <div class="card postcode_card">
+    <h2>Login</h2>
+    <form method="post">
+      {% csrf_token %}
+      {{ form.as_p }}
+      <button class="button" type="submit">Login</button>
+    </form>
+    <a href="{% url "password_reset"%}">Forgot password?</a>
```
chris48s: If I request a password reset locally, it does this:

```
You're receiving this email because you requested a password reset for your user account at 127.0.0.1:8000.

bla bla...

The 127.0.0.1:8000 team
```

Do you happen to know if we need to set a `SITE_NAME` for this to work nicely when we deploy it, or will it just magically work it out from the request? If you don't know we can always deploy it and see - just wondering if you know up-front. Also we probably need to set up some mail transport settings over in https://github.com/DemocracyClub/ee_deploy/blob/master/webapp_settings/production.py for this to work properly as I don't think anything else in EE sends emails.

every_election/apps/elections/views/moderation_queue.py
```diff
@@ -0,0 +1,70 @@
...
+    event = ModerationHistory(
+        election=election,
+        status_id=status,
+        user=user,
+        notes='moderation queue'
+    )
+    event.save()
+
+
+class ModerationQueueView(UserPassesTestMixin, TemplateView):
```
chris48s: I wasn't quite sure where to put this. I've grouped the moderation queue for suggested election in with the 'election radar' stuff from a visual perspective because I like the idea that "suggested by anonymous user" is essentially another source like "ALDC" or "Notice of Election scraper" (except someone has already done the data entry). In terms of the code, I've put all the view code for this in the `elections` app because all the models it talks to (`Election`, `ModerationHistory`, `ModerationStatuses`) are (correctly) defined in the elections app and `election_snooper`/radar app is mostly about scraping. The other way to view it would be to put this view in `election_snooper` so the code is laid out closer to how we present the front-end. I could be convinced either way though. What do you think?

chris48s:
Hm. I wonder why coveralls didn't run on this. Maybe it will catch up at some point

every_election/apps/election_snooper/views.py
```diff
@@ -1,16 +1,22 @@
...
 from election_snooper.forms import ReviewElectionForm
 
 
-class SnoopedElectionView(TemplateView):
+class SnoopedElectionView(UserPassesTestMixin, TemplateView):
     template_name = "election_snooper/snooped_election_list.html"
 
+    def test_func(self):
+        group = Group.objects.get(name='moderators')
+        return group in self.request.user.groups.all()
```
symroe: I think it would be nicer to use `django.contrib.auth.models.Permissions` directly rather than groups. That is, create a `can_moderate_election` (or whatever) permission and then let the moderators group have that permission.

Then this method could be replaced by [`PermissionRequiredMixin`](https://ccbv.co.uk/projects/Django/2.0/django.contrib.auth.mixins/PermissionRequiredMixin/) and `user.has_perms()`.

This method also allows for finer permissions in future if we wanted to add them.

every_election/templates/registration/login.html
symroe: I think this is the Django template:

https://github.com/django/django/blob/master/django/contrib/admin/templates/registration/password_reset_email.html. That's using `get_current_site` from the sites framework, so yes will need to be set and then should work fine.

every_election/apps/elections/models.py
```diff
@@ -303,6 +304,8 @@ def clean(self):
     @transaction.atomic
     def save(self, *args, **kwargs):
         status = kwargs.pop('status', None)
+        user = kwargs.pop('user', None)
+        notes = kwargs.pop('notes', '')
```
symroe: `[:255]` here?

every_election/apps/elections/views/moderation_queue.py
symroe: I don't have strong feelings that should block this PR, but I think `election_snooper` might be better renamed to something that can contain all input or possible elections, and we merge (as you say) the concept of bots and humans when it comes to unverified creation of elections.

That turns one app in to a queue and the others can just focus on the business of elections.

every_election/apps/elections/views/id_creator.py
```diff
@@ -270,6 +271,13 @@ def done(self, form_list, **kwargs):
                 notes=notes
             )
 
+        if not user_is_moderator(self.request.user) and len(context['all_ids']) > 0:
+            message = """
+                New election suggested by anonymous user:\n
+                <https://elections.democracyclub.org.uk/moderation_queue/>
```
symroe: Can this include any more details, or even a hash fragment for the ID? If there's a long list it might be hard to know what the notification is about.

every_election/apps/election_snooper/views.py
chris48s: discussed offline: sticking with groups

every_election/apps/elections/views/id_creator.py
chris48s: Added a bit more info in de6b785

every_election/apps/elections/views/moderation_queue.py
chris48s: I've moved it all into the `election_snooper` app in adedf82
I'll raise an issue to rename the app another time

every_election/templates/registration/login.html
chris48s: :+1: Right, yep so if we don't have sites framework enabled (which we don't need because we only have one 'site'), it'll just infer it from the request object:

https://github.com/django/django/blob/stable/1.11.x/django/contrib/sites/shortcuts.py
https://github.com/django/django/blob/stable/1.11.x/django/contrib/sites/requests.py

chris48s:
> I think we should allow keen annon users to upload documents

Made notes on this based on earlier discussion in #460

PR #474, opened 2018-12-21
Update pytest/pluggy

chris48s:


symroe:
I guess that makes sense – are you thinking we could move the views and forms to the respective `views` and `forms` modules, or would you rather keep admin stuff somewhere else?

chris48s:
My original instinct was to keep all the admin app content grouped together and just move it to an app called `administrative` or `internal` or something (just to free up the name `admin`). Then we can have an `admin.py`. Maybe it would make sense to organise it into:

`/organisations/views/public`
`/organisations/views/admin`

`/organisations/forms/public`
`/organisations/forms/admin`

instead?

symroe:
`/organisations/views/admin` etc makes more sense to me (class types grouped together), but I get that it might be a little odd. 

I don't have very strong feelings about it, so maybe just do whatever you think it most sensible.

Issue #378, opened 2018-08-14
Make it more obvious that org types are links on /organisations


chris48s:
Because the links on https://elections.democracyclub.org.uk/organisations/ use the `<code>` style, they don't have a 'link' cursor when you hover over them. We should either:

* just use a normal text style
* apply some custom styling to make it more obvious they are links

PR #475, opened 2018-12-21
Fix styling of details>summary element

chris48s:
refs #378

It turns out the main reason I found this really confusing is because the base theme styles this element sensibly in Chromium-based browsers but not Firefox, and I use Firefox :fox_face: 

Firefox:

screenshot at 2018-12-21 16-00-24

Chromium Stable:

screenshot at 2018-12-21 16-00-36

I've proposed a hacky fix here mainly just so that I remember it, but we should probably fix this in the base theme, tag a patch release and upgrade EE. Happy to propose a PR to the theme if I can get a pointer on where in the v0.2 theme this should live.

### DIFF
--- a/every_election/apps/organisations/templates/organisations/supported_organisations.html
+++ b/every_election/apps/organisations/templates/organisations/supported_organisations.html
@@ -1,5 +1,14 @@
 {% extends "base.html" %}
 
+{% block extra_page_css %}
+<style>
+@-moz-document url-prefix() {
+  summary {
+    display: list-item;
+  }
+}
+</style>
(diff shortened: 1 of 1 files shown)


symroe:
I reckon a `_summary.scss` file in the `elements` folder (and imported in the [`_elements` file](https://github.com/DemocracyClub/dc_base_theme/blob/v0.2/dc_theme/static/dc_theme/scss/dc/_elements.scss)) would be the best place for this. 

Happy to merge this, will make an issue to remove it when it's in the DC theme.

PR #492, opened 2019-01-25
add management command for copying divisions from one DivSet to another

chris48s:
This is a process we'll need to go through again in future to account for minor/ 'technical' boundary changes or ward names where internally we need to model the change as a whole new DivisionSet but in reality most of the wards are staying the same.

TODO: once merged, update https://github.com/DemocracyClub/EveryElection/wiki/Copying-Geographies-from-one-DivisionSet-to-another

### DIFF
--- a/every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
+++ b/every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
@@ -6,9 +6,9 @@ The --all flag may also optionally be passed if a code exists in multiple
 DivisionSets and we want to import the boundary against all occurrences.
 
 Example calls:
-manage.py boundaryline_boundaries --code gss:W09000043 --source bdline_gb-2018-05 -f /foo/bar/bdline_gb-2018-05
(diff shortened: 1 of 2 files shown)


symroe:
Yep, this looks fine – couple of tiny points that you can ignore if you like:

1. If an ObjectDoesNotExist exception is raised on https://github.com/DemocracyClub/EveryElection/pull/492/files#diff-12ddbabbfda13ccb2d1f1d99a23fc586R19 or the line below it might not be _super_ obvious what line raised it (or rather, what ID passed in is wrong) or what the value was. Might it be nice to raise with a error "src not found" or something?

2. Is it worth counting the divisions in `dest` after? I can't really see why they wouldn't be the same, but this is the sort of area where checking everything twice is a good idea, to save confusing debugging later on.

chris48s:
rebased this, but did those 2 things:

> Might it be nice to raise with a error "src not found" or something?

https://github.com/DemocracyClub/EveryElection/pull/492/files#diff-cd205ed44395984ddf43a5456bac8da5R21

> Is it worth counting the divisions in dest after?

https://github.com/DemocracyClub/EveryElection/pull/492/files#diff-cd205ed44395984ddf43a5456bac8da5R56

Issue #340, opened 2018-06-27
Ensure importing ONSPD doesn't leave us in an inconsistent state


chris48s:
We need to import a new ONSPD on the prod database 4 times a year. This takes ~10 mins. While we are doing this, what state are we in? Do we exhibit spurious behaviour? Do some testing to work out the current behaviour and make improvements if necessary. This may require some tweaks to the import scripts in https://github.com/DemocracyClub/uk-geo-utils to make them behave more safely when running in a live environment.

chris48s:
Gave this a quick bash:

As things are at the moment, importing against a 'live' DB will leave the API responsive while its happening but requests will fall back to the MS mapit.

If we do [this](https://github.com/DemocracyClub/uk-geo-utils/commit/03d0f6973d025b7e2182933db52ceb083756492b) (I've not decided whether to package that yet) then basically the table is read-locked while the import takes place so all the API requests will probably time out and serve a `502 Bad Gateway` while the import happens and then it will come back. The rest of the site will stay responsive though, and we're always in a consistent state.

Not sure which is the lesser of those 2 evils tbh :thinking:

chris48s:
Did some more thinking/testing. Decisions:

* I've added the ability to import in a transaction as a non-default option in uk-geo-utils: https://github.com/DemocracyClub/uk-geo-utils/commit/c7bd24a46b526ca077cd46076d7550a944a6874b This might be a useful feature anyway.
* I'm not going to change anything here. If a row doesn't exist, we just fall back mySociety's mapit. Given in ONSPD a postcode maps to a single row (unlike in AddressBase where we need to query multiple rows for a single postcode query) and the row is is either there or not there, I think its OK to import the table "live" and query mySociety's mapit while we're importing as a fall-back behaviour.

chris48s:
disabled mapit fallback in 0fa3261d674aa244b7a1a9fdf65365425709a602

revisit how to deal with the window while we're importing ONSPD before updating to May ONSPD release

/remind me ping on 2019-05-13

Issue #544, opened 2019-05-29
Import ONS Postcode Directory (May 2019)


polling-bot-4000:
@chris48s - New ONS Postcode Directory (May 2019) available at http://geoportal.statistics.gov.uk/datasets/5194b0c9d5a042789058609a4435ec7c

PR #558, opened 2019-06-20
zero downtime ONSPD import

chris48s:
Quick background on this:

* Importing ONSPD takes ~10 mins, or whatever.
* If you do the import in a transaction, it locks the table for the full 10 mins and every API request with a postcode issues a 503 while you're doing it and that has a knock-on effect on WDIV, WCIF, etc.
* If you don't do it in a transaction, you can get spurious/incorrect results during that 10 minute window.
* Now that we've got rid of the code that queries mySociety's mapit, we can't use that as a fallback anymore (also, rate limits, etc).

At a high-level, this PR solves the problem by:
* Creating a temp table
* Importing the data into the temp table
* Swapping the tables over
(its slightly more fiddly than that)

With this approach, we stay queryable throughout the process.
Once I've run this in production a few times and I'm happy with the process, I'll probably generalise and upstream this to the uk-geo-utils package but for now I think its easier for it to live here. In general, its probably a useful thing to have worked out a technique for.

Closes https://github.com/DemocracyClub/EveryElection/issues/340
Closes https://trello.com/c/6hk83eIr
Closes https://trello.com/c/ks9LfU3K
Refs https://github.com/DemocracyClub/EveryElection/issues/544

### DIFF
--- a/every_election/apps/core/management/commands/import_onspd_remote.py
+++ b/every_election/apps/core/management/commands/import_onspd_remote.py
@@ -3,12 +3,22 @@ import psutil
 import shutil
 import tempfile
 import urllib.request
+import sqlparse
 from django.core.management.base import BaseCommand
+from django.db import connection
+from django.db import transaction
 from storage.zipfile import unzip
 from uk_geo_utils.management.commands.import_onspd import Command as LocalImporter
+from uk_geo_utils.helpers import get_onspd_model
 
 
 class Command(BaseCommand):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.cursor = connection.cursor()
+        self.table_name = get_onspd_model()._meta.db_table
+        self.temp_table_name = self.table_name + "_temp"
+
     def add_arguments(self, parser):
         parser.add_argument("url", action="store")
 
@@ -23,6 +33,68 @@ class Command(BaseCommand):
         gb = ((mem.total / 1024) / 1024) / 1024
         return gb >= 2
 
+    def get_index_statements(self):
+        self.cursor.execute(
+            "SELECT indexdef FROM pg_indexes WHERE tablename='%s' ORDER BY indexname LIKE '%%_pk' DESC;"
+            % (self.table_name)
+        )
+        results = self.cursor.fetchall()
+
+        indexes = []
+        statements = [row[0] for row in results]
+        for statement in statements:
+            indexes.append(
+                {
+                    "original_index_create_statement": statement,
+                    "temp_index_create_statement": "",
+                    "index_rename_statement": "",
+                }
+            )
+
+        for index in indexes:
+            index["temp_index_create_statement"] = index[
+                "original_index_create_statement"
+            ].replace(self.table_name, self.temp_table_name)
+
+            # split the SQL string into a list of sqlparse.sql.Token objects
+            # https://sqlparse.readthedocs.io/en/latest/analyzing/
+            parsed_sql = sqlparse.parse(index["original_index_create_statement"])[0]
+
+            # we expect the statement to be of the form
+            # CREATE [UNIQUE] INDEX $index ON $table USING $fields"
+            # attempt to extract $index and $table
+            identifiers = [
+                token.value
+                for token in parsed_sql.tokens
+                if not token.ttype and self.table_name in token.value
+            ]
+            if len(identifiers) != 2:
+                raise Exception("Expected 2 identifiers, found %i" % len(identifiers))
+
+            original_index_name = identifiers[0]
+            temp_index_name = original_index_name.replace(
+                self.table_name, self.temp_table_name
+            )
+            index["index_rename_statement"] = "ALTER INDEX %s RENAME TO %s" % (
+                temp_index_name,
+                original_index_name,
+            )
+
+        return indexes
+
+    def create_temp_table(self):
+        self.cursor.execute("DROP TABLE IF EXISTS %s;" % (self.temp_table_name))
+        self.cursor.execute(
+            "CREATE TABLE %s AS SELECT * FROM %s LIMIT 0;"
+            % (self.temp_table_name, self.table_name)
+        )
+
+    def swap_tables(self):
+        self.cursor.execute("DROP TABLE %s;" % (self.table_name))
+        self.cursor.execute(
+            "ALTER TABLE %s RENAME TO %s;" % (self.temp_table_name, self.table_name)
+        )
+
     def handle(self, **options):
         if not self.check_memory():
             raise Exception(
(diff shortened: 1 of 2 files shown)


chris48s:
Cheers coveralls. I didn't quite get the message the first 175 times :D

every_election/apps/core/management/commands/import_onspd_remote.py
```diff
@@ -35,11 +65,40 @@ def handle(self, **options):
...
+            # (e.g: uk_geo_utils_onspd_pcds_9d376544_uniq )
+            # so we can still run migrations and stuff on it
+            index_create_statements = self.get_index_create_statements(table_name)
+
+            self.stdout.write("Swapping tables..")
+            self.swap_tables(table_name, temp_table_name)
+
+            # create the indexes outside of the transaction block
+            # this will mean we block queries for the absolute minimum time
+            # the table will be queryable (but a bit slower) while the indexes rebuild
```
symroe: Why not build the indexes on the new table and rename them to the expected names? That would mean there's no slow down on queries that require indexes, and rebuilding indexes can generally have a big performance impact.

every_election/apps/core/management/commands/import_onspd_remote.py
```diff
@@ -35,11 +65,40 @@ def handle(self, **options):
...
+            cmd.table_name = temp_table_name
+            cmd.path = data_path
+            cmd.import_onspd()
+
+            # grab the index CREATE statements from the old table before
+            # we drop it. This will ensure we create the indexes on the
+            # new table with the exact names django expects them to have
+            # (e.g: uk_geo_utils_onspd_pcds_9d376544_uniq )
+            # so we can still run migrations and stuff on it
+            index_create_statements = self.get_index_create_statements(table_name)
```
symroe: I think this is reasonable, but did you see [Django has named indexes](https://docs.djangoproject.com/en/1.11/ref/models/indexes/), with a lot more options for index type etc? There might be value in moving to use them / using the index classes to re-create the table in future.

Not for this PR, but worth having opinions about if you don't already

every_election/apps/core/management/commands/import_onspd_remote.py
```diff
@@ -35,11 +65,40 @@ def handle(self, **options):
...
+            # create the indexes outside of the transaction block
+            # this will mean we block queries for the absolute minimum time
+            # the table will be queryable (but a bit slower) while the indexes rebuild
+            self.stdout.write("Building indexes..")
+            cursor = connection.cursor()
+            for statement in index_create_statements:
+                cursor.execute(statement)
         finally:
+            self.stdout.write("Cleaning up temp files..")
             self.cleanup(tempdir)
```
symroe: Worth a `DROP TABLE IF EXISTS` here too? I know it's done before the temp table creation, but I presumable that's because you think it might have been left in an unclean state from a previous run?

every_election/apps/core/management/commands/import_onspd_remote.py
symroe: This might get easier with the named indexes, as above

every_election/apps/core/management/commands/import_onspd_remote.py
chris48s: Indeed it does and I did look at whether that helps us here as my first approach. It doesn't seem to be possible to make your primary key index named or use an `Index` object to infer the name of a non-named index though.

Also I've used the name `index_create_statements` here, but more accurately this should really be `index_and_constraint_create_statements`. Named constraints aren't available in django 1.11, although once we upgrade to 2.2, we do also get named constraints: https://docs.djangoproject.com/en/2.2/ref/models/constraints/ so until then we do still need to rely on inspection of `pg_indexes`.

Might be worth having another look in the context of 2.2, as that does seem to expose some of the internals for working out what the constraint/index names should be.

every_election/apps/core/management/commands/import_onspd_remote.py
chris48s: Yep. I was considering that. I'd definitely have done it that way if building the indexes locked the table, took ages or slowed down queries by a lot, but testing locally it seems like an acceptable tradeoff to do it this way because you can just grab the SQL and run it. It saves having to parse the SQL statements from `pg_indexes` out into bits, change them, stick them back together, etc which is a bit more error prone/likely to break in future.
That said, I've not seen how that translates to doing it on RDS. Maybe I'll give it a go on staging first before deciding on that..

every_election/apps/core/management/commands/import_onspd_remote.py
chris48s: Yeah exactly. If something really unexpected happens (I'm thinking of an out of memory error or the front-end server shuts down partway through the process or something like that), the `finally` clause would get skipped too, so I reckon its still worth having the `DROP TABLE` call in `create_temp_table()`, but I've duplicated it in the `finally` clause in e21cc97 . It won't do any harm.

every_election/apps/core/management/commands/import_onspd_remote.py
chris48s: Having tested it on staging, I think you're right and we do need to build the indexes on the temp table and then rename them.

chris48s:
I've rejigged this to build the indexes on the temp table first and rename them when we swap the tables over, tested it on staging and its a big improvement :+1:  I've pushed extra commits so its easier to see what's changed, but I'll probably squash all this into one commit once you've had another look.

every_election/apps/core/management/commands/import_onspd_remote.py
```diff
@@ -26,14 +27,45 @@ def check_memory(self):
...
+                "original_index_create_statement"
+            ].replace(table_name, temp_table_name)
+            parsed_sql = sqlparse.parse(index["original_index_create_statement"])[0]
+            identifiers = [
+                token.value
+                for token in parsed_sql.tokens
+                if not token.ttype and table_name in token.value
+            ]
+            if len(identifiers) != 2:
+                raise Exception("Expected 2 identifiers, found %i" % len(identifiers))
```
symroe: I think this could do with some documentation, as I don't really understand what the tokens or identifiers are. What are the cases where we throw this exception, for example?

chris48s: I've added a couple of comments in a6d0d5f which hopefully makes that clearer.

The reason for the exception is if we add some constraint in future that generates a statement which doesn't match the pattern we're expecting, I want that to fail in a controlled way as opposed to doing something unexpected.

every_election/apps/core/management/commands/import_onspd_remote.py
```diff
@@ -45,14 +77,8 @@ def create_temp_table(self, table_name, temp_table_name):
...
         cursor = connection.cursor()
-        # drop the old table and swap in the new one
-        # do these 2 in a transaction so if it fails
-        # we don't leave ourselves in an inconsistent state
-        with transaction.atomic():
-            cursor.execute("DROP TABLE %s;" % (table_name))
-            cursor.execute(
-                "ALTER TABLE %s RENAME TO %s;" % (temp_table_name, table_name)
-            )
+        cursor.execute("DROP TABLE %s;" % (table_name))
```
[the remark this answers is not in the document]
chris48s: In a parametrized query, parameters are only for supplying values. Using a parameter would quote/escape the input, which we do want for values, but don't want for a table/column name.

For example:
`cursor.execute("SELECT * FROM table WHERE field=%s", ['foo'])`
is desirable, but
`cursor.execute("SELECT * FROM table WHERE %s='foo'", ['field'])`
doesn't execute
```sql
SELECT * FROM table WHERE field='foo'
```
it executes
```sql
SELECT * FROM table WHERE 'field'='foo'
```
or to simplify that:
```sql
SELECT * FROM table WHERE False
```
They aren't equivalent.

In the `DROP TABLE` example, you'd just throw a syntax error because you can't drop a string literal.

There is no escaping that makes user-entered input "safe" to pass as a table or column (or index, or whatever) name into an SQL statement.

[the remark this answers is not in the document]
symroe: Ah, of course :) 🙄

PR #562, opened 2019-07-03
remove fields we don't need anymore

chris48s:
This PR completely removes the `mapit_generation_*` fields which aren't used anywhere and the `geography_curie` field, which is now just an alias for `official_identifier`.

I debated removing `tmp_election_id` here too, but I decided against it for now. At the moment there is no way to populate it, but we talked again recently about the concept of wanting to store information about an election which is expected to happen but not totally confirmed yet. I can see that maybe that feature could be resurface at some point, so I'll just stick with stripping it out in devs.DC for now.

We want to deploy https://github.com/DemocracyClub/yournextrepresentative/pull/981 ahead of this so that the sync doesn't break.


chris48s:
:+1: renamed to `0057_remove_mapit_fields.py`

PR #565, opened 2019-07-05
add identifier_type property

chris48s:
Following on from my last comment in https://github.com/DemocracyClub/aggregator-api/pull/56#issue-292868216

> Finally, one thing I realise as I think about documenting the API outputs is that the way we identify a ballot is not very nice. At the moment `"group_type": null` is the thing that tells you if an election is a ballot or a group. As an outsider, it feels like a bit of a strange way to expose quite a key bit of info though. What do you think of changing that to `"group_type": "ballot"` or adding an extra key like `"ballot": true/false`?

I wanted to add an `is_ballot` property to the API. Inevitably once I got into it, some worms came out  of the can.. They were small worms though.

I think it makes more sense to characterise this as `"is_ballot": true/false` as opposed to `"group_type": "ballot"` but I'll leave this one open for feedback.

This might also possibly allow us to slightly simplify the EE import code in YNR, but I've not checked it..

### DIFF
--- a/every_election/apps/api/management/commands/export_boundaries.py
+++ b/every_election/apps/api/management/commands/export_boundaries.py
@@ -134,13 +134,13 @@ class Command(BaseCommand):
         return geojson.FeatureCollection(features, election_group=parent.election_id)
 
     def get_ballots(self, group):
-        " Return the leaf-level ballots for a group of elections. "
+        " Return the ballots for a group of elections. "
         to_visit = [group]
         leaf_nodes = []
         while len(to_visit) > 0:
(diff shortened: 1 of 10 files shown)


chris48s:
Have updated this to do the opposite thing and expose an `identifier_type` property + corresponding filter.

There's a bit of future work to fully get rid of `group_type` internally and replace it with `identifier_type`, but I'm going to punt the rest of that to an issue.

I've updated https://github.com/DemocracyClub/aggregator-api/pull/63 to match.

PR #597, opened 2019-09-20
Re-define "current"

chris48s:


every_election/settings/base.py
```diff
@@ -226,6 +226,13 @@
 BASICAUTH_DISABLE = True
 
 
+# elections where polling day is in the range
+# (NOW - CURRENT_PAST_DAYS) - (NOW + CURRENT_FUTURE_DAYS)
+# are considered "current"
+CURRENT_PAST_DAYS = 20
+CURRENT_FUTURE_DAYS = 90
```
chris48s: I'm not wedded to 90 here. It seems like a reasonable number, but we can change it if we want (also doing this makes it an assumption we can revisit in future)

symroe: In this commit you've also removed:

`TODO replace this with a current status of the election model`

Is this because you think that's not a good idea?

I'm asking because, I think 90 days is a reasonable default for most elections we deal with (90 days from now is Christmas, that feels like a bit of a way off, and that's all the science I need), but if we ever get an actual scheduled general election ever again then it's likely to be too small a number. We'd almost want 6 months for it. 

Of course, that makes some assumptions about how we use the `current` status (that this batch of PRs is working on), but it might be nice to have a think about setting this differently for some elections. 

Maybe a `CURRENT_FOR_ADDITIONAL_FUTURE_DAYS` property on the model, with a number that gets added to whatever `CURRENT_FUTURE_DAYS`?

I expect we don't need to worry about this at the moment as a scheduled general election isn't coming along any time soon.

Issue #566, opened 2019-07-08
Discover elections by organisation


chris48s:
If I'm looking at a page like

https://elections.democracyclub.org.uk/organisations/local-authority/SCA/1974-04-01/

it would be really useful if it linked to:

* http://elections.democracyclub.org.uk/elections/local.south-cambridgeshire.2016-05-05
* http://elections.democracyclub.org.uk/elections/local.south-cambridgeshire.2017-05-04
* http://elections.democracyclub.org.uk/elections/local.south-cambridgeshire.2018-05-03

chris48s:
done in #647

PR #647, opened 2020-01-14
Adds an elections section to the organiation_detail template.

GeoWill:
This is to implement #566.
Concerned this might be too much logic in the template, but unsure.

### DIFF
--- a/every_election/apps/organisations/templates/organisations/organisation_detail.html
+++ b/every_election/apps/organisations/templates/organisations/organisation_detail.html
@@ -58,6 +58,18 @@
             </table>
         {% endfor %}
         </ul>
+        <h2>Elections</h2>
+        <ul>
+            {% for election in object.election_set.all %}
+                {% if election.identifier_type == 'organisation' %}
+                    <li>
+                        <a href="{{ election.get_absolute_url }}">
+                        {{ election.get_id }}</a>
+                    </li>
+                {% endif %}
+            {% endfor %}
+        </ul>
+
         <h3>API</h3>
         <ul>
             <li><a href="{{ api_detail }}">
--- a/every_election/apps/organisations/templates/organisations/organisation_detail.html
+++ b/every_election/apps/organisations/templates/organisations/organisation_detail.html
@@ -61,11 +61,18 @@
         <h2>Elections</h2>
         <ul>
             {% for election in object.election_set.all %}
-                {% if election.identifier_type != 'ballot' %}
+                {% if election.election_type_id == 7 or election.election_type_id == 8 %}
                     <li>
                         <a href="{{ election.get_absolute_url }}">
                         {{ election.get_id }}</a>
                     </li>
+                {% else %}
+                    {% if election.identifier_type != 'ballot' %}
+                        <li>
+                            <a href="{{ election.get_absolute_url }}">
+                            {{ election.get_id }}</a>
+                        </li>
+                    {% endif %}
                 {% endif %}
             {% empty %}
                 <li>No elections found for {{ object.name }}</li>
--- a/every_election/apps/organisations/templates/organisations/organisation_detail.html
+++ b/every_election/apps/organisations/templates/organisations/organisation_detail.html
@@ -61,12 +61,14 @@
         <h2>Elections</h2>
         <ul>
             {% for election in object.election_set.all %}
-                {% if election.identifier_type == 'organisation' %}
+                {% if election.identifier_type != 'ballot' %}
                     <li>
                         <a href="{{ election.get_absolute_url }}">
                         {{ election.get_id }}</a>
                     </li>
                 {% endif %}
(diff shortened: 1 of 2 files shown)


GeoWill:
Thanks for the review. The PCCs throw up a tricky case where the elections for individual orgs like Northumbria Police look like ballots. In the case of the PCCs the three 'non-ballot' elections are `pcc.2016-05-05`, `pcc.2019-07-18`, `pcc.2020-05-07`.  Options seems to be:

1. Show non-ballot elections for all election types except PCCs (and maybe others)  where we show the ballots - I guess this behavior could be controlled by checking the `election_type_id`.
2. Show all elections whether they're a ballot or not - maybe sorting them into sub-sections by group type.
3. Something I've not thought of.

Not having used EE I don't have a good feeling for which are most useful to show, so recommendations welcome.

chris48s:
Good point. PCC and mayor both have this characteristic where the representative is elected directly to the organisation so there are no divisions.

There's a couple of other places where we special-case `election_type in ["mayor", "pcc"]` to deal with this. Lets just do the same here.

GeoWill:
Think this is getting there. One more potential issue is where there are elections subtypes - i.e. GLA, NAW and SP. Here the subtype elections (eg `gla.a.2016-05-05`) are listed alongside their parent. Not sure if this matters?

chris48s:
Also a good spot. Lets exclude the ones where `identifier_type == "subtype"` so we've just got the top-level ID (so it works similarly to UK parliament etc)

every_election/apps/organisations/templates/organisations/organisation_detail.html
```diff
@@ -58,6 +58,27 @@ <h3>{{ divisionset.short_title }}</h3>
             </table>
         {% endfor %}
         </ul>
+        <h2>Elections</h2>
+        <ul>
+            {% for election in object.election_set.all %}
+                {% if election.election_type_id == 7 or election.election_type_id == 8 %}
```
chris48s: These are basically just arbitrary numbers. PCC is only ID 7 by coincidence. It doesn't really mean anything and if I deleted one of these records and re-created it, it would get a different number. Can we use the labels so the code is

a) easier to read and
b) won't break if we change the arbitrary numbers

chris48s:
The other thing you've already kind of hinted at in the first post is that we're now putting quite a lot of logic in the template (and because of the limitations of templates, some of the things we're trying to do might be a bit cumbersome to do like that)

One thing you could also look at here as we add yet another condition is you could move the logic that decides which elections we're going to show on this page and which ones we aren't into an ORM query in

`OrganisationDetailView.get_context_data()`

so just to simplify it a bit, lets say we just take the original case where we just had

```django
{% for election in object.election_set.all %}
  {% if election.identifier_type == 'organisation' %}
    <li>
      <a href="{{ election.get_absolute_url }}">
      {{ election.get_id }}</a>
    </li>
  {% endif %}
{% endfor %}
```

another way we could have done this would be to change our model query in `OrganisationDetailView.get_context_data()` to something like

```py
from django.db.models import Prefetch
elections = Election.public_objects.all().filter(group_type="organisation")
obj = (
    Organisation.objects.all()
    .prefetch_related(Prefetch("election_set", elections))
    .get_by_date(**kwargs)
)
```

and then in the template

```django
{% for election in object.election_set.all %}
  <li>
    <a href="{{ election.get_absolute_url }}">
    {{ election.get_id }}</a>
  </li>
{% endfor %}
```

Now that we're building up quite a few conditions, taking that idea forward might help us clean this up a bit. Simultaneously, there is a slight tradeoff that if we're assembling a DB query, we can only use model properties that actually corespond to a DB column - we can't use SQL to filter by something that is an `@property` on the python object, so it might make sense to mix and match a bit. I'll leave it with you to have a play...

Good job we picked a nice simple one to start with, eh :D

GeoWill:
:D I will have a play.

symroe:
Drive by review, I know, sorry!

Consider that `order_by` isn't possible in the template, and we might have opinions about the order of elections in future. See also `select_related` etc. It feels like putting this in the view would be useful here, and then just iterating over the QS if it exists in the template

GeoWill:
This is meant to filter the election Queryset down to just include elections we want. I think the logic is clearer in the view than in the nested template. However this needs to be tested, and I'm not sure where such a test should go...

If this is the approach we go with I will put some detail in the commit message about why those elections are the ones being selected.

chris48s:
Sorted. This looks great. Happy to just squash and merge it as-is tbh.

Re: tests. At the moment, we haven't actually got tests for any of the organisation views. Its probably a bit disproportionately laborious to create all the fixtures to do this ad-hoc so there's probably a bit of work there for us to get some generic fixtures in shape to make that process a bit easier. Lets not block merging this PR on that, but it might be a good thing to get into down the line..

Issue #466, opened 2018-12-19
/organisations should group divisions by subtype


chris48s:
For example https://elections.democracyclub.org.uk/organisations/naw/naw/1999-05-06/ shows regions and constituencies in one table sorted by name, but it probably makes more sense to group the regions together and the constituencies together

chris48s:
closed in #649

PR #649, opened 2020-01-15
Order subdivisions by type on organisation page

GeoWill:
Implements #466 
At the moment the regions and constituencies are all in one table, but grouped together. This didn't seem unreasonable - but I wondered if the intention was to have one table per subdivision type?
At any rate its worth checking that the general approach is sensible before doing any more.


chris48s:
I reckon the way you've done it is fine :+1:

Issue #427, opened 2018-10-19
split up election_detail template


chris48s:
This template https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/elections/templates/elections/election_detail.html

has got quite a lot of nested logic in it and could usefully be split into some smaller chunks which are a bit easier to digest

chris48s:
done in #653

PR #653, opened 2020-01-20
Refactor elections template

GeoWill:
This is tackling #427 and #557. 
On #427 I'm not sure how far to break it down. I've made the indents line up a bit better in the new election_summary.html file. This helps with readability, and maybe more splitting adds confusion.

### DIFF
--- a/every_election/apps/elections/templates/elections/election_summary.html
+++ b/every_election/apps/elections/templates/elections/election_summary.html
@@ -47,6 +47,11 @@
       {# Only do this for non-group IDs #}
     {% endif %}
 
+    {% if object.seats_total and object.seats_contested %}
+      <dt>Total Seats</dt><dd>{{ object.seats_total }}</dd>
+      <dt>Seats Contested</dt><dd>{{ object.seats_contested }}</dd>
+    {% endif %}
+
     {% if object.cancelled %}
       {% if object.cancellation_notice %}
         {% include './official_document.html' with document=object.cancellation_notice type="Notice of Cancellation Document" only %}
--- a/every_election/apps/elections/templates/elections/election_summary.html
+++ b/every_election/apps/elections/templates/elections/election_summary.html
@@ -17,22 +17,14 @@
     <dt>ID</dt> <dd>{{ object.election_id }}</dd>
     <dt>Type</dt> <dd>{{ object.election_type }}</dd>
     <dt>Date</dt> <dd>{{ object.poll_open_date }}</dd>
+
     {% if object.organisation %}
       <dt>Organisation</dt><dd><a href="{{ object.organisation.get_absolute_url }}">
       {{ object.organisation }}</a></dd>
     {% endif %}
 
     {% if object.division %}
-      <dt>Division</dt>
-      <dd>
-        {% if object.division.format_geography_link %}
-          <a href="{{ object.division.format_geography_link }}.html">
-            {{ object.division.official_identifier }}
-          </a>
-        {% else %}
-          <strong>Missing</strong>
-        {% endif %}
-      </dd>
+      <dt>Division</dt><dd>{{ geography_html }}</dd>
     {% endif %}
 
     {% if object.identifier_type == "organisation" or object.identifier_type == "ballot" %}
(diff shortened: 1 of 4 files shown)


chris48s:
I reckon we could definitely split it down a bit further. In general there are 2 approaches we can employ here.

* One tactic that is available to us is be to split the template itself down into more includes.
* Another tactic we can employ is to just do more in python code before we get to the template.

So like, lets say this block is the issue:

```django
  <h1>
    {% if object.election_title %}
      {{ object.election_title }}
    {% else %}
      {{ object.election_id }}
    {% endif %}
    {% if object.cancelled %}
      <span class="label alert">cancelled</span>
    {% endif %}
  </h1>
```

One tactic is to just to move that block into another template and then replace it with `{% include "./election_title.html" %}`. Out of sight, out of mind, but then you've also got extra cognitive overhead of having lots of nested templates.

But then the other approach to the same problem is move some of that logic up to the view (python code is a more appropriate place to express it because we can group it into functions and stuff if we want):

```py
if context["object"].election_title:
    context["election_title"] = context["object"].election_title
else:
    context["election_title"] = context["object"].election_id
```

and then our template becomes:

```django
  <h1>
    {{ election_title }}
    {% if object.cancelled %}
      <span class="label alert">cancelled</span>
    {% endif %}
  </h1>
```

---

You could probaly pull out the `<form>...</form>` block to its own template as that's a fairly self-contained block - it makes sense to be its own template.

I reckon the section for rendering links to notice of election/cancellation docs can proabably be cleaned up quite a bit. Maybe this is a good candidate for moving the logic for working out which notice doc we want to render (we want to render either one or zero documents) into the view, setting some vars `context["document"]`, `context["document_title"]` and then simplifying that whole thing to 

```django
{% if document %}
    {% include './official_document.html' with document=document type=document_title only %}
{% endif %}
```

That probably also allows us to move the closing `</dl>` into a less odd place too.

---

I've not actually tried running any of the code above, but hopefully it conveys the ideas. Also I'm interested in your views on what makes it more/less readable. As someone who has just come to this code today and tried to work out what it does, you probably have a clearer idea of which bits of it are easy/hard ot understand than me :)

symroe:
> So like, lets say this block is the issue:

To take this example even further than Chris, I'd also consider moving the logic on to a model property, so in the template we can use `{{ object.nice_election_title }}` (or whatever the property is called). 

This allows us to use the same logic in other views, rather than adding to the view context every time we want to show the title like this.

(I've also not really read the code, but did read Chris's comments and wanted to nit pick ;) )

GeoWill:
Thanks for the feedback both.

every_election/apps/elections/templates/elections/election_summary.html
```diff
@@ -17,22 +17,14 @@ <h3>{{ object.election_subtype }}</h3>
...
-      <dd>
-        {% if object.division.format_geography_link %}
-          <a href="{{ object.division.format_geography_link }}.html">
-            {{ object.division.official_identifier }}
-          </a>
-        {% else %}
-          <strong>Missing</strong>
-        {% endif %}
-      </dd>
+      <dt>Division</dt><dd>{{ geography_link | safe }}</dd>
```
GeoWill: Is using safe in this way OK?

every_election/apps/elections/templates/elections/election_summary.html
```diff
@@ -48,23 +40,15 @@ <h3>{{ object.election_subtype }}</h3>
...
-      {% endif %}
+    {% if document %}
+      {% include './official_document.html' with document=document type=document_type only %}
     {% endif %}
+  </dl>
+
+  {% if not object.notice and not object.group.notice and user_can_upload_docs and not object.cancelled %}
+    <hr />
+    {% include "./election_form.html" %}
+  {% endif %}
```
GeoWill: Thought I should check that the previous intention to only show the form if the election wasn't cancelled was still the case.

GeoWill:
Working on this has been a bit broken up by other things. I imagine there is more that could be done, but it's also further along than it was, and hopefully close a stage that enough has been done.

every_election/apps/elections/views/general.py
```diff
@@ -77,12 +77,44 @@ def get_object(self, queryset=None):
...
+        if (
+            context["object"].division
+            and context["object"].division.format_geography_link
+        ):
+            division = context["object"].division
+            geography_link = "<a href={}.html>{}</a>".format(
+                division.format_geography_link(), division.official_identifier
+            )
+        else:
+            geography_link = "<strong>Missing</strong>"
```
chris48s: :bug:  This case isn't working as expected for me. If you look at http://127.0.0.1:8000/elections/local.cambridgeshire.duxford.by.2020-02-27/ , I'd expect that to hit the "missing" case, (as on https://elections.democracyclub.org.uk/elections/local.cambridgeshire.duxford.by.2020-02-27/ ), but its linking me to http://127.0.0.1:8000/elections/local.cambridgeshire.duxford.by.2020-02-27/None.html

every_election/apps/elections/templates/elections/election_summary.html
chris48s: Yeah you can do.
The other option is to use `django.utils.safestring.mark_safe` to mark it as safe on the python side.
Example of this: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/e5c5e7c1291fb7ac19c72ca06fcef5a45dbb3272/wcivf/apps/elections/models.py#L253-L261

There isn't really a single "right" way to do this, but it is something we've discussed before: https://github.com/DemocracyClub/WhoCanIVoteFor/pull/305#discussion_r230672116 Following that pattern, I'd probably call the function/var `geography_link_html` and use `mark_safe()` in python, but as I say there is more than one way to skin this cat and there isn't really a "correct" answer.

every_election/apps/elections/templates/elections/election_summary.html
chris48s: Good spot I think you've actually uncovered a TODO that never got done here, which is that the DB fields exist to attach a "cancelled election" notice, but there isn't a corresponding form in the front-end. Lets keep this PR to a straight refactoring job (and keep the existing functionality), but I'll raise it as an issue.

GeoWill:
Think this is good to go bar failing coveralls - but I assume that's for another time...

PR #661, opened 2020-02-03
Remove division to organisation fk

GeoWill:
MAIN CHANGES:
- Edits the division model to remove the organisation fk.
  N.B. The constraint for 'unique_together' is now on divisionset and official_identifier.
- Adds 'organisation' and 'organisation_id' properties to the division model.
- Ensures divsion sets are saved before divisions when importing from csv.
- Edits the admin views for Division problem and organisation_division.
- Removes the organisation subfactory from the division factory.
- Adds a migration for the db changes.

TESTING ISSUE:
At the moment I'm heading in the direction of removing the ability to pass 'organisation' as a kwarg to the `OrganisationDivisionFactory`. This makes sense to me because I don't think we need to pass an organisation when we create a division in the main codebase, so it would just be a testing convenience. It does mean that the tests now need intermediate divisionsets to be created to ensure the divisions are children of the organisations. 

QUESTION:
Initially I was trying to improve the factory but this didn't seem straightforward. The approach I describe above makes me nervous because it feels like I'm cheating by changing the tests! Is this approach going to come back bite us? 

TODO:
- ~Get test_backport_codes to pass. At the moment it's failing around the application of the `spatial_filter` in `Boundaryline.get_division_code` when the layer ends up with no features after the filter. However something isn't right as it's still failing when the filter is commented out. I hope~ this ~isn't~ is because I've still got the wrong GDAL installed.
- ~Remove `exclude('organisation')` from `OrganisationDivisionFactory` (i.e. delete 
498108e) and make sure none of the tests are passing it as a kwarg OR improve on the `OrganisationDivisionFactory` to be able to use the organisation kwarg to make sure the resulting division is a child of the organisation.~ Done


chris48s:
The purpose of the factories is to allow us to quickly produce realistic fake data for testing purposes.

If its currently possible to use the factories to produce a bunch of divisions that are related to an organisation without being a child of a divisionset which is related to the same organisation, that's a problem we should fix. In real life, we don't (or shouldn't!) have any divisions which are a child of a divisionset related to a different organisation, so if we're creating that problem under test, we aren't really testing with completely realistic test data. Even if it means setting up the test data gets a bit more complicated, the correct way to solve this is that every `OrganisationDivision` object created by a `OrganisationDivisionFactory` class should be related to its parent `Organisation` via a `OrganisationDivisionSet` created by a corresponding `OrganisationDivisionSetFactory`.

We should definitely remove the FK from `OrganisationDivision` to `Organisation` in the test data as a test object should behave like a real object. If we need to change the tests to support that, that's fine: its not "cheating" - we're fixing a problem :)

I don't really have a handle on the _complete_ implications of this across the entire test suite, but I think:

* If we just want to create a division but we don't really care about the divset/organisation, we can just continue to call `div = OrganisationDivisionFactory()` and it will create an arbitrary `OrganisationDivisionSet` and `Organisation` for us.
* If we want to create a division but we do care about the organisation it is related to, maybe before we were doing:

```py
org = OrganisationFactory()
div = OrganisationDivisionFactory(organisation=org)
```

but now that must become:

```py
org = OrganisationFactory()
divset = OrganisationDivisionSetFactory(organisation=org)
div = OrganisationDivisionFactory(divisionset=divset)
```

I'm aware that in some cases we're using a bit of a mix of objects created by factories and `Model.objects.create`, but the same principle holds.

Does that help?

every_election/apps/organisations/models/divisions.py
```diff
@@ -113,7 +110,15 @@ def __str__(self):
...
-        unique_together = ("organisation", "divisionset", "official_identifier")
+        unique_together = ("divisionset", "official_identifier")
+
+    @property
+    def organisation(self):
+        return self.divisionset.organisation
+
+    @property
+    def organisation_id(self):
+        return self.organisation.id
```
chris48s: I hadn't thought of doing this, but its quite a cunning way to make sure code that is still _reading_ `.organisation` or `.organisation_id` keeps working without having to trawl through every reference to it. We do need to be careful about any code that tries to _write_ to those properties though..

GeoWill:
Thanks for that. The orgfactory->divfactory becoming orgfactory->divsetfactory->divfactory is the pattern I'd started implementing. But I was worried that I should be able to handle the initial case. Not needing to is good.

every_election/apps/organisations/models/divisions.py
GeoWill: That's a good point. I guess I could:
- Add tests asserting that they raise an Exception when you try and write to them.
- Add an explanatory comment.
- Add a setter method which says 'oi this isn't writable'
- Do nothing and hope it's obvious :grinning:

every_election/apps/organisations/management/commands/import_divisionsets_from_csv.py
```diff
@@ -164,14 +164,15 @@ def create_divisions(self, csv_data):
...
             self.divisions.append(division)
 
     @transaction.atomic
-    def save_all(self):
+    def save_division_sets(self):
         for _, record in self.division_sets.items():
             record.save()
+
+    @transaction.atomic
+    def save_divisions(self):
```
chris48s: The issue with this is a failure in `save_divisions()` won't roll back the changes in `save_division_sets()` because they've already been persisted in a different transaction.

2 options:

1. Move `@transaction.atomic` out a layer to wrap `handle()`

2. We could also do:

    ```py
    @transaction.atomic
    def save_all(self):
        for record in self.divisions:
            record.divisionset.save()

            # hack: see https://code.djangoproject.com/ticket/29085
            record.divisionset = record.divisionset

            record.save()
    ```

    and keep all the saves at the end.

Either way, my original implementation was a bit ham-fisted anyway, so whichever will be an improvement.

every_election/apps/organisations/models/divisions.py
chris48s: No I don't think we need to do anything extra. The standard library does a perfectly good job of raising `AttributeError: can't set attribute` etc. I just mentioned it because while we can ignore instances of `x = div.organisation`, we need to make sure we're catching all the occurrences of `div.organisation = x`

every_election/apps/organisations/management/commands/import_divisionsets_from_csv.py
GeoWill: Gone for second option because I think it's going to be fixed in [the future](# https://github.com/django/django/commit/519016e5f25d7c0a040015724f9920581551cab0)  and it makes it obvious which bit needs to be atomic.

every_election/apps/organisations/management/commands/ni_import_district_electoral_areas.py
```diff
@@ -63,10 +63,11 @@ def create_divisions(self, csv_data):
         for line in csv_data:
             org = self.get_org_from_line(line)
             id_ = "gss:{}".format(line["District Electoral Area GSS code"])
+            div_set = OrganisationDivisionSet(organisation=org)
             div = OrganisationDivision(
                 official_identifier=id_,
                 temp_id="",
-                organisation=org,
+                divisionset=div_set,
```
chris48s: I can't leave a diff comment on the exact line, but a few lines down from here we set `div.divisionset = None` again. That'll want removing :)

GeoWill: Doh!

every_election/apps/organisations/management/commands/ni_import_district_electoral_areas.py
```diff
@@ -63,26 +63,25 @@ def create_divisions(self, csv_data):
         for line in csv_data:
             org = self.get_org_from_line(line)
             id_ = "gss:{}".format(line["District Electoral Area GSS code"])
+            div_set = OrganisationDivisionSet(organisation=org)
```
chris48s: also, this wants to be `self.division_sets[org.official_identifier]`

GeoWill: ...because it's not a factory... :roll_eyes:

PR #664, opened 2020-02-05
Fix divisionset date display (closes #663)

### LINKED ISSUE #663: DivSets created to reflect future changes have confusing dates.
When we create a divisionset that will come into effect in the future we describe it as `Active from 5 May 2022 to now.` 
Should be fine if we drop the 'to now'.

GeoWill:


every_election/apps/organisations/models/divisions.py
```diff
@@ -19,6 +20,15 @@ def __str__(self):
...
         )
 
+    @property
+    def end_date_to_display(self):
+        if self.end_date:
+            return "to {}".format(self.end_date.strftime("%d %b %Y"))
+        elif self.start_date > date.today():
+            return ""
+        else:
+            return "to now"
```
chris48s: Good call. From "[date in future] to now" makes no sense :+1: 

1. Can we also use this in the model's `__str__` method so we get a nicer description on the console as well as in the front-end.
2. Do you reckon it would be better to just say "onwards" in both cases as we never really mean the end date is "now", we just mean we don't know when the actual end date is yet.

GeoWill: OK, incorporated those, and changed the property to `active_period_text` which saves fiddling with date formats in `__str__`.

PR #674, opened 2020-02-17
Better titles for mayoral elections/subtype elections

chris48s:
Closes #264
Closes #11

PR #675, opened 2020-02-17
Mark `gla.a` elections as a ballot

symroe:


chris48s:
What's here is great, but were you also going to do a data migration to null out the group type for `gla.a.2020-05-07`?

symroe:
Yep, sorry, pushed this to get an idea of direction from you, but then got stuck in the test fails. 

I'll work on a data migration now.

chris48s:
cool. Probably makes sense to merge #674 first and rebase onto that so we don't end up with 2 migration 0057s in the elections app

PR #930, opened 2020-12-17
Alter field migration

GeoWill:
I'm not sure where this has come from...
Django Extensions changed their [TimeStampedModel in 3.0](https://github.com/django-extensions/django-extensions/blob/master/CHANGELOG.md#300) but that's not part of this afaict.


chris48s:
Probably should have been added in https://github.com/DemocracyClub/EveryElection/pull/882 . https://github.com/DemocracyClub/EveryElection/commit/d0b7d1c7b2027322824ef970a6f4db0e238dbd0c changes the enum but there's no migration.

I recently learned a cunning trick to help with this: Add `python manage.py makemigrations --check` to your CI build and it will exit with non-zero if a PR makes model changes which are missing migrations.

GeoWill:
Thanks, I've given the migration and commit a clearer name now.

Issue #954, opened 2021-01-08
Where should we keep seats_total for at large constituencies?


GeoWill:
For elections with a `Division`, `seats_total` lives there. For 'at large' constituencies (`gla.a`, `pcc`, `mayor`) this doesn't work.
I think it makes sense that `seats_contested` is only a property of the ballot, but it would be nice to have defaults for `seats_total` somewhere.  
At the moment the default is essentially `1`: 
https://github.com/DemocracyClub/EveryElection/blob/dde5a199e5a3a4de3deb381e168a082a7152d8d0/every_election/apps/elections/utils.py#L230-L231

So the simplest fix would be to special case gla in the same place. Alternatives would be adding it as a field to the `Organisation` model, or perhaps the `ElectionType` and `ElectionSubType` models. 

I think my current preference would just be for an extra if statement in the `get_seats_contested` method.

chris48s:
Random thoughts on this topic that you didn't ask for:

* For mayor/PCC, they inherently elect one representative: There can only be one mayor, there can only be one commissioner (PCC) so I think it is fine to just assume 1 for those election types and the simplest place to do that is in code. Those two are already fairly heavily special-case in the wizard/builder code.
* Its kind of odd to store a `seats_total` on the organisation as it is really a function of the `Organisation` and something else. e.g: Watford council has one mayor, but it also has 26 councillors so Watford Council / `seats_total`=1 is conceptually wierd. Also the GLA has 25 total members (14 geographic, 11 additional) and although they're elected differently functionally they're all London Assembly members so London Assembly / `seats_total`=11 would be properly odd.
* Another possible model if you want to store it in the DB instead of code would be to say if there's no division, the number of reps can be defined on the `ElectedRole` or `ElectionType`? That might be a slightly nicer way to express "there can only be one mayor of a place"?
* Sym can probably fill in better on this, but I think the whole "elects directly to an organisation" thing for Mayors/PCCs was basically to avoid the IDs being `mayor.watford.watford.2018-05-03` etc. I've previously thought that another possible model for this would be that mayor/PCCs have an "invisible" divisionset with one division in it (but that is probably **big**).
* `gla.a` is its own unique special snowflake and is special-cased everywhere. Again, I think the only reason it is like that is really to make the ID be `gla.a.2021-05-06` not `gla.a.a.2021-05-06`. Again another possible model would be that it exists as an 'invisible' region-like division (its basically the same as SP/Senedd regions except there is only one of them).

[...]

PR #955, opened 2021-01-08
Check Division when setting seats_contested

GeoWill:
Hopefully this doesn't create new ways of being wrong, while making us right more of the time...
It definitely doesn't completely fix the issue. Outstanding is what to do with at large constituencies (#954) and the all up tickbox for locals described in the `ToDo`.

### DIFF
--- a/every_election/apps/elections/tests/test_election_builder.py
+++ b/every_election/apps/elections/tests/test_election_builder.py
@@ -161,6 +161,69 @@ class TestElectionBuilder(BaseElectionCreatorMixIn, TestCase):
         self.assertEqual(1, election.seats_contested)
         self.assertEqual(3, election.seats_total)
 
+    def test_get_seats_contested(self):
+        sp_election_type = ElectionType.objects.get(election_type="sp")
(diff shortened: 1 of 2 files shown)


chris48s:
The issue with setting `election.seats_contested` to `division.seats total` by default is you essentially reverse the assumption in the comment i.e: this assumes all-up local elections are the most common case, when 1/3 up is really the more common situation (AFAIK). Obviously both `1` and `division.seats total` are wrong some of the time (we need information we don't have to make the decision), but I think `division.seats total` is wrong more of the time than 1 if the type is `local`.

symroe:
>  this assumes all-up local elections are the most common case

This is in the conditional: 

`if self.election_type.election_type != "local":`

So only applies for non-local elections? Or have I misread the logic here?

GeoWill:
Yep, I didn't want make any inference/guesses about `locals`. But might not have fully understood what's going on.

chris48s:
Right yes sorry - it is probably sensible to set `election.seats_contested = division.seats total` for everything other than local elections and by-elections :+1:

symroe (approved):

PR #977, opened 2021-01-29
Add seats contested for election groups

GeoWill:
seats_up_locals

This is a bit 'nobody asked for it' except, that @pmk01 sort of did, just not for it to be on the website. ~It makes page load for `/elections/local.2021-05-06/` really slow, which is reason enough not to merge - if we did want it the perhaps hiding it behind a button or `if user.is_staff` would be helpful.~ Doesn't seem to add much if anything to page load now. 

The main gains to this sort of thing seem to me:
- easy reporting for tweets and blogs 
- indicating possibility of 'out by one' errors when these sorts of numbers get compared to lists compiled by others. 

It made me think about what other numbers it would be nice to have exposed easily on EE, however that's probably a conversation for how to make the frontend of EE more wieldly in general.

### DIFF
--- a/every_election/apps/elections/models.py
+++ b/every_election/apps/elections/models.py
@@ -263,6 +263,37 @@ class Election(models.Model):
         near_future = date.today() + timedelta(days=settings.CURRENT_FUTURE_DAYS)
         return self.poll_open_date >= recent_past and self.poll_open_date <= near_future
 
+    def get_ballots(self):
+        """
+        If self has a group_type this returns all ballots that are descended from self.
+        If self doesn't have a group_type (i.e. is a 'ballot') it returns itself.
+        """
+        if self.group_type:
+            group, date = self.election_id.rsplit(".", 1)
+            return Election.public_objects.filter(
+                election_id__startswith=group + ".",
+                election_id__endswith=date,
+                group_type=None,
+            )
+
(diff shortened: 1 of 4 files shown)


symroe:
I wonder if we can use the format of the election IDs to our advantage here:

```py
pre, suff = "local.2021-05-06".rsplit(".", 1)
Election.public_objects.filter(
    election_id__startswith=pre, 
    election_id__endswith=suff, 
    group_type=None
).aggregate(
    Sum("seats_contested")
)
```

Produces `{'seats_contested__sum': 3765}` (on my old local DB - the number isn't the important part!)

Or

```py
pre, suff = "local.stroud.2021-05-06".rsplit(".", 1)
Election.public_objects.filter(
    election_id__startswith=pre, 
    election_id__endswith=suff, 
    group_type=None
).aggregate(
    Sum("seats_contested")
)
```

`{'seats_contested__sum': 51}`

GeoWill:
That's much better! I thought there must be something clever with the ids but couldn't get past accounting for the subtypes.

symroe:
I think `pre` needs to include the full stop at the end, otherwise `local.cambridge` and `local.cambridgeshire` will clash.

symroe (approved):

PR #990, opened 2021-02-11
Add election timetable to Election pages

symroe:
A quick implementation of the new `.timetable` method of `uk-election-timetables`.

Screenshot from 2021-02-11 11-11-07

### DIFF
--- a/every_election/apps/elections/models.py
+++ b/every_election/apps/elections/models.py
@@ -21,6 +21,8 @@ from django.urls import reverse
 from django_extensions.db.models import TimeStampedModel
 from django_markdown.models import MarkdownField
 from storages.backends.s3boto3 import S3Boto3Storage
+from uk_election_timetables.calendars import Country
+from uk_election_timetables.election_ids import from_election_id
 from uk_geo_utils.models import Onspd
 
 from .managers import PublicElectionsManager, PrivateElectionsManager
@@ -279,6 +281,27 @@ class Election(models.Model):
             )
         else:
             return None
+
+    @property
+    def get_timetable(self):
+        country_map = {
+            "WLS": Country.WALES,
+            "ENG": Country.ENGLAND,
+            "NIR": Country.NORTHERN_IRELAND,
+            "SCT": Country.SCOTLAND,
+        }
+        area = self.division or self.organisation
+        if not area:
+            return None
+
+        territory_code = area.territory_code or self.organisation.territory_code
+        if not territory_code:
+            return None
+
+        timetable = from_election_id(
+            self.election_id, country=country_map[territory_code]
+        ).timetable
+        return timetable
 
     def get_id(self):
         if self.election_id:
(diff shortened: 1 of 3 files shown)


every_election/apps/elections/models.py
```diff
@@ -280,6 +282,25 @@ def get_example_postcode(self):
...
 
+    @property
+    def get_timetable(self):
+        country_map = {
+            "WLS": Country.WALES,
+            "ENG": Country.ENGLAND,
+            "NIR": Country.NORTHERN_IRELAND,
+            "SCT": Country.SCOTLAND,
+        }
+        area = self.division or self.organisation
```
GeoWill, line 293: TIL you can assign like this in python. Must've seen it before, but never clicked.

symroe: Actually I've never used this much, but I was inspired by @michaeljcollinsuk's PR here: https://github.com/DemocracyClub/WhoCanIVoteFor/commit/42968771c1b0c2b3a379287ccb3fcdae43ef40d1#diff-8e7e5cffbf2702ce7fa6dbe7ccaf8383a8ee8094d04bbf5879327c13f2fa4720R95

every_election/apps/elections/models.py
```diff
@@ -280,6 +282,25 @@ def get_example_postcode(self):
...
+
+        if area:
+            territory_code = area.territory_code or self.organisation.territory_code
+            if territory_code:
+                timetable = from_election_id(
+                    self.election_id, country=country_map[territory_code]
+                ).timetable
+                return sorted(timetable, key=lambda r: r["date"])
```
GeoWill, line 301: Am I right in thinking https://github.com/DemocracyClub/uk-election-timetables/pull/11 will make the `sorted` unnecessary?

symroe: Yes - unnecessary, but also it won't matter to be sorted twice. Added here and then decided to upstream it

every_election/apps/elections/models.py
```diff
@@ -280,6 +282,25 @@ def get_example_postcode(self):
...
+    def get_timetable(self):
+        country_map = {
+            "WLS": Country.WALES,
+            "ENG": Country.ENGLAND,
+            "NIR": Country.NORTHERN_IRELAND,
+            "SCT": Country.SCOTLAND,
+        }
+        area = self.division or self.organisation
+
+        if area:
```
michaeljcollinsuk, line 295: I tend to prefer to use an inverted conditional/return early pattern to help avoid nesting if statements like this

```
if not area:
    return None

territory_code = area.territory_code or self.organisation.territory_code
if not territory_code:
    return None

timetable = from_election_id(
    self.election_id, country=country_map[territory_code]
).timetable
return sorted(timetable, key=lambda r: r["date"])
```

symroe: Good call, I've made this change

symroe:
Updated to `2.1.1` and removed the sorting - 2.1.1 adds sorting by default

michaeljcollinsuk (approved):

Issue #815, opened 2020-09-23
Store NUTS1 area for divisions and organisations


symroe:
A feature request from YNR that's easy to imagine being useful to other users is to be able to filter ballots by _some regional grouping_. 

It makes sense to use the NUTS1 groups, e.g the old `europarl` divisions for this. There are 9 English areas, and 3 more for Wales, Scotland and Northern Ireland.

Seems like a good approach would be to add a `nuts_area` column to `OrganisationDivision` and `Organisation`. This can be populated using the existing boundaries we have, e.g:

```python
regions = OrganisationDivision.objects.filter(division_type="EUR")
for region in regions:
    OrganisationDivision.objects.annotate(cent=Centroid("geography__geography")).filter(cent__within=region.geography.geography).update(nuts_area=region.slug)
```

(The actual code wouldn't be the slug, but we can use a lookup between slug and code).

This would then be exposed in the API for consumers to use.

GeoWill:
Another way that would let us add a variety of regional groupings would be to use a jsonb column on `OrganisationDivision`. It could be called `tags` or `groups` or similar.
It would look like:

```json
{
    "nuts1": {"value": "foo", "label": "bar"},
    "historic_county": {"value": "foo", "label": "bar"}
}
```

This would get populated with a management command, so could be flexible about whether the geometries we're grouping by live in a file or the db. Given this'll be consumed by YNR we would need to pass back an issue when an expected tag was missing. 

--------------------------

Also a couple of possible issues with using centroid: 
- In case of islands the centroid might not fall within the grouping geom.
- If we're grouping by geometries that won't neatly contain divisions then we might want to base tagging on a threshold overlap - eg at least 50% contained by this geometry.

pmk01:
As part of this it would also be good if we could filter by District, County, Unitary & Metropolitan District councils...

GeoWill:
Tags for NUTS1 areas are in the DB and being pulled through to the API:
```json
[...,
{
    "election_id": "local.brentwood.hutton-central.2021-05-06",
    "tmp_election_id": null,
    "election_title": "Brentwood local election Hutton Central",
    "poll_open_date": "2021-05-06",
    "election_type": {...},
    "election_subtype": null,
    "organisation": {...},
    "group": "local.brentwood.2021-05-06",
    "group_type": null,
    "identifier_type": "ballot",
    "children": [],
    "elected_role": "Local Councillor",
    "seats_contested": 1,
    "division": {...},
[...]

PR #992, opened 2021-02-14
Area tags

GeoWill:
This is a stab at a general fix for #815.

The management command still leaves something to be [desired](https://github.com/DemocracyClub/EveryElection/blob/57b73c32abea089a84e8047c097d6c48dc92c063/every_election/apps/organisations/management/commands/add_tags.py#L71-L73), but it does work, and isn't too slow thanks to using `set_jsonb`.

I departed from the spec in the issue by putting the tags on the `Election` model - but at the moment, only where 'group_type' is none. This is because we expect a common use case is to derive these from geographic data, and if we only attach the tags to `OrganisationDivisions` we will miss elections where the 'division' uses an `OrganisationGeography`. By attaching the tags to an election we only have to store them once.


every_election/apps/organisations/management/commands/add_tags.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def get_layer(data, layer_index=0):
+    data_source = DataSource(data.name)
+    if len(data_source) < layer_index + 1:
+        raise ValueError(f"Expected layer at index: {layer_index}, None found")
+    return data_source[layer_index]
+
+
+def get_ballots_by_area(area: GEOSGeometry):
```
symroe, line 47: This feels like a really useful method to have on the [QuerySet](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/elections/managers.py#L9) - do you mind:

1. moving it there and updating the line below to `ballots = Election.private_objects.get_ballots_by_area(feature.geom.geos)`
2. Accordingly changing `queryset` to use `self.filter` rather than hard coding the `private_objects`

I think this would also help with your point about adding more filters for this command, e.g:

```
qs = Election.private_objects.get_ballots_by_area(feature.geom.geos)
if options['current_only']:
    qs = qs.filter(current=True)
```

etc

every_election/apps/organisations/management/commands/add_tags.py
```diff
@@ -0,0 +1,113 @@
...
+from django.core.management import BaseCommand
+from django.db import models
+
+from core.mixins import ReadFromFileMixin
+from elections.models import Election
+
+from django.db.models.expressions import Func
+
+
+class JsonbSet(Func):
```
symroe, line 15: As per the other comment - I think this would be useful elsewhere, maybe in `core/models.py`?

every_election/apps/organisations/management/commands/add_tags.py
```diff
@@ -0,0 +1,113 @@
...
+        .filter(
+            models.Q(division_centroid__within=area)
+            | models.Q(organisation_centroid__within=area, division_centroid=None)
+        )
+    )
+    return queryset
+
+
+class Command(ReadFromFileMixin, BaseCommand):
+    """
```
symroe, line 67: If you assign this string to a `help` var on the class then Django will use it when printing `python manage.py add_tags --help` [e.g](https://github.com/DemocracyClub/yournextrepresentative/blob/master/ynr/apps/elections/uk/management/commands/uk_create_elections_from_every_election.py#L12)

every_election/apps/organisations/management/commands/add_tags.py
```diff
@@ -0,0 +1,113 @@
...
+            action="store",
+            dest="layer_index",
+            help="Index of layer in dataset",
+        )
+        super().add_arguments(parser)
+
+    def handle(self, *args, **options):
+        field_map = json.loads(options["field_map"])
+        tag_name = options["tag_name"]
+        print("Loading data...")
```
symroe, line 98: For management commands it's always best to use `self.stdout.write` - this is so that [we can capture the output in tests](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/organisations/tests/test_import_lgbce.py#L86) rather than the output getting printed to the test runner's `stdout`

every_election/apps/organisations/management/commands/add_tags.py
GeoWill: I did this originally (so I essentially agree :)), but took it off because I thought the 'centroid in area' implementation was a bit hacky - when I tried it with an `overlaps` lookup it was way slower and returned many fewer objects - I didn't dig into this, but made me think we should stick to the centroid method. 
I'm up for sticking it on the manager, but would like to rename it so it's clearer what it's doing: does `Election.private_objects.ballots_with_centroid_in_area(area)`, sound OK?

symroe: > does Election.private_objects.ballots_with_centroid_in_area(area), sound OK?

Yes! I was actually thinking about suggesting a rename, but I couldn't think of anything :)

every_election/apps/organisations/management/commands/add_tags.py
GeoWill: I'm not sure about this because I'm aware of at least one way my implementation doesn't work as one might expect:
 
If you're calling `jsonb_set` from psql you can do the following:

```sql
update elections_election
set tags = jsonb_set(tags, '{"BAZ"}', '{"bar":"foo"}', true)
```
and after running the above:

```sql
update elections_election
set tags = jsonb_set(tags, '{"BAZ","bar"}', "different_foo", true)
```

The equivalent in python:

```python
# qs is a queryset containing a single election object
qs.update(tags=JsonbSet("tags",keyname="BAZ",value={"bar":"foo"},create_missing=True))
# Returns 1
qs.update(tags=JsonbSet("tags",keyname='"BAZ","bar"',value="foo",create_missing=True))
# Error
```
Where the error is:
```
InvalidTextRepresentation: malformed array literal: "{""BAZ","bar""}"
LINE 1: ...T "tags" = jsonb_set("elections_election"."tags", '{""BAZ","...
                                                             ^
DETAIL:  Unexpected array element.
```
This is because of how I'm managing the string around the `keyname` arg, which is fine for the use case of 'update a top level key with a value' but not so much for 'update a nested key with a value'. I'm sure this can be sorted out, I just wasn't sure it was worth the time to do so if we didn't have a need. 
Is this a reason not to have it in the code base more generally? or is it enough to document its (bad) behaviour?

symroe: I see - how about we punt to an issue then - in general I think it would be nice to maintain these helper functions we have all over our projects in a more reusable way, but let's not block this PR on that :)

GeoWill: Found this https://github.com/django/django/pull/13152/files which put me on the right track. 
It does mean you have to be a bit clever about how you call it, but that was always the case.

chris48s:
Couple of drive-by comments:

> I thought the 'centroid in area' implementation was a bit hacky

Given centroid isn't even gauranteed to be inside the polygon, `ST_PointOnSurface()`?

> when I tried it with an overlaps lookup it was way slower and returned many fewer objects - I didn't dig into this

Probably needs to be covers or overlaps, rather than just covers.

That said, you probably do need to take a shortcut for performance. I suspect a "coverlaps" query on a (multi-)polygon with lots of points (e.g: a region) is going to be too slow for interactive usage whichever way you slice it.

FWIW, This is how I've handled this in management commands
https://github.com/DemocracyClub/EveryElection/blob/33eff41ee51d6bdc2d8d2ba57a0afbed48ca5ef4/every_election/apps/organisations/boundaries/helpers.py#L4-L10
which is a python implementation as we're doing that before we have imported the thing. Suspect an equivalent PostGIS implementation would not perform any better than what you already tried.

GeoWill:
Good points about `ST_PointOnSurface()` (for some reason I thought it wasn't in the ORM, but it's right there in the GEOS and db APIs...) and 'coverlaps' - I glossed over the 'doesn't contain' clause for overlaps. 

I shied away from your approach of calculating the intersection due to the time it takes - not a massive issue as imports should be occasional, but I thought why not see how much we use it first. Also our first usage is for NUTS1 areas, where most `DivisionGeographies` aren't going to be straddling two areas. If we were doing something more fine grained - eg 'historic counties' (or modern counties for that matter) - then intersection approach would definitely bring more value.

symroe (approved):

PR #1053, opened 2021-04-12
Performance tweaks

symroe:
## Before 

Screenshot from 2021-04-12 14-27-58

## After

Screenshot from 2021-04-12 14-28-13


GeoWill (approved):
Nice - good win, look forward to seeing a satisfying step down in the graphs :D.
For anyone following along, the reason for the `postcode` check is there's some number of elections where this is actually slower. However the case we care about (devs API) generally (always?) queries with a postcode.

PR #1144, opened 2021-07-06
Add admin action to allow bulk soft-delete

michaeljcollinsuk:
Allows elections to be "soft deleted" in bulk - as there are a number of old elections that @pmk01 has requested be removed.
- Admin action to soft delete elections in bulk
- Some additional admin configuration to add latest status to the election admin
- Disables the built in hard "delete" action in the admin

If these changes seem reasonable I will update the Wiki page https://github.com/DemocracyClub/EveryElection/wiki/Cancelled-Elections-and-Soft-Deletes

### DIFF
--- a/every_election/apps/elections/admin.py
+++ b/every_election/apps/elections/admin.py
@@ -4,7 +4,14 @@ from django import forms
 from django.contrib import admin
 from django.forms.widgets import Textarea
 from django_markdown.admin import MarkdownModelAdmin
-from .models import ElectedRole, Election, Explanation, MetaData, ModerationHistory
+from .models import (
+    ElectedRole,
+    Election,
+    Explanation,
+    MetaData,
+    ModerationHistory,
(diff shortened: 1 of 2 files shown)


GeoWill (approved):
This looks good to me - and seems to work as expected.
Couple of thoughts, but I think they're more for documentation/for sticking in issues:
- Soft-deleting an election object which has children won't soft delete the children. I can't really think of a circumstance where you would not want to also soft delete the children. Leaving 'orphans' seems like the biggest potential footgun with this.  
- The process for 'un-soft-deleting' is still 'add a moderation event'. I think this is worth noting because it might be assumed that 'setting to current' would have some effect - i.e. there's no 'ctrl-z' if someone does a bulk soft-delete and then realises it was wrong.
- I don't think the current flag does anything once an election is soft deleted, but the actions are still there. Not a big issue, but maybe worth a note.

Issue #182, opened 2018-02-10
West Midlands CA and West Midlands Police have same 'official identifier'


chris48s:
http://elections.democracyclub.org.uk/organisations/west-midlands/

throws

```
MultipleObjectsReturned at /organisations/west-midlands/
get() returned more than one Organisation -- it returned 2!
```

"West Midlands Combined Authority" and "West Midlands Police" both have the 'official identifier' `west-midlands`. There's several things to think about/deal with here:

* These organisations should have different identifiers (I assume using a slug here was a proxy for real identifiers, but both of these do have ONS codes. WM Police is `E23000014` and WM CA is `E47000007`. Obviously those bring the usual assortment of boundary change fun with them :) )
* 'official identifier' should probably have a unique constraint on it
* This brings up some interesting issues around different areas with the same slug. In this case they're differentiated, because the elections will be `mayor.west-midlands` and `pcc.west-midlands` but I'm already starting to imagine a plethora of horrible edge cases we haven't yet considered..

chris48s:
This is fixed by #315 because `official_identifier` is now one component of a composite key

PR #1157, opened 2021-07-20
Filter elections endpoint by org identifier

### LINKED ISSUE #1151: API enhancement: filter elections by organisation
### Context
I'm building a data visualisation tool to show the latest local election results for all wards within a specified area.

### Issue
Currently, the most efficient way of finding the latest election for each local authority / ward is by calling the `/elections` endpoint with `group_type=organisation` (organisation level elections only) and `poll_open_date__gte=2017-01-01` (a bit more than 4 years to capture a full council term) to retrieve all organisation level elections. However, this returns ~1600 results, takes around 18s in total and loads a huge amount of unnecessary data.

### Proposed solution
It would be really useful to be able to filter the `/elections` response by `organisation.official_identifier` so that I can quickly and easily find the latest elections for each local authority. Since the election object contains the `organisation` relationship already, hopefully it's not a huge task!

### Additional information

It seems like the query I'm looking for already exists in some form, as it's visible at the bottom of each Organisation page under "Elections". See [Bristol](https://elections.democracyclub.org.uk/organisations/local-authority/BST/1996-04-01/) for example!
https://github.com/DemocracyClub/EveryElection/blob/19da682c2a1827d1c7017cf9a052237676dd0669/every_election/apps/organisations/views/public.py#L38-L54
https://github.com/DemocracyClub/EveryElection/blob/19da682c2a1827d1c7017cf9a052237676dd0669/every_election/apps/organisations/templates/organisations/organisation_detail.html#L67-L73

michaeljcollinsuk:
- Reorganise ElectionViewset filterset_fields to ElectionFilter class
- Add organisation_identifier filter
- Closes #1151

### DIFF
--- a/every_election/apps/api/filters.py
+++ b/every_election/apps/api/filters.py
@@ -9,6 +9,14 @@ class ElectionFilter(django_filters.FilterSet):
         field_name="organisation__official_identifier",
         lookup_expr="exact",
     )
+    organisation_type = django_filters.CharFilter(
+        field_name="organisation__organisation_type",
+        lookup_expr="exact",
+    )
+    organisation_start_date = django_filters.DateFilter(
+        field_name="organisation__start_date",
+        lookup_expr="exact",
+    )
 
     class Meta:
         model = Election
--- a/every_election/apps/api/filters.py
+++ b/every_election/apps/api/filters.py
@@ -1,0 +1,18 @@
+import django_filters
+
+from elections.models import Election
+
+
+class ElectionFilter(django_filters.FilterSet):
+
+    organisation_identifier = django_filters.CharFilter(
+        field_name="organisation__official_identifier",
+        lookup_expr="exact",
+    )
+
+    class Meta:
+        model = Election
+        fields = {
+            "group_type": ["exact"],
+            "poll_open_date": ["exact", "gte", "lte"],
+        }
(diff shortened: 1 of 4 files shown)


symroe (approved):
:+1:

chris48s:
Organisations in Every Election have a combined key `(organisation_type, official_identifier, start_date)`.
See https://github.com/DemocracyClub/EveryElection/issues/182

This is why the API URL for Aberdeen (for example) has to be
https://elections.democracyclub.org.uk/api/organisations/local-authority/ABE/1996-04-01/
not
https://elections.democracyclub.org.uk/api/organisations/ABE

This filter is going to give you issues in situations where `official_identifier` in non-unique (which is rare but there is a non-zero number of cases and more could happen) because you'll get >1 organisations worth of elections merged together (which I don't think is the intent of the feature).

My suggestion for how to do avoid this problem in a nice way would be:

Instead of going from

https://elections.democracyclub.org.uk/api/elections?organisation_identifier=ABE
to
https://elections.democracyclub.org.uk/api/elections?organisation_type=local-authority&official_identifier=ABE&start_date=1996-04-01

to solve the issue, you make the endpoint

https://elections.democracyclub.org.uk/api/organisations/local-authority/ABE/1996-04-01/elections

michaeljcollinsuk:
Thanks for the additional context @chris48s 

@symroe I have updated to allow add additional filters to the Elections endpoint, and implemented the `/elections` endpoint on the OrganisationViewSet

symroe (approved):

PR #1166, opened 2021-07-28
Update election cache after saving new election

### LINKED ISSUE #1162: Can't create Medway by-election
Creating `local.medway.strood-north.by.2021-08-26` throws an error - `local.medway.princes-park.by.2021-08-26` was created fine

image
image

symroe:
This isn't an ideal solution, but the problem is a little more complex
to refactor than I'd anticipated.

Before the May 2021 local elections we were getting timeouts when adding
new elections. This was because the ID Builder performed an huge amount
of duplicate lookups.

As a quick fix, I added a cache, however the way I implemented it was to
set up a global, in memory cache. This will be cleared as gunicorn
workers cycle out, but it won't be cleared for each IdBuilder instance.

The result of this is that we can have a stale cache hanging around for
too long, and in turn this causes a false negative look up, meaning we
end up trying to create a duplicate election, and in the end causing a
500 due to a postgres duplicate key.

The real solution is to invalidate the cache after all the IDs have been
created. This solution invalidates the cache after each creator instance
is closed down. It's not clear when exactly this it, but it's sure to be
once per organisation.

This will cause more queries than needed, but will ensure that the bug
isn't hit.

Fixes #1162

### DIFF
--- a/every_election/apps/elections/tests/test_election_builder.py
+++ b/every_election/apps/elections/tests/test_election_builder.py
@@ -258,3 +258,35 @@ class TestElectionBuilder(BaseElectionCreatorMixIn, TestCase):
         self.assertEqual(default_status, ballot.moderation_status.short_label)
         self.assertEqual(default_status, org_group.moderation_status.short_label)
         self.assertEqual(default_status, election_group.moderation_status.short_label)
+
+    def test_can_create_duplicate_groups(self):
+        """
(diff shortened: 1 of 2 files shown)


GeoWill (approved):
Seems like a sensible fix to me - I guess it's one to revisit if using the builder is too slow in the future.

PR #1190, opened 2021-08-24
Add an overwrite flag to add_tags command

GeoWill:
This makes the command check for a top level key in the tags field that
matches the 'tag_name' passed to the command. If a key exists then it
will only be overwritten if the '--overwrite' flag has been passed.

### DIFF
--- a/every_election/apps/elections/management/commands/add_tags.py
+++ b/every_election/apps/elections/management/commands/add_tags.py
@@ -49,6 +49,11 @@ class Command(ReadFromFileMixin, BaseCommand):
             dest="layer_index",
             help="Index of layer in dataset",
         )
+        parser.add_argument(
+            "--overwrite",
+            action="store_true",
+            help="Overwrite existing tags where tag_name exists as a top level key",
+        )
         super().add_arguments(parser)
 
     def handle(self, *args, **options):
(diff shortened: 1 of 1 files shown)


every_election/apps/elections/management/commands/add_tags.py
```diff
@@ -64,9 +69,15 @@ def handle(self, *args, **options):
...
-            )
+
+            if options["overwrite"]:
+                ballots = Election.private_objects.ballots_with_point_in_area(
+                    feature.geom.geos
+                )
+            else:
+                ballots = Election.private_objects.ballots_with_point_in_area(
+                    feature.geom.geos
+                ).exclude(tags__has_key=tag_name)
```
symroe, line 80: This could be

```
ballots = Election.private_objects.ballots_with_point_in_area(
    feature.geom.geos
)
if not options["overwrite"]:
    ballots = ballots.exclude(tags__has_key=tag_name)

I think?

Comment (GeoWill) [reply every_election/apps/elections/management/commands/add_tags.py:80]: I flipped between the two :slightly_smiling_face:. Settled on this because I thought the ORM would turn 
```python
 ballots = Election.private_objects.ballots_with_point_in_area(
                    feature.geom.geos
                ).exclude(tags__has_key=tag_name)
```
into a single sql statement - presumably resulting in a more efficient query - which seemed worth the code duplication if it was going to run every day. 
Happy to flop back though!

symroe: Not sure if this is what you mean, but did you know that [querysets are lazy](https://docs.djangoproject.com/en/2.2/topics/db/queries/#querysets-are-lazy)?

so 
```python
foo = Ballots.objects.all()
foo = foo.filter(a=1)
```
Won't actually _do_ anything in terms of the database, it's just building the query.

`list(foo)`, or some other function that calls `foo` _will_ run SQL.

GeoWill: Of course - I forgot :slightly_smiling_face: 
That is what I was driving at - fixup in.

symroe (approved):

PR #1206, opened 2021-09-09
Add management command to add a refendeum election

michaeljcollinsuk:
- Added to allow us to add Croydon referendum initially, with
the possibility of being able to use again in the future as required

### DIFF
--- a/every_election/apps/elections/management/commands/add_referendum.py
+++ b/every_election/apps/elections/management/commands/add_referendum.py
@@ -1,0 +1,103 @@
+from django.core.management import BaseCommand
+
+from elections.models import (
+    Election,
+    ElectionType,
+    ModerationHistory,
+    ModerationStatuses,
+)
+from organisations.models.organisations import Organisation
+
+
+class Command(BaseCommand):
+    help = """
+    Adds an election with an election type of referendum
+    Example usage:
+    python manage.py add_referendum --date 2021-10-07 \
+        --council croydon \
+        --election-title "Governance referendum" \
+        --official-identifier CRY \
+        --org-type local-authority
+    """
+
+    def add_arguments(self, parser):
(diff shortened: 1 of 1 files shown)


every_election/apps/elections/management/commands/add_referendum.py
```diff
@@ -0,0 +1,89 @@
...
+            required=True,
+        )
+        parser.add_argument(
+            "--org-type",
+            action="store",
+            dest="org-type",
+            default="local-authority",
+            help="The council area the referendum is taking place",
+            required=True,
+        )
```
michaeljcollinsuk, line 59: If I have gone overboard with making these arguments I can hardcode them

symroe:
This looks decent. At the moment it only deals with referendums that cover the _whole_ organisation, not a division. It's worth at least adding that in the help text, but if possible allowing an optional `--division_id` that would allow adding a division.

When we do point in polygon lookups we try to use the division first, so this should be all that's needed.

every_election/apps/elections/management/commands/add_referendum.py
```diff
@@ -0,0 +1,100 @@
...
+            election=ref_election,
+        )
+
+        org = Organisation.objects.get_by_date(
+            date=election_date,
+            official_identifier=options["official_identifier"],
+            organisation_type=options["org_type"],
+        )
+        ref_election.organisation = org
+        ref_election.organisation_geography = org.geographies.latest()
```
symroe, line 99: I think you need to set the division geography on the table too?

michaeljcollinsuk: Sorry, unfamiliarity with the data model! Added in latest commit - if that looks right I can rebase before merging

symroe (approved):

PR #1220, opened 2021-09-24
Adds filter to exclude elections by election type

michaeljcollinsuk:
- Allows the elections endpoint to be filtered by individual or
multiple election types
- Adds tests
- Adding so that the election importer management command in YNR can
be updated to exclude referendums, but adding as a filter allows it to
be used in other circumstances where necessary


every_election/apps/api/filters.py
```diff
@@ -17,6 +17,12 @@ class ElectionFilter(django_filters.FilterSet):
         field_name="organisation__start_date",
         lookup_expr="exact",
     )
+    excluded_election_types = django_filters.AllValuesMultipleFilter(
+        label="Exclude election types",
+        field_name="election_type__election_type",
```
symroe, line 22: To save doing a join here (as per the NumQueries below), can we just filter on `election_id__startswith=f"{type}."`?

Or, now I'm thinking about it: regular expressions should be really fast for the election IDs, so we could just expose a `include_regex` and `exclude_regex` filter, where in this case we could pass `exclude_regex=ref.*`.

This would be quite powerful for 3rd parties too, as you could quickly pass `include_regex=*.by.*.2019` to get all by-elections in 2019, etc.

michaeljcollinsuk: Have had a go at filtering with a regex in a shell, but I think due to my unfamiliarity with them generally, not getting very far with it! So maybe on to discuss, or if you have any examples of how to go about it that would be useful.

I think the issue with using `startswith` is that it would limit it excluding/including a single election type. Which in practice for how we want to exclude referendums would be fine, I had just thought it would be better to include the option of excluding multiple types.

symroe: Sure, here are some examples:

```
# All referendums
Election.public_objects.filter(election_id__regex="^ref.*")

# All elections in September of any year
Election.public_objects.filter(election_id__regex="-09\-\d{2}")

# All elections in a division with "ham" in the name
Election.public_objects.filter(election_id__regex="^local.*.*ham*")

# All elections with a 3 letter part
Election.public_objects.filter(election_id__regex="\.[a-z]{3}\.")

# All by elections in 2019
Election.public_objects.filter(election_id__regex="\.by\.2019")

# All by elections in Norfolk
Election.public_objects.filter(election_id__regex="norfolk.*\.by\.")
```

michaeljcollinsuk: Updated the PR to add two filters to include/exclude by regex on the election id.

I have set a max limit of 20 chars - but happy to tweak

symroe (approved):

PR #1329, opened 2022-01-17
Update Election modified field when using QS update

michaeljcollinsuk:


michaeljcollinsuk:
@GeoWill fixed those failing tests - amended the update method to allow the `modified` date to be passed in through the kwargs, which is what is happening when an Organisation is saved

GeoWill (approved):
Looks good :+1:

PR #1336, opened 2022-01-26
SPIKE: filter by local authority area

symroe:
Endpoint for filtering ballots by local authority geographies.

I'm not sure about the magic string. Some initial testing and getting confused by the wikipedia page suggests it's current, but it's very much not in my head properly.

I tested the Highlands. It's really slow. So I moved to using `bboverlaps` as a _really_ rough proxy for the same thing, without all the fiddly bits. This only kicks in above a given area,but I don't know what the unit is!

It's going to give more false positives, but I think it's literally just the Highlands that have the problem. I also think that they'll mostly have all-up-Scotland elections, so it's not really going to matter.

### DIFF
--- a/every_election/apps/api/filters.py
+++ b/every_election/apps/api/filters.py
@@ -1,9 +1,63 @@
 import django_filters
+from django.core.exceptions import ValidationError
+from django.db.models import Q
 
 from elections.models import Election
+from organisations.models import OrganisationGeography
 
 
 class ElectionFilter(django_filters.FilterSet):
+    def election_intersects_local_authority_filter(self, queryset, name, value):
+
+        og_qs = OrganisationGeography.objects.filter(
+            organisation__official_identifier=value,
+            organisation__organisation_type="local-authority",
+        ).select_related("organisation")
+
+        if not og_qs.exists():
+            raise ValidationError(
+                """Only local authorities supported""",
+                code="invalid",
+            )
+
+        if "poll_open_date" in self.data and self.data["poll_open_date"]:
+            og_qs = og_qs.filter(
+                Q(start_date__lte=self.data["poll_open_date"]) | Q(start_date=None)
+            )
+
+        if (
+            "organisation_start_date" in self.data
+            and self.data["organisation_start_date"]
+        ):
+            og_qs = og_qs.filter(
+                Q(start_date__lte=self.data["organisation_start_date"])
+                | Q(start_date=None)
+            )
+
+        try:
+            og = og_qs.get()
+        except OrganisationGeography.MultipleObjectsReturned:
+            raise ValidationError(
+                """Organisation has more than one geography,
+                please specify a `poll_open_date` or organisation_start_date""",
+                code="invalid",
+            )
+        # Large geographies don't perform well with the __relate filter.
+        # If the area of the organisation is larger than "2", use a different
+        # method. "2" is less than the highlands but greater than North
+        # Yorkshire County Council.
+        if og.geography.area < 2:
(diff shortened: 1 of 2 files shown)


every_election/apps/api/filters.py
```diff
@@ -1,9 +1,63 @@
...
+            )
+        # Large geographies don't perform well with the __relate filter.
+        # If the area of the organisation is larger than "2", use a different
+        # method. "2" is less than the highlands but greater than North
+        # Yorkshire County Council.
+        if og.geography.area < 2:
+            # See https://en.wikipedia.org/wiki/DE-9IM for magic string
+            magic_string = "T*****T**"
+            return queryset.filter(
+                division_geography__geography__relate=(
```
GeoWill, line 53: I haven't got my head around how this goes from a queryset of `division_geographies` to a list of `elections`. However I suspect this will miss out Mayors and PCCs which have an `organisation_geography_id` but no `division_geography_id`. See https://github.com/DemocracyClub/EveryElection/blob/0f2e89b0cd7b200da5899c8ef51ba674b3a005ed/every_election/apps/elections/managers.py#L42-L43 for how this was handled in the `ElectionManager`

every_election/apps/api/filters.py
```diff
@@ -1,9 +1,63 @@
...
+                please specify a `poll_open_date` or organisation_start_date""",
+                code="invalid",
+            )
+        # Large geographies don't perform well with the __relate filter.
+        # If the area of the organisation is larger than "2", use a different
+        # method. "2" is less than the highlands but greater than North
+        # Yorkshire County Council.
+        if og.geography.area < 2:
+            # See https://en.wikipedia.org/wiki/DE-9IM for magic string
+            magic_string = "T*****T**"
```
GeoWill, line 51: I think your second `T` should be  `*` i.e. `T********`. It is referring to the relationship between the division `exterior` and the Local Authority `interior`. In the case where the division contains the LA the intersection of these two will be the empty set, i.e. `F`. The other way to think about this is just 'we always want the interiors to overlap, we don't care about anything else'.

every_election/apps/api/filters.py
symroe: > I haven't got my head around how this goes from a queryset of division_geographies

So `queryset` is passed in from Django-Filter, and in this case is given to us from [DRF in the API View](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/api/views.py#L57)

So, it starts as a QS of `Election` models. `division_geography` and `organisation_geography` are fields on `Election`, so the rest if just filtering on that.

I've updated this PR to use `Q` and `|` to get the two geography types now

every_election/apps/api/filters.py
symroe: Managed to push this change in the last commit by mistake...no matter I think I'll squash the commits down before merging anyway. Can you test?

GeoWill: Some tests - they fail with `magic_string = "T*****T**"`
https://github.com/DemocracyClub/EveryElection/pull/1336/commits/57cb53f737063577db6c7211a38105508ebf7232

every_election/apps/api/filters.py
GeoWill: Thanks! That makes much more sense now :slightly_smiling_face:

every_election/apps/api/filters.py
GeoWill: Also had a poke about sense checking it :)

symroe: Cool, I didn't really understand it in the first place :) Thanks for testing it!

PR #1343, opened 2022-02-04
Only use bboverlaps for election_intersects_local_authority_filter

GeoWill:


GeoWill:
This introduces a bug documented here: #1344

every_election/apps/api/filters.py
```diff
@@ -35,34 +35,14 @@ def election_intersects_local_authority_filter(self, queryset, name, value):
             )
 
         try:
-            og = og_qs.get()
+            og_qs.get()
```
GeoWill, line 38: If I've understood https://docs.djangoproject.com/en/4.0/ref/models/querysets/#get this should still raise, and it keeps pyflakes happy.

PR #1431, opened 2022-05-03
Making silly maps for fun and profit

symroe:


symroe:
I had a noodle and made the maps look a bit nicer. 

They generate very quickly for smaller areas but some of the Scottish parliamentary constituencies are too much for my computer to deal with. 

Anyway, they look quite nice I think 

Screenshot from 2023-03-26 18-39-47
Screenshot from 2023-03-26 18-40-14
Screenshot from 2023-03-26 18-40-29
Screenshot from 2023-03-26 18-41-56

PR #1533, opened 2022-07-20
Denormalize current status

symroe:
Before this change, we used to have to perform a SQL JOIN on most election queries (always for public elections). This is because we store a log of moderation events for an election and then filter based on the most recent.

This is neat, but did result in a couple of problems:

1. Each query was slower than it needed to be. It's not a major problem, but this project is 99% read only, so we may as well optimise for that
2. We had to dance around the annotation when adding other queries. For example, you can't use `distinct` and `annotate` at the same time in Django

This PR moves to storing the most current status on the election model itself, meaning we can do away with the complex JOIN. 

While I was at it, I also optimised the elections API endpoint as it was performing way too many queries.


GeoWill:
I'll do a more proper review soon, but after our chat about triggers I realised this was a good use case for them. I'm not actually convinced it's worth the extra dependencies/installation faff, but it was interesting how easy it was to implement using [`django-pgtrigger`](https://django-pgconnection.readthedocs.io/en/latest/). Seems like a useful ecosystem of tools. 

Patch below, extra steps:
- `pip install django-pgtrigger`
- Apply patch
- Add `DATABASES = pgconnection.configure({"DEFAULT": {...})` to `local.py` 
- `python manage.py pgtrigger install`
- `python manage.py pgtrigger ls` should output `elections.Election:protect_current_status       default INSTALLED       ENABLED`

 
```
diff --git a/every_election/apps/elections/models.py b/every_election/apps/elections/models.py
index 3feb028..d69bfde 100644
--- a/every_election/apps/elections/models.py
+++ b/every_election/apps/elections/models.py
@@ -100,6 +100,22 @@ class ModerationStatus(models.Model):
 DEFAULT_STATUS = ModerationStatuses.suggested.value
 
 
+import pgtrigger
+
+
+@pgtrigger.register(
+    pgtrigger.Trigger(
+        name="protect_current_status",
+        operation=pgtrigger.Update,
+        when=pgtrigger.Before,
+        func="""
+            IF NEW.current_status <> OLD.current_status THEN
+                RAISE EXCEPTION 'Do not update current_status directly. Create a new ModerationHistory object instead.';
+            END IF;
+            RETURN NEW;
+        """,
+    )
+)
 class Election(TimeStampedModel):
     """
     An election.
@@ -498,6 +514,7 @@ class ModerationHistory(TimeStampedModel):
     user = models.ForeignKey(User, null=True, blank=True, on_delete=models.SET_NULL)
     notes = models.CharField(blank=True, max_length=255)
 
+    @pgtrigger.ignore("elections.Election:protect_current_status")
     def save(self, **kwargs):
         # if this is the initial status no need to update the related election
         # so return early. This is because the default status is identical on
diff --git a/every_election/settings/base.py b/every_election/settings/base.py
index 7597d9a..d0c1508 100644
--- a/every_election/settings/base.py
+++ b/every_election/settings/base.py
@@ -32,6 +32,7 @@ INSTALLED_APPS = [
     "dc_signup_form",
     "corsheaders",
     "uk_geo_utils",
+    "pgtrigger",
 ]
 
 PROJECT_APPS = [
```

symroe:
I really like the idea of using a trigger for this! However I think we need to think harder about using `pgtrigger`:

1. [This issue about migrations](https://github.com/Opus10/django-pgtrigger/issues/25) raises some interesting questions about how this might work in the long run. I think I'd want migration support before diving in on this, especially as this project is used a little like a library on the AMIs we build...might one of the other installs get confused in some way I can't quite thing about at the moment? (e.g, we're not just managing a single database here)
2. I tried to write a test for this (getting an election, setting the `current_status` and saving, asserting an exception). The test didn't raise an exception (it didn't trigger the trigger) and ~10 minutes of looking in to it didn't come up with a reason why. It might be the way this library works isn't compatible with Django's testing system in some way.

Because of both of these, I wonder if we can just write the SQL to implement the trigger in a migration, and then find a way to disable it in the `save()` method somehow? Or could we use another trigger to update the `current_status` when a moderation status is added too?

Either way, I'd like to know what the problems might be with manually managing the triggers before trying to commit to this library at the moment (although it looks like it could be really useful in a lot of ways with a little improvement)

every_election/apps/api/views.py
```diff
@@ -55,19 +57,33 @@ def geo(self, request, election_id=None, format=None):
         )
 
     def get_queryset(self):
-        queryset = Election.public_objects.all()
-        queryset = queryset.select_related(
+        select_realated = [
```
GeoWill, line 60: Is this a typo?

GeoWill:
These recent changes look good. Glad it caught some errors. The final thought I had - and this might be better punted to an issue - is whether we can follow the pattern [here](https://github.com/DemocracyClub/EveryElection/blob/master/every_election/apps/organisations/models/admin.py#L17) of having a 'Status Problem Manager' that shows any elections where `current_status!=moderation_status_history.latest()`

PR #1796, opened 2023-02-06
Small fixes

symroe:


serverscripts/backup_db_to_s3.sh
```diff
@@ -18,15 +18,19 @@ SHORT_TERM_BUCKET='dc-ee-production-database-backups'
 BACKUP_PG_DUMP_CONNECTION_STRING="-d every_election -h $EE_DATABASE_HOST -Fc"
 export PGPASSWORD=$EE_DATABASE_PASSWORD
 
+pg_dump $BACKUP_PG_DUMP_CONNECTION_STRING > /tmp/ee-backup.dump
+
 if  [ $DAY_OF_MONTH = 01 ] ;
 then
-  pg_dump $BACKUP_PG_DUMP_CONNECTION_STRING | aws s3 cp - \
+   aws s3 cp /tmp/ee-backup.dump \
```
GeoWill, line 25: I don't think this is going to do what you want. The whole bucket has a lifecycle policy (AWS Console -> s3 -> $SHORT_TERM_BUCKET -> `Management` Tab -> `Lifecycle Rules`) to expire objects after 60 days. So (afaict) all the 'first of the month' backups are going to get deleted after that 60 days as well. Two possible fixes are to delete the lifecycle policy, or `cp` the 1st of the month ones to another bucket.

GeoWill (approved):
Looks good apart from some bucket lifecycle faffing

serverscripts/backup_db_to_s3.sh
symroe: Oh good spot! This isn't actually new work in this PR, it's been there for ages. But I'll simplify it here anyway. The backups are only really for client instances to pull from, so not needed for longer than 60 days.

symroe: As per 9d47f3052b4e56f82d0b2b383c354536b8ee0856

Issue #1836, opened 2023-02-23
Missing confirmation page


VirginiaDooley:
### Issue and Steps to Reproduce
After successfully creating an election, the confirmation page does not appear.  The user was not logged in.

PR #1837, opened 2023-02-23
Slack notifications for creating election ids

VirginiaDooley:
Ref https://github.com/DemocracyClub/EveryElection/issues/1836

As Sym expected, the `SLACK_WEBHOOK URL` was missing in settings. I've added that back and set the variable in the parameter store in prod.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  node: circleci/node@5.0.2
+  node: circleci/node@5.1.0
 
 jobs:
   build_and_test:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
 
       - node/install:
-          node-version: '10.19.0'
+          node-version: '19.7.0'
```
symroe, line 28: This looks fine, but it would be useful to know why you needed to do this here. Will we need to do the same in other repos?

VirginiaDooley: Thanks for asking, yes I expect so. I've needed to bump to v16 at a minimum in other repos. `node: circleci/node@5.0.2` is no longer in use and the most recent version of the orb `node: circleci/node@5.1.0` requires v16 at a minimum.

VirginiaDooley:
[Missing election creation confirmations](https://trello.com/c/LjFYuMhN/3290-missing-election-creation-confirmations)

symroe (approved):
Nice one. Don't forget to pick out the CI commit :)

PR #1865, opened 2023-03-10
Management command for adding missing territory codes

symroe:


Bekabyx:
This looks okay to my admittedly untrained eye. Have you had a chance to test it on a recent prod db backup?

symroe:
I have tested this locally on a backup from Jan and it gets 100% of the divisions. We have made more since then but I expect it will work with them in the same way as all the previous ones.

PR #1887, opened 2023-03-23
Subdivided

symroe:
Spike to test an idea of subdividing geographies to improve query performance

### DIFF
--- a/every_election/apps/organisations/tests/factories.py
+++ b/every_election/apps/organisations/tests/factories.py
@@ -7,6 +7,8 @@ from organisations.models import (
     OrganisationDivision,
     OrganisationGeography,
     DivisionGeography,
+    DivisionGeographySubdivided,
+    OrganisationGeographySubdivided,
 )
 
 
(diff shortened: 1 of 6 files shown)


symroe:
@GeoWill I reckon all these commits can be squashed in to one, but I'll keep them here for review.

every_election/apps/organisations/tests/factories.py
```diff
@@ -55,13 +57,28 @@ class Meta:
...
 
+    @factory.post_generation
+    def subdivide(self: DivisionGeography, create, extracted, **kwargs):
+        for geom in self.geography:
+            SubdividedDivisionGeographyFactory(division_geography=self, geography=geom)
```
GeoWill, line 89: Just to check I understand what's going on here. You're creating a row in the `*_subdivided_geography` table for each row in the `*_geography` table. Because it's just tests it doesn't matter that they're not actually subdivided, because the ORM will still treat the relationship as 1-to-many  (eg `Organisation` to `OrganisationGeographySubdivided`), even though the data in this case is 1-to-1.

GeoWill (approved):
Looks good. 
I've also checked out and run through locally. See a massive improvement on something like `ab -n 100 'http://127.0.0.1:8000/api/elections/?coords=51.5010089365,-0.141587600123'`. So figure it's working as intended.

every_election/apps/organisations/tests/factories.py
GeoWill: Ha I wrote this as you were writing [466d399](https://github.com/DemocracyClub/EveryElection/pull/1887/commits/466d3992e87c6a384aa35df557e8dcdce7c35293) :D

GeoWill:
Agreed that squashing before merge is probably right.

PR #1894, opened 2023-03-23
Populate subdivided tables

symroe:
This PR should be safe to deploy as it doesn't do anything to the application. We should deploy it, run the management command and then do another PR/deploy to change the queries used.


every_election/apps/organisations/models/divisions.py
```diff
@@ -139,3 +139,19 @@ class DivisionGeography(models.Model):
     )
     geography = models.MultiPolygonField()
     source = models.CharField(blank=True, max_length=255)
+
+
+class DivisionGeographySubdivided(models.Model):
+    geography = models.PolygonField(db_index=True, spatial_index=True)
+    dg = models.ForeignKey(
```
GeoWill, line 146: Maybe bike-shedding (apols if so) but I'd go for the verbose name here and with the orgs.

Issue #1869, opened 2023-03-14
Improve election radar notifications


VirginiaDooley:
When creating an election in dev, the notifucation in slack causes confusion about whether the election requires action (despite a dead end to the mod queue). 

We improve election radar notifications from dev with a “TEST from dev…” similar to what we’ve done with [confirmation emails in WDIV](https://github.com/DemocracyClub/UK-Polling-Stations/blob/efa3362479381dfedd64ccc044a24a5413813854/polling_stations/apps/file_uploads/models.py#L175).

PR #1965, opened 2023-07-13
Hotfix/specify env in slack posts

VirginiaDooley:
Ref https://github.com/DemocracyClub/EveryElection/issues/1869

This change concatenates a test message to the core message for any notification posted to Slack. It also posts test messages from `dev` or `staging` to a new webhook url in `#bots`.

I've had to address some security vulnerabilities to get to a passing state in CI; a more comprehensive set up upgrades will come shortly with https://trello.com/c/gI6suc2g/3416-ee-django-42-upgrades

To test, make sure you have `SERVER_ENVIRONMENT` and `TEST_SLACK_WEBHOOK_URL` set in your `local.py` and create an election. Message me for the webhook url to add to settings. A slack post similar to below should appear in #bots.

Screenshot 2023-07-18 at 3 50 57 PM

### DIFF
--- a/every_election/apps/election_snooper/helpers.py
+++ b/every_election/apps/election_snooper/helpers.py
@@ -1,4 +1,5 @@
 import json
+import os
 import textwrap
 import warnings
 
@@ -8,14 +9,26 @@ import requests
 
 
 def post_to_slack(message):
-    if not hasattr(settings, "SLACK_WEBHOOK_URL"):
+    env = os.getenv("SERVER_ENVIRONMENT", getattr(settings, "SERVER_ENVIRONMENT", None))
+    if env in ["test", "development", "staging"]:
+        prefix = "TEST for {} environment: ".format(env)
+        message = prefix + message
+    url = settings.SLACK_WEBHOOK_URL
+
+    if not hasattr(settings, "SLACK_WEBHOOK_URL") and env == "production":
         warnings.warn("settings.SLACK_WEBHOOK_URL is not set")
         return
+    if not hasattr(settings, "SLACK_WEBHOOK_URL") and env in [
+        "test",
+        "development",
+        "staging",
+    ]:
+        print("SLACK_WEBHOOK_URL is not set")
 
     payload = {
         "icon_emoji": ":satellite_antenna:",
         "username": "Election Radar",
         "text": textwrap.dedent(message),
     }
-    url = settings.SLACK_WEBHOOK_URL
+
     requests.post(url, json.dumps(payload), timeout=2)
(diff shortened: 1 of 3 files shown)


VirginiaDooley:
[Post-election tasks](https://trello.com/c/LzXzoxbe/3391-post-election-tasks)

symroe:
This is useful, but can you also [set up a new webhook](https://democracyclub.slack.com/apps/A0F7XDUAZ-incoming-webhooks?tab=settings&next_id=0) for the dev and stage envs that posts to `#bots` rather than `#elections`?

Can you then change the URL in the dev and stage accounts to use the bots URL rather than the live one (but keep prod the same)?

every_election/apps/election_snooper/helpers.py
```diff
@@ -8,6 +9,12 @@
 
 
 def post_to_slack(message):
+    env = os.getenv("SERVER_ENVIRONMENT", getattr(settings, "SERVER_ENVIRONMENT", None))
+
+    if env == "test" or env == "development" or env == "staging":
```
symroe, line 14: I think these can be shortened to `if env in ["test", ...]`

VirginiaDooley:
> This is useful, but can you also [set up a new webhook](https://democracyclub.slack.com/apps/A0F7XDUAZ-incoming-webhooks?tab=settings&next_id=0) for the dev and stage envs that posts to `#bots` rather than `#elections`?
> 
> Can you then change the URL in the dev and stage accounts to use the bots URL rather than the live one (but keep prod the same)?
> 
> 
Happy to, but integrations are restricted. Can you provide access?

symroe:
Forgive me if I'm missing something here, but do we need `TEST_SLACK_WEBHOOK_URL`?

If we just have `SLACK_WEBHOOK_URL` then we can change the URL in the dev and stage accounts to be the testing URL, with the prod one being the prod URL.

We also need to make sure that it works if no URL is set, e.g in local dev. Maybe print to the command line?

every_election/settings/local.example.py
```diff
@@ -12,6 +12,10 @@
 # google custom search API key
 GCS_API_KEY = ""
 
+# don't use the real url in prod
```
GeoWill, line 15: Do you mean 'don't use the real url in local dev'? 
I assume pros is where we _do_ want to use the real url.

VirginiaDooley: That is what I meant. I'll edit. Thank you!

VirginiaDooley:
> URL

Yup, you're right. I'll make those changes. Thanks.

symroe (approved):

Issue #1972, opened 2023-07-25
Move all cancellation logic to EE


symroe:
At the moment we can either cancel elections in EE or have YNR mark them as uncontested.

This gets complex quickly and means we have two sources of truth.

We also don't really present the reason for cancellation in EE, rather we have a boolean field and some other ways (metadata, seats contested) to work out what's going on.

The devs.DC API manages some of this logic for us:

https://github.com/DemocracyClub/aggregator-api/blob/master/api/endpoints/v1/voting_information/stitcher.py#L87-L105

We can move this 'up' into EE and simplify all downstream logic.

This looks like:

```[tasklist]
### Tasks
- [x] Move the CANCELLATION_REASONS dict in the above devs.DC link to EE
- [x] Add cancellation reason to Election model
- [x] Add the reasons as a drop down in the admin interface.
- [x] Migrate all previously cancelled elections to use cancellation reason (might be manual)
- [x] Expose CANCELLATION_REASON in the API
- [x] Update WDIV to consume this API and re-publish to devs.DC
- [x] Update devs.DC API to show cancellation reason from WDIV's API
- [ ] https://github.com/DemocracyClub/EveryElection/issues/1685
- [ ] https://github.com/DemocracyClub/EveryElection/issues/1939
- [ ] https://github.com/DemocracyClub/EveryElection/issues/1082
- [ ] https://github.com/DemocracyClub/EveryElection/issues/656
```

We're going to need to think about this a little more, but that's the basic outline.

Ideally we'd find a way for YNR to push new information to EE when we lock an uncontested ballot, but this might be too complex for the initial version. Better to improve the admin interface to allow users to to this manually, as this needs to be done anyway.

Once this is done, we can clean up the logic from the widget, WDIV, WCIVF and the Devs.DC API.

Bekabyx:
Finished in DemocracyClub/EveryElection#2004

PR #1977, opened 2023-08-04
Cancellation reason groundwork

Bekabyx:
Ref #1972 

This PR addresses three separate cards in Asana - normally I'm in the habit of 1 card per PR but the cards were actually too granular and it felt redundant to split across multiple PRs.

Asana cards: 
[Move the CANCELLATION_REASONS dict from Devs DC to EE](https://app.asana.com/0/1204880927741389/1205170296632723) (instead implemented this as an enum)
[Add cancellation reason to Election model](https://app.asana.com/0/1204880927741389/1205170296632724)
[Add cancellation reason as a dropdown on the admin interface](https://app.asana.com/0/1204880927741389/1205170296632725)

I also upgraded uk-election-ids to the latest version to fix a typo in one of the ID requirements.

```[tasklist]
### PR Checklist
- [x] References a specific issue or if not, describes the bug or feature in detail
- [x] Note what card in Asana this work relates to
- [ ] Instructions for how reviewers can test the code locally
- [ ] Tests have been added and/or updated
- [x] Did I added or updated any new dependencies? Explain why.

```


every_election/apps/elections/models.py
```diff
@@ -29,6 +29,13 @@
 from .managers import PrivateElectionsManager, PublicElectionsManager
 
 
+class ElectionCancellationReason(models.TextChoices):
+    NO_CANDIDATES = "NC", "No candidates"
```
symroe, line 33: Could you make the values the verbose version rather than two letters? e.g `NO_CANDIDATES` rather than `NC`? I think this will make things easier for ourselves when debugging this in future.

Bekabyx: Agreed - I've updated this now!

symroe (approved):

PR #1978, opened 2023-08-07
Fix and test id creator

symroe:
PR with some drive by work.

Firstly, there was a bug in the ID creator where `>` and `<` had been flipped. I expect this was related to a mistake during the Ruff work.

But, that wouldn't have ended up being deployed if we had more tests. So I made them.

This involved using Playwright. I wanted to make a basic test, so I tested the JS date picker stuff first, and ended up finding some bugs in that. And on we go :sunglasses:   

```[tasklist]
### PR Checklist
- [x] Note what card in Trello this work relates to
- [x] Instructions for how reviewers can test the code locally
- [x] Tests have been added and/or updated
- [x] Did I explain all possible solutions and why I chose the one I did?
- [x] Added any comments to make new functions clearer
- [x] Have I rebased with the latest version of master?
- [x] Added PR labels
```


every_election/apps/elections/tests/test_id_creator_wizard.py
```diff
@@ -0,0 +1,130 @@
...
+        "UK elections are almost always on a Thursday"
+    )
+    page.locator("#id_date-date_0").fill("5")
+    page.locator("#id_date-date_0").blur()
+    expect(page.locator("#id_creator_date")).not_to_contain_text(
+        "UK elections are almost always on a Thursday"
+    )
+
+
+def test_full_id_creation(page, live_server, id_creator_data, settings):
```
Bekabyx, line 88: Is there a reason this test doesn't use the `BaseElectionCreatorMixin` instead of having its own pytest fixture? I ask only because there's a lot of overlap in content between the two.

symroe: Yes good point. This was initially because there is (or I thought there was) a bug when mixing pytest functions, `live_server` and `UnitTest`. I'll have a play around to see if I can DRY it up

symroe: @Bekabyx Right, I think I might be defeated on this.

Basically, `live_server` spins up a new thread and does stuff with the database that's...confusing. Will and I spent a couple of hours on this yesterday and I can't get the tests to pass by changing it to use the same mixin.

Are you ok with this duplication?

Bekabyx: Yep, it's not a blocker for me. It felt worth mentioning but if this is how we have to do it then that's how we'll do it 😁

Bekabyx (approved):

every_election/apps/elections/tests/test_id_creator_wizard.py
symroe: Thanks!

PR #1984, opened 2023-08-21
Document how to update elections.json

Bekabyx:
This held me back for quite a while in some recent WDIV work so I thought I'd bulk out the docs 🙃


VirginiaDooley (approved):
Thank you!

PR #1991, opened 2023-09-07
Ouroboros 2.0

symroe:
This is a work-not-so-in-progress of a script to replace the old `dumpdata`/`loaddata` system with the API and `modified` queries. 

It's designed to to run often on EE consumers and only import recently changed elections, once most of the data is synced. 

It doesn't get around the need to seed the loca database with other data like postcodes, boundaries, organisations, etc. This is simply for the "normal" day to day changes. 

At the moment the main problem with it is that it's not possible to perfectly replicate the remote database based on the API. I've found two places where this is true:

1. Explanations
2. Metadata

The reason we can't perfectly replicate the objects is because have "title" and PK fields for both that aren't in the API. The content isn't unique (e.g we can have duplicate metadata objects with different titles) and we don't know how to re-create the whole object.

To get around this I started making up fake titles. This works in terms of the API and end user content being the same, but if you were to use the admin interface it wouldn't look the same as the remote source. 

We could expose this in the API somehow, or just not worry about it.

---
  - https://app.asana.com/0/0/1205127245342547


VirginiaDooley (approved):

every_election/apps/elections/management/commands/sync_elections.py
```diff
@@ -1,14 +1,271 @@
...
+        while self.url:
+            print(last_modified)
+            req = requests.get(self.url)
+            print(req.url)
+            req.raise_for_status()
+            resp_json = req.json()
+            unordered_results = resp_json["results"]
+            results = sorted(
+                unordered_results, key=lambda d: d["election_id"].count(".")
+            )
```
GeoWill, line 268: These lines mean we process parents first right? i.e. go through `group_type`  from `election` -> `organisation` -> `subtype` -> `None`, (and by-elections last of all). 
Does this matter when the returned list is split across pages, and parents are on subsequent pages? I think this is this what's handled by the `ParentDoesNotExists` exception and the direct import of the parent, but wanted to be sure.

GeoWill (commented):
Basically looks good, but I have found a way of breaking it. This is to run sync on a db where the official identifiers for divisions have changed. This happens when we backport the codes from boundaryline after an election. I'm not sure exactly what the solution for this is. Maybe it's just that a new image is required, as with new divisionsets and new organisations.

GeoWill:
Sorry. It didn't break as in raise an exception, it broke as in gave messages like `Can't add local.cornwall.falmouth-boslowick.2021-05-06 because the Division doesn't exits.` when the division does exist. 

It did raise an exception because an org doesn't exist, but I think that's expected. 

To reproduce: 

```shell
aws --profile prod-ee-dc s3 cp s3://dc-ee-production-database-backups/every_election/2023-01-20-10-backup.dump ~/Downloads/ 
sudo -u postgres dropdb every_election --if-exists
sudo -u postgres createdb every_election
psql -U postgres -d every_election -c "create extension postgis;"
pg_restore -c --if-exists -U dc -d every_election -c --if-exists -U dc -d every_election ~/Downloads/2023-01-20-10-backup.dump
```

symroe:
Oh interesting ok. What do you think is best here? Feels like we want to know (via snetry / email) if an embedded EE is out of date like this?

GeoWill:
Yeah - I wonder if a sentry warning is a good model for this? If we can say where it comes from (i.e. where that version of EE is embedded!) I guess that might be complicated though?

symroe:
I've added some more test for this after realising that we weren't going to deal with the case of changing the parl divisionset dates. That is one place where we want embedded versions to update quickly without having to do a re-deploy (because until they're created in the onboard WDIV EE the general election isn't going to show up in the API)

PR #1996, opened 2023-09-27
Clarifying City of London elections

pmk01:
I should have done this two years ago.


VirginiaDooley (approved):

symroe:
@GeoWill is there a reason not to rename the organisations?  I remember why we didn't do this at the time, but I suspect There Are Reasons™

GeoWill:
Not sure. Other cases I can think of the change of name accompanies a change in the type of thing - i.e. district to unitary or similar. So there is actually a new thing. In the case of Welsh Assembly -> Sennedd I think you could also make this case, which is why the org type `sennedd` was added. In this case they're both lumped as `local-authority` with subtype `CC`. Which is probably right, but they have different election names. All that to say I think it's probably fine...

symroe:
Ok thanks! I'll get #2012 in and then merge this

PR #2000, opened 2023-10-16
Command to import divisions from geodata

GeoWill:
This command really exists to do the work of importing https://pages.mysociety.org/2025-constituencies/datasets/parliament_con_2025/latest

It probably isn't that useful in day to day, but it means we can re-import the boundaries quite quickly if needed - i.e. after importing the first time, we can delete the divisionset, divisions and division geographies, and re-run this command.

Hopefully it is a useful reference and/or can be adapted for other boundary updates that don't map on to the ECO process.

### DIFF
--- a/every_election/apps/organisations/management/commands/import_divisionset_from_geodata.py
+++ b/every_election/apps/organisations/management/commands/import_divisionset_from_geodata.py
@@ -1,0 +1,208 @@
+import datetime
+
+from core.mixins import ReadFromFileMixin
+from django.contrib.gis.gdal import DataSource
+from django.core.management import BaseCommand
+from django.db import transaction
+from django.utils.text import slugify
+from organisations.models import (
+    DivisionGeography,
+    Organisation,
+    OrganisationDivision,
+    OrganisationDivisionSet,
+)
+from storage.shapefile import pre_process_layer
+
+TERRITORY_LOOKUP = {
+    "england": "ENG",
+    "northern ireland": "NIR",
+    "scotland": "SCT",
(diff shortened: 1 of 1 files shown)


VirginiaDooley (approved):

every_election/apps/organisations/management/commands/import_divisionset_from_geodata.py
```diff
@@ -0,0 +1,208 @@
...
+    def get_territory_code(self, feature):
+        territory_code = feature[self.territory_field].value
+        if territory_code not in ["ENG", "NIR", "SCT", "WLS"]:
+            territory_code = TERRITORY_LOOKUP[territory_code.lower()]
+        return territory_code
+
+    def create_division(self, feature):
+        identifier = self.get_id(feature)
+        name = self.get_name(feature)
+        return OrganisationDivision(
```
symroe, line 142: Do you need to set `seats_total` here? At least default it to `1` rather than `None`?

GeoWill: OK. I went for the default of None, as it was 'obviously' wrong, rather than potentially subtly wrong. Maybe better to require either a field name or fixed value.

symroe: Oh good point. I guess we need to remember to set them, maybe adding some print statement at the end of the command?

symroe (approved):
Not run this, but it looks good to me and I reckon the only real test is the data showing up on prod :+1:

PR #2005, opened 2023-11-01
Use Django>4.2.5 to address safety vulnerability

VirginiaDooley:
ADVISORY
Django 3.2.21, 4.1.11 and 4.2.5 include a fix for CVE-2023-41164: Potential denial of service vulnerability in django.utils.encoding.uri_to_iri().
https://www.djangoproject.com/weblog/2023/sep/04/security-releases

https://data.safetycli.com/v/60956/f17/


requirements/base.txt
```diff
@@ -1,6 +1,6 @@
 beautifulsoup4==4.12.1
 boto3==1.28.7
-Django==4.2.6
+Django>4.2.5
```
symroe, line 3: With a simple `>` this will also include Django 5, so I think we need to either use `==` or `<=5` here

VirginiaDooley: Fixed!

symroe: I think that latest change _downgrades_ Django?

symroe: Oh sorry, `4.2.5` is the version. Never mind!

symroe: Actually I still think it's cleaner to say "less than Django 5.0" than "less than or equal to 4.2.5"

VirginiaDooley: less than 5.0?

Bekabyx:
I wonder if the problem in the pipeline is actually due to cached packages?

I say this because now I've upgraded my local environment to use Django 4.2.6 I'm not getting anything on `safety check`. Previously I was on 4.2.3 which raised the error you've described in your PR body.

VirginiaDooley:
Addressed in https://github.com/DemocracyClub/EveryElection/pull/2006

PR #2009, opened 2023-11-08
Seats contested in ID creator

### LINKED ISSUE #1406: Enable seats_contested to be modified at front end
This would be especially useful b/c I would like to modify the seats_contested and upload a notice of election document at the same time.

symroe:
Fixes #1406 

It's not quite as good as @pmk01's drawing in the issue, but it's something

image

WIP commit: the background is I did this work and then managed to `git reset --hard` when checking out another PR :sob: 

Pycharm to the rescue, but the commits got messed up. Pushed so I don't screw up again. I need to:

```[tasklist]
### Tasks
- [ ] Fix up the commits nicely
- [x] Write /update tests
- [x] Figure out why the JS screws up the form validation (it's currently commented out)
``` 

For the time being, some manual testing would be welcome please!

---
  - https://app.asana.com/0/0/1205127245342563


pmk01:
What happens when you selected 'scheduled' or 'by-election'? Ideally I think they would default to 1, with an option to set all to max.

symroe:
> What happens when you selected 'scheduled' or 'by-election'? Ideally I think they would default to 1, with an option to set all to max.

Yes, there's some interaction like that I've added but not fully implemented yet. The "mark all" button will mark them all as "scheduled" with the `seats_contested` value being set to the seats total for the division. 

Setting to `1` when either option is manually selected is neat, I'll add that :+1:

pmk01:
Bear in mind that about 50-60% of wards will only have one seat up each year, so setting them all to max as default would mean a lot of extra work to put them down to 1. I suggest 'mark all' defaults to 1, and then a separate button can move them all to max if needed. Does that make sense?

symroe:
Updated to support group types and improved the interaction a little as per Peter's requests

[Screencast from 2023-11-11 13-19-26.webm](https://github.com/DemocracyClub/EveryElection/assets/242329/041f02f7-7c47-4c52-a255-118f508f8fd6)

PR #2012, opened 2023-11-13
ID creator bugfixes

symroe:
The problem here was that we had two list: initial and kwargs and I was only sorting one of them. 

This change sorts the divisions in the database and then builds both lists up from the pre-sorted divisions.

### DIFF
--- a/every_election/apps/elections/forms.py
+++ b/every_election/apps/elections/forms.py
@@ -1,10 +1,9 @@
-from typing import List, Union
+from typing import Union
 
 from dc_utils import forms as dc_forms
 from django import forms
 from django.core.exceptions import ValidationError
-from django.db import models
-from django.db.models import Q
+from django.db.models import F, Q, QuerySet
 from organisations.models import (
     Organisation,
     OrganisationDivision,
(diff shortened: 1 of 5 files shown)


every_election/apps/elections/forms.py
```diff
@@ -159,39 +158,59 @@ def clean(self):
...
+
+                divisions_qs: Union[
+                    QuerySet[OrganisationDivision], DivisionManager
+                ] = (
+                    div_set.divisions.select_related(
+                        "divisionset__organisation"
+                    )
+                    .filter(**division_filter_args)
+                    .annotate(group=F(group_field))
+                    .order_by("group")
```
GeoWill, line 204: I think this db ordering is replacing:
```
kwargs["initial"] = sorted(
            initial_data, key=lambda el: (el["group"], el["division_name"])
        )
```

Is it OK that we don't have the sort by `division_name` as well?

GeoWill (commented):
I haven't tried this out, but apart from my question it looks sensible, assuming you've tried it.

every_election/apps/elections/forms.py
symroe: Oh good catch, thanks. At least this change means that _both_ lists would be equally out of order, but I've added `name` in a fixup commit to this

GeoWill (approved):

PR #2032, opened 2023-12-07
Feature/boundary review frontend

awdem:
Frontend for Boundary Reviews:

Completed:
- List view
     - Groups BRs by organisation
     - Has a filter component with filters for:
         - org
         - status
- Detail view
- Adds generic title property to BR model that determines what title to display for a review (w/ test)
- Adds a link to the list view from supported orgs page

Not yet implemented: 
- Adding maps to detail view

Notes:

- I wasn't sure if the filter logic should stay in filters.py or be added to the BR queryset model, but can easily move it over if that is better practice. Likewise, I wasn't sure if I should make tests for each filter option?
- I can add more properties to filter by or to display in the views if appropriate too

Testing:

Navigate to [supported orgs](http://localhost:8000/organisations/) and click on boundary reviews
Run `pytest`

### DIFF
--- a/every_election/apps/organisations/templates/organisations/organisationboundaryreview_detail.html
+++ b/every_election/apps/organisations/templates/organisations/organisationboundaryreview_detail.html
@@ -1,0 +1,40 @@
+{% extends "base.html" %}
+{% block page_title %}Boundary Review Summary{% endblock page_title %}
+
+{% block content %}
+<div class="ds-card">
+	<div class="ds-card-body">
+		<h1>
+			{{boundary_review_name}}
+		</h1>
+
+		<dl class="ds-descriptions">
+			<div>
+				<dt>Organisation</dt>
+				<dd><a href={{boundary_review.organisation.get_absolute_url}}>{{boundary_review.organisation.name}}</a></dd>
+			</div>
+			<div>
+				<dt>Status</dt>
+				<dd>{{ boundary_review.get_status_display }}</dd>
+			</div>
+			<div>
+				<dt>Consultation</dt>
+				{% if boundary_review.consultation_url %}
+					<dd>Read <a href="{{ boundary_review.consultation_url }}">the consultation</a> for this review</dd>
+				{% else %}
+					<dd>Consultation Link Missing</dd>
+				{% endif %}
+				</div> 
+				<div>
+					<dt>Legislation</dt>
+				{% if boundary_review.legislation_url%}
+					<dd>Read <a href="{{ boundary_review.legislation_url }}">the legislation</a> for this review</dd>
+				{% else %}
+					<dd>Legislation Link Missing</dd>
+				{% endif %}
+			</div>
+		</dl>
+	</div class="ds-card-body">
+</div>
+
+{% endblock content %}
--- a/every_election/apps/organisations/templates/organisations/organisationboundaryreview_detail.html
+++ b/every_election/apps/organisations/templates/organisations/organisationboundaryreview_detail.html
@@ -5,7 +5,7 @@
 <div class="ds-card">
 	<div class="ds-card-body">
 		<h1>
-			{{boundary_review_name}}
+			{{boundary_review.generic_title}}
 		</h1>
 
 		<dl class="ds-descriptions">
--- a/every_election/apps/organisations/templates/organisations/organisationboundaryreviews_list.html
+++ b/every_election/apps/organisations/templates/organisations/organisationboundaryreviews_list.html
@@ -1,0 +1,15 @@
+{% extends "base.html" %}
+{% block page_title %}Boundary Reviews{% endblock page_title %}
+
+{% block content %}
+	{% for review in object_list%}
+		<h3>
+			<a href="{% url 'single_boundary_review_view' review.id %}">{{review.lgbce_review_title}}</a>
+		</h3>
+		<li><strong>Organisation</strong>: <a href={{review.divisionset.organisation.get_absolute_url}}>{{review.divisionset.organisation.name}}</a></li>
+		<li><strong>Consultation</strong>: <a href={{review.consultation_url}}>Link</a></li>
+		<li><strong>Legislation</strong>: <a href={{review.legislation_url}}>Link</a></li>
+
+	{% endfor %}
+
+{% endblock content %}
--- a/every_election/apps/organisations/templates/organisations/organisationboundaryreviews_list.html
+++ b/every_election/apps/organisations/templates/organisations/organisationboundaryreviews_list.html
@@ -2,13 +2,45 @@
 {% block page_title %}Boundary Reviews{% endblock page_title %}
 
 {% block content %}
-	{% for review in object_list%}
-		<h3>
-			<a href="{% url 'single_boundary_review_view' review.id %}">{{review.generic_title}}</a>
-		</h3>
-		<li><strong>Organisation</strong>: <a href={{review.divisionset.organisation.get_absolute_url}}>{{review.divisionset.organisation.name}}</a></li>
-		<li><strong>Consultation</strong>: <a href={{review.consultation_url}}>Link</a></li>
-		<li><strong>Legislation</strong>: <a href={{review.legislation_url}}>Link</a></li>
+
+<style>
+	.boundary_reviews_table {
+		font-size: 1rem;
+	}
+	.boundary_reviews_table td:first-child {
+		width: 50%;
+	}
+</style>
+
+{%regroup boundary_reviews by organisation as boundary_reviews_by_organisation%}
+{% for group in boundary_reviews_by_organisation %}
+	<h5>{{group.grouper}}</h5>
+	<div class="ds-table ds-bordered">
+		<table class="boundary_reviews_table">
+			<tr>
+				<th>Review</th>
+				<th>Consultation</th>
+				<th>Legislation</th>
+				<th>Status</th>
+			</tr>
+			{% for review in group.list%}
+				<tr>
+					<td><a href="{% url 'single_boundary_review_view' review.id %}">{{review.generic_title}}</a></td>
+					{%if review.consultation_url%}
+						<td>Read the <a href={{review.consultation_url}}>Consultation</a></td>
+					{%else%}	
+						<td>Consultation Link Missing</a></td>
+					{%endif%}
+					{%if review.legislation_url%}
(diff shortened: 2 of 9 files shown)


every_election/apps/organisations/templates/organisations/organisationboundaryreviews_list.html
```diff
@@ -0,0 +1,76 @@
...
+
+			<li class="current">Page {{ page_obj.number }} of {{ page_obj.paginator.num_pages }}.</li>
+
+			{% if page_obj.has_next %}
+
+					<li class="pagination-next"><a href="?page={{ page_obj.next_page_number }}">Next</a></li>
+			{% endif %}
+	</ul>
+{% endif %}
+{% endblock content %}2
```
GeoWill, line 76: Stray '2'

every_election/apps/organisations/templates/organisations/organisationboundaryreviews_list.html
```diff
@@ -0,0 +1,76 @@
...
+				<th>Review</th>
+				<th>Consultation</th>
+				<th>Legislation</th>
+				<th>Status</th>
+			</tr>
+			{% for review in group.list%}
+				<tr>
+					<td><a href="{% url 'single_boundary_review_view' review.id %}">{{review.generic_title}}</a></td>
+					<td><a href={{review.consultation_url}}>Link</a></td>
+					<td><a href={{review.legislation_url}}>Link</a></td>
```
GeoWill, line 53: These might not be set, so we should only show a link if they exist.

every_election/apps/organisations/templates/organisations/organisationboundaryreview_detail.html
```diff
@@ -0,0 +1,32 @@
...
+				<dt>Status</dt>
+				<dd>{{ boundary_review.status }}</dd>
+			</div>
+			<div>
+				<dt>Consultation</dt>
+				<dd>Read <a href="{{ boundary_review.consultation_url }}">the consultation</a> for this review</dd>
+			</div> 
+			<div>
+				<dt>Legislation</dt>
+				<dd>Read <a href="{{ boundary_review.legislation_url }}">the legislation</a> for this review</dd>
```
GeoWill, line 26: These urls might not be set (though the legislation _should_ exist) so we should have a conditional about whether so show the link

every_election/apps/organisations/views/public.py
```diff
@@ -71,3 +71,33 @@ def get_context_data(self, **kwargs):
                 "api:organisation-geo", "json"
             )
         return context
+
+
+class AllBoundaryReviewsView(ListView):
+    template_name = "organisations/organisationboundaryreviews_list.html"
+    model = OrganisationBoundaryReview
+    context_object_name = "boundary_reviews"
+    paginate_by = 50
```
GeoWill, line 80: See comment above re. pagination.

GeoWill (changes requested):
This is looking good.
I like the layout of the list page. The filters are great. The detail page is a good base for adding more, um, detail... like maps, after May/GE.

There are a few changes that I've made comments about inline.

In answer to your questions:

> I wasn't sure if the filter logic should stay in filters.py or be added to the BR queryset model, but can easily move it over if that is better practice. Likewise, I wasn't sure if I should make tests for each filter option?

I'm not actually sure myself, but I think what you've done works (and looks) fine to me. @symroe might have more insight on this one. 

> I can add more properties to filter by or to display in the views if appropriate too

I don't think this is needed

every_election/apps/organisations/templates/organisations/organisationboundaryreview_detail.html
```diff
@@ -0,0 +1,32 @@
...
+<div class="ds-card">
+	<div class="ds-card-body">
+		<h1>
+			{{boundary_review.generic_title}}
+		</h1>
+
+		<dl class="ds-descriptions">
+			<div>
+				<dt>Organisation</dt>
+				<dd><a href={{boundary_review.divisionset.organisation.get_absolute_url}}>{{boundary_review.divisionset.organisation.name}}</a></dd>
```
awdem, line 14: I noticed that BR that are currently in review weren't showing an org link or name because they don't have a division set yet so I removed the divisionset steps on these lines. Now they just use boundary_review.organisation for name and absolute_url.

every_election/apps/organisations/templates/organisations/organisationboundaryreview_detail.html
```diff
@@ -0,0 +1,32 @@
...
+			</div>
+			<div>
+				<dt>Status</dt>
+				<dd>{{ boundary_review.status }}</dd>
```
awdem, line 18: Changed this to get_status_display to use the 'human-readable' status choices

every_election/apps/organisations/templates/organisations/organisationboundaryreviews_list.html
```diff
@@ -0,0 +1,76 @@
...
+					<li class="pagination-previous"><a href="?page={{ page_obj.previous_page_number }}">Previous</a></li>
+			{% endif %}
+
+			<li class="current">Page {{ page_obj.number }} of {{ page_obj.paginator.num_pages }}.</li>
+
+			{% if page_obj.has_next %}
+
+					<li class="pagination-next"><a href="?page={{ page_obj.next_page_number }}">Next</a></li>
+			{% endif %}
+	</ul>
```
GeoWill, line 74: This isn't working. Remove as discussed.

every_election/apps/organisations/templates/organisations/organisationboundaryreviews_list.html
```diff
@@ -0,0 +1,76 @@
...
+
+<aside class="ds-filter" aria-labelledby="filter-label">
+  <div class="ds-filter-cluster">
+    <ul>
+      <li id="filter-label" class="ds-filter-label" aria-hidden="true">Filter:</li>
+      <li><a href="{% url "boundary_reviews_view" %}" {% if request.get_full_path == "/organisations/boundary_reviews/" %}aria-current="true" {% endif %}>All</a></li>
+    </ul>
+  </div>
+	<form>
+		<details {% if filter.data %}open=""{% endif %}>
```
GeoWill, line 22: Doesn't need a detail, or a distinction between filters and advanced filters/

every_election/apps/organisations/templates/organisations/organisationboundaryreviews_list.html
symroe: This isn't a project that's really been optimised / designed for accessibility, but as a general rule the text of a link should be descriptive. This is because screen reader software can read out the links on the page for quick navigation. On this page the user will hear "link: Link. link: Link" when they could hear "link: Consultation, link: Legislation". I know these are in the table header, but I think it's still better to use descriptive links.

symroe:
> > I wasn't sure if the filter logic should stay in filters.py or be added to the BR queryset model, but can easily move it over if that is better practice. Likewise, I wasn't sure if I should make tests for each filter option?
> 
> I'm not actually sure myself, but I think what you've done works (and looks) fine to me. @symroe might have more insight on this one.

Are you talking about `organisation_filter` and `status_filter` here? If so, I think it's fine to be in the filter as is. I don't have hard rules, but broadly speaking if the query is above some complexity threashold (undefined!) then it's better on the model. In both of these cases you're applying a simple filter in a way that I would expect the filter to work, so I don't really see the value in having them on the model. `queryset.filter(status=value)` is just as simple to understand as `queryset.for_status(value)` or whatever.

awdem:
> > > I wasn't sure if the filter logic should stay in filters.py or be added to the BR queryset model, but can easily move it over if that is better practice. Likewise, I wasn't sure if I should make tests for each filter option?
> > 
> > 
> > I'm not actually sure myself, but I think what you've done works (and looks) fine to me. @symroe might have more insight on this one.
> 
> Are you talking about `organisation_filter` and `status_filter` here? If so, I think it's fine to be in the filter as is. I don't have hard rules, but broadly speaking if the query is above some complexity threashold (undefined!) then it's better on the model. In both of these cases you're applying a simple filter in a way that I would expect the filter to work, so I don't really see the value in having them on the model. `queryset.filter(status=value)` is just as simple to understand as `queryset.for_status(value)` or whatever.

Yes those are the two I meant. That makes sense! I was thinking that it could make sense to put them on the on model if they end up being used elsewhere in the codebase, but neither filter is complicated enough that it would even save that much on repetition.

every_election/apps/organisations/templates/organisations/organisationboundaryreview_detail.html
```diff
@@ -0,0 +1,40 @@
...
+			<div>
+				<dt>Status</dt>
+				<dd>{{ boundary_review.get_status_display }}</dd>
+			</div>
+			<div>
+				<dt>Consultation</dt>
+				{% if boundary_review.consultation_url %}
+					<dd>Read <a href="{{ boundary_review.consultation_url }}">the consultation</a> for this review</dd>
+				{% else %}
+					<dd>None</dd>
```
GeoWill, line 25: I think something like 'link missing' might be better here - when I look at the page and see `Consultation: None` it makes me think there wasn't a consultation/there's no legislation.

every_election/apps/organisations/templates/organisations/organisationboundaryreviews_list.html
```diff
@@ -0,0 +1,67 @@
...
+<h2> Boundary Reviews </h2>
+
+<aside class="ds-filter" aria-labelledby="filter-label">
+	<form>
+		<div class="ds-filter-cluster">
+			<ul>
+				<li id="filter-label" class="ds-filter-label" aria-hidden="true">Filter:</li>
+				<li><a href="{% url "boundary_reviews_view" %}" {% if request.get_full_path == "/organisations/boundary_reviews/" %}aria-current="true" {% endif %}>All</a></li>
+			</ul>
+		</div>
```
GeoWill, line 24: The 'Filter: All' row looks a bit off to me. There's no choice there, so it's a little confusing. I wonder if it would make more sense to have a heading that just says 'Filters' instead 
image

GeoWill (commented):
Looks great, and all working as expected. 
Couple more thoughts from playing around with it, happy for them to be fixed and all the fixup commits squashed.

GeoWill (approved):

PR #2053, opened 2024-01-25
Don't invalidate cache for each class instance

symroe:
This improves performance when creating a lot of elections

Before

image

After 
Screenshot from 2024-01-25 15-01-13

:sunglasses:

### DIFF
--- a/every_election/apps/elections/utils.py
+++ b/every_election/apps/elections/utils.py
@@ -29,6 +29,14 @@ CACHE = {
 }
 
 
+def reset_cache():
+    global CACHE
+    NEW_CACHE = {}
+    for key in CACHE:
+        NEW_CACHE[key] = {}
+    CACHE = NEW_CACHE
+
+
 def get_voter_id_requirement(election: Election) -> Optional[str]:
     """
     Given an Election object, if eligible for voter ID return the related voter ID legislation code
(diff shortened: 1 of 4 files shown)


VirginiaDooley (approved):

every_election/apps/elections/utils.py
```diff
@@ -63,9 +71,9 @@ def get_cached_election_subtype(election_type):
 
 def get_cached_valid_election_types(organisation):
     if organisation not in CACHE["valid_election_types"]:
-        CACHE["valid_election_types"][
-            organisation
-        ] = organisation.election_types.all()
+        CACHE["valid_election_types"][organisation] = list(
+            organisation.election_types.all()
+        )
```
GeoWill, line 76: Could you explain why you made this change? I figure it's essentially forcing the evaluation of the QuerySet, but am curious as to why it's preferable to do that here.

GeoWill (approved):
LGTM. One question for my own learning!

every_election/apps/elections/utils.py
symroe: Actually this isn't needed. It's left over from when I thought it was being evaluated / executed each time. I've removed it and force pushed.

PR #2059, opened 2024-01-31
Logo fix

VirginiaDooley:
This change implements the footer changes with tags from dc_utils and fixes the right margin on the logo with a new version of design-system.

Dependent on https://github.com/DemocracyClub/dc_django_utils/pull/65 and https://github.com/DemocracyClub/design-system/pull/115


VirginiaDooley:
Fixed in https://github.com/DemocracyClub/EveryElection/pull/2094

PR #2062, opened 2024-02-12
Add exterior ring flag to create_combined_authority command

GeoWill:


symroe (approved):
I don't really understand the consequences of this, but I'm sure it's MostlyFine™.

I guess I'm worried about overlapping organisations. We might want to check this at some point (not needed now)

PR #2086, opened 2024-03-15
Hotfix/cut down org elections endpoint queries

GeoWill:
How I've tested this:
* Checkout master
  * `time curl http://127.0.0.1:8000/api/organisations/parl/parl-hoc/1832-06-07/elections/?format=json | jq .[] > parl-before.json` (real	0m7.770s)
* checkout branch
  * `time curl http://127.0.0.1:8000/api/organisations/parl/parl-hoc/1832-06-07/elections/?format=json | jq .[] > parl-after.json` (real	0m0.962s)
* `git diff --no-index parl-before.json parl-after.json` (This removes all the elections with `"deleted": true`)

This is a change to the API, but it brings it into line with the `/api/elections/` endpoint

Before there were thousands of requests. Now there are 4. Using curl locally time has gone from around 7 seconds to under 1 second. This will be higher on prod, but hopefully enough of an improvement to stop it crashing. 

Curling dev deploy takes just under 8 seconds

fixes https://democracy-club-gp.sentry.io/issues/5054562655/?project=1281142&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=4

### DIFF
--- a/every_election/apps/api/serializers.py
+++ b/every_election/apps/api/serializers.py
@@ -197,10 +197,10 @@ class BaseElectionSerializer(serializers.ModelSerializer):
     def get_deleted(self, obj: Election):
         return obj.current_status == ModerationStatuses.deleted.value
 
-    def get_current(self, obj):
+    def get_current(self, obj: Election):
         return obj.get_current
 
-    def get_voting_system(self, obj):
+    def get_voting_system(self, obj: Election):
         if (
             obj.group_type == "organisation"
             or obj.group_type == "subtype"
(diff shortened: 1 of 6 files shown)


GeoWill:
Failing because of safety, but tests pass locally.

symroe (approved):
I think given the endpoint this is a really good improvement. 

Let's leave it for now, but I'd not be against just returning  a list of election IDs here rather than the full object. I don't think there's loads of value in getting _everything_ in one go like this!

PR #2091, opened 2024-03-20
Ensure that elections without divisions get ID requirements set

symroe:


every_election/apps/elections/sync_helper.py
```diff
@@ -118,6 +118,8 @@ def add_single_election(self, result: dict):
 
                 continue
             if key == "identifier_type":
+                if value == "ballot":
+                    value = None
```
GeoWill, line 122: I think this is already in master, but rebase and merge should strip it out

PR #2094, opened 2024-03-22
Hotfix/dc utils footer

VirginiaDooley:
This change implements the footer changes with tags from dc_utils and fixes the right margin on the logo with a new version of design-system.

Although ultimately no changes were needed, this issue caused some confusion so flagging it here for others pyca/pyopenssl#1143 

Dependent on https://github.com/DemocracyClub/dc_django_utils/pull/65 and https://github.com/DemocracyClub/design-system/pull/115

Screenshot 2024-03-22 at 1 20 31 PM
Screenshot 2024-03-22 at 2 19 40 PM

### DIFF
--- a/every_election/templates/base.html
+++ b/every_election/templates/base.html
@@ -61,44 +61,29 @@
         {% endblock content %}
     </main>
 {% endblock main_base %}
-
-{% block footer_base %}
-    <footer class="ds-footer">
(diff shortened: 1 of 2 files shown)


VirginiaDooley:
There is another dc_utils version on the way but this mostly involves translations so I am going to merge this, esp since the site title in the footer is currently broken on prod.

PR #2096, opened 2024-03-27
Compare date string with date string when checking for end dates

symroe:


every_election/apps/elections/sync_helper.py
```diff
@@ -102,7 +102,11 @@ def add_single_election(self, result: dict):
...
 
-                if divisionset.end_date != value["divisionset"]["end_date"]:
+                if (
+                    divisionset.end_date
+                    and divisionset.end_date.isoformat()
+                    != value["divisionset"]["end_date"]
+                ):
                     # Special case where the end date has changed (likely due to an upcoming boundary change)
                     # in this case, update the division set's end date
                     election_model.division.divisionset.end_date = value[
```
GeoWill, line 112: I'm unclear why this fix works! This line still references `election_model.division.divisionset` - if we call ` election_model = Election(election_id=result["election_id"])` on line 49, then I think `election_model.division` will be `None` at this point. 

Is there a reason why the updated if statement means we won't hit this line if `election_model.division` is `None`?

symroe: You're right there might be another bug where division is None, but I don't think that's the bug we had last night!

GeoWill: The [sentry error](https://democracy-club-gp.sentry.io/issues/5106374500/events/8642ee39c7c34e84af8bfc0f06ee66be/?project=4504713373876224) I'm looking at says:
`AttributeError object has no attribute 'divisionset'`
So I thought that was the error. But as we're discussing in slack, I'm finding Sentry pretty confusing! 
Perhaps there's more than one - though I'm not seeing it on Sentry.

GeoWill:
Potentially superseded by: https://github.com/DemocracyClub/EveryElection/pull/2097

PR #2097, opened 2024-03-28
Fix election sync date compare

GeoWill:
Alternative to #2096

Didn't want to blitz your changes @symroe. 

Hopefully commits and tests make the two changes clear.

### DIFF
--- a/every_election/apps/elections/sync_helper.py
+++ b/every_election/apps/elections/sync_helper.py
@@ -102,6 +102,12 @@ class ElectionSyncer:
                     divisionset.start_date = value["divisionset"]["start_date"]
                     divisionset.save()
 
+                election_model.division = OrganisationDivision.objects.get(
+                    official_identifier=value["official_identifier"],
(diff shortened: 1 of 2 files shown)


every_election/apps/elections/sync_helper.py
```diff
@@ -102,6 +102,12 @@ def add_single_election(self, result: dict):
                     divisionset.start_date = value["divisionset"]["start_date"]
                     divisionset.save()
 
+                election_model.division = OrganisationDivision.objects.get(
+                    official_identifier=value["official_identifier"],
+                    divisionset__start_date=value["divisionset"]["start_date"],
+                    divisionset__organisation=election_model.organisation,
+                )
```
GeoWill, line 109: I had a think about whether we want to error in the case that the election object already exists, but the division defined in the incoming ballot differs from the division that the existing election object has an fk to. 

I decided that we probably _do_ want to change the division if that's the case, so this code is fine as is. 

`election_model.save` gets called on line 201, so don't need that here either.

PR #2108, opened 2024-04-18
Wkt ballot exporter

GeoWill:
I hacked it into dev account to test, and it worked. It should also deploy to dev assuming CI goes through.

### DIFF
--- a/cdk_stacks/stacks/command_runner.py
+++ b/cdk_stacks/stacks/command_runner.py
@@ -84,6 +84,13 @@ class EEOncePerTagCommandRunner(Stack):
                 "output-on-error ee-manage-py-command scrape_lgbce",
             )
 
+            # Export WKT Ballots
+            self.add_job(
+                "export_wkt_ballots",
+                "cron(30 2 * * ? *)",
+                f"output-on-error ee-manage-py-command export_ballots_as_wkt_csv --bucket 'ee.data-cache.{dc_environment}' --prefix 'ballots-with-wkt'",
+            )
+
     def add_job(
         self,
         command_name,
--- a/every_election/apps/elections/management/commands/export_ballots_as_wkt_csv.py
+++ b/every_election/apps/elections/management/commands/export_ballots_as_wkt_csv.py
@@ -1,0 +1,100 @@
+import csv
+import datetime
+import io
+
+import boto3
+import settings
+from django.core.management.base import BaseCommand
+from django.db import connection
+
+
+def export_sql(date: str):
+    return f"""
+    SELECT
+        ee.election_id,
+        COALESCE(odd.id, ogd.id) AS geography_id,
+        COALESCE(st_astext(odd.geography), st_astext(ogd.geography)) AS geography_text,
+        CASE
+           WHEN ogd.id IS NOT NULL THEN 'Organisation'
+           WHEN odd.id IS NOT NULL THEN 'Division'
+           ELSE 'None'
+        END AS source_table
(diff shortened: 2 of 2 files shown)


cdk_stacks/stacks/command_runner.py
```diff
@@ -84,6 +84,13 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
                 "output-on-error ee-manage-py-command scrape_lgbce",
             )
 
+            # Export WKT Ballots
+            self.add_job(
+                "export_wkt_ballots",
+                "cron(30 2 * * ? *)",
+                f"output-on-error ee-manage-py-command export_ballots_as_wkt_csv --bucket 'ee.data-cache.{dc_environment}' --key 'ballots-with-wkt/current-elections.csv'",
```
GeoWill, line 91: This just uses the default value for the `from-date`, which defaults to `now().date()` Is this what we want? or do we want the default to be last week?

symroe: I reckon this should be the same as: https://github.com/DemocracyClub/EveryElection/blob/6851c26c99fb0e96b2420cc6f0768eb790e56e92/every_election/settings/base.py#L284

every_election/apps/elections/management/commands/export_ballots_as_wkt_csv.py
```diff
@@ -0,0 +1,87 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--bucket",
+            help="s3 bucket to export to",
+            action="store",
+            required=True,
+        )
+        parser.add_argument(
+            "--key",
+            help="s3 key (without bucket) to export to",
```
symroe, line 47: (sort of applies to the cron, but there's a line comment there already and I didn't want to confuse things)

If we have `ballots-with-wkt/current-elections.csv` then the prefix `ballots-with-wkt` _cannot_ be used for anything else, or it will break the Glue table in ways that will be annoying and hard to debug.

I think we should enforce a _prefix_ here, and make the actual file name always the same. This should mean that we can't make more than one file per 'folder' by mistake

GeoWill: Do you mean something like `ballots-with-wkt/<date>/current-elections.csv`?

symroe: (and then the key is `f"{prefix}/current_elections.csv"`)

GeoWill: Gotcha

cdk_stacks/stacks/command_runner.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
             self.add_job(
                 "export_wkt_ballots",
                 "cron(30 2 * * ? *)",
-                f"output-on-error ee-manage-py-command export_ballots_as_wkt_csv --bucket 'ee.data-cache.{dc_environment}' --key 'ballots-with-wkt/current-elections.csv'",
+                f"output-on-error ee-manage-py-command export_ballots_as_wkt_csv --bucket 'ee.data-cache.{dc_environment}' --key 'ballots-with-wkt'",
```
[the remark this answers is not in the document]
GeoWill: Good spot!

symroe (approved):

PR #2149, opened 2024-05-27
Save the instance before adding to Elections

VirginiaDooley:
Ref https://democracy-club-gp.sentry.io/share/issue/b06dd1f5d4a14e25b22c2d65aece04ab/ 

---
  - https://app.asana.com/0/0/1207407429226442


symroe (approved):

symroe (changes requested):
Sorry, still a `pdb` in here

PR #2198, opened 2024-07-25
postgres 16/python 3.12/node 18

chris48s:
Right. I've tried to keep the scope for this PR in check, but it is still a bit of a chunky monkey.

Things I am doing in this PR:

- Run on/test against postgres 16/python 3.12/node 18 in local dev and CI
- Make code/packages changes necessary to run on/test against postgres 16/python 3.12/node 18 (keeping the scope as small as possible)
- Run lambda "cron" jobs on python 3.12
- Replace black with ruff format (there's a bit of scope creep around this, but I've tried to keep it in line)

There is a decent amount of info in the individual commit messages about what I am doing/why at each stage.

Next jobs:

- I haven't really touched any of the CDK stuff yet. As I understand it, the next job is to build a new base image using Ubunutu 24.04, pull that in and base off that. Somehow the CI build has magically deployed to dev without me doing that though. Might need to have a chat about this to get my head round it.
- Postgres upgrades: This is a ClickOps job. Given we don't actually need to upgrade python or django in order to migrate to the latest Posgres, I wonder if it makes sense to just migrate all the RDS instances to Postgres 16 (starting with dev, then staging, then prod) before we merge this. Just to reduce the amount of stuff we are changing at once.

Reckon this is probably still worth a round of review at this point anyway.


pyproject.toml
```diff
@@ -16,20 +16,23 @@ extend-select = [
...
     "--ignore=*/__pycache__",
     "--ignore=.*",
     "--ignore=.download-cache",
     "--ignore=every_election/settings/local.py",
     "--ignore=lib",
     "--ruff",
-    "--black",
-]
\ No newline at end of file
+    "--ruff-format",
```
symroe, line 37: I think `--ruff` was the same as the newer `--ruff-check`, rather than format? By removing `black` we want to use both check and format here to be a like-for-like replacement

chris48s: We are invoking with both `--ruff` and `--ruff-format`

https://github.com/DemocracyClub/EveryElection/blob/2c974b0df451651b063a8a52c45933d20eb69ce8/pyproject.toml#L36-L37

which I think are the right options for the version of pytest-ruff we're pinned to, but I'll double check before I finalise this one for review

chris48s: I tried introducing some linting and formatting errors, running pytest and making sure both were flagged.

This is the right combination of options for this version of pytest-ruff.
This might change when we upgrade, but I'm trying not to scope-creep this PR too much.

symroe: Right sorry, I read the diff wrong and though you'd replaced `ruff` with `ruff-format` :+1:

requirements/testing.txt
```diff
@@ -1,6 +1,5 @@
 -r base.txt
 
-black==23.7.0
```
chris48s, line 3: Worth noting here that we're still on a version of `dc_django_utils` that doesn't have https://github.com/DemocracyClub/dc_django_utils/pull/84 in it yet, so black (and loads of other stuff) actually still gets installed for now.

package.json
```diff
@@ -22,5 +22,8 @@
     "topojson": "^3.0.2",
     "aws-cdk": "2.50.0",
     "leaflet": "^1.9.2"
+  },
+  "engines": {
+    "node": ">=18"
```
chris48s, line 27: I've added an `engines` declaration here.
What I've written here will mean we log a warning if you try to `npm ci` on a version of node below 18, but it will try anywhere.

There's a few things we can do to be more strict here.
One is we could make this `"node": "^18"` so it will also warn if you're trying to `npm ci` on a major version _above_ 18 here. We can also specify a specific `npm` version here too.

The other thing we can do is create a `.npmrc` with `engine-strict=true`. That will mean we refuse to install if the node version violates the version in `engines` rather than just logging a warning.

Normally for my node applications (as opposed to libraries), I'd pin this to a specific node major version and create a `.npmrc` with `engine-strict=true` in it. Given the only thing we're using node for is to invoke `npm`, maybe we want to be less strict here.

Thoughts?

symroe: I don't know enough about that ecosystem to have an informed opinion here, I'm afraid. Do you know if there are significant changes in the way `npm` works between versions? Like, are lock files generated in different ways, or is dependency resolution different in ways that will break things?

chris48s: > Do you know if there are significant changes in the way `npm` works between versions? Like, are lock files generated in different ways

The answer to this is mostly no, but sometimes yes.
In general, npm is quite stable.
However, there was a period when NPM transitioned from the lockfile version 1 format to lockfile version 2 format where it was _super annoying_ if you had 2 devs on the same team using 2 different NPM versions because if person A was the last person to touch the lockfile and person B came along and installed something then it would rewrite the entire lockfile format.
Coincidentally, it was around this time that I started setting a `npm` version in `engines` and using `engine-strict=true` :thinking:
However, for the last few NPM versions it has been quite stable in that respect.

Here is maybe a better question:

If you (or Virginia, or Will) came to this project (or another DC project) and typed `npm ci` and it refused to install anything because you were using a different major node/npm version than the ones we expected, how annoying would that be? Would you simply `nvm use` the correct version and go about your day or would that present a major barrier to getting stuff done?

chris48s: Also.. I feel like this specific point is something we could always punt to another day.

chris48s:
Build failure is legit. New image fails health checks with

```
Health checks failed with these codes: [400]
```

but I can't reproduce.

Currently stumped :deciduous_tree:

chris48s:
OK. The Load Balancer ping is getting a `DisallowedHost` back.. next job: What changed?

chris48s:
The difference is that the new VM is set to "IMDSv2: Required", whereas the previous one was "IMDSv2: Optional".
I suspect this is probably a result of basing the dc_base_image on a newer first-party AMI and we'll have to solve this everywhere. Currently unconfirmed.

Side note: I can see why we're silently passing this exception, but this feels like a bit of a foot-gun tbh :foot::gun:

chris48s:
OK, so I reckon this is finally now sensible to review.

I ended up picking
https://github.com/DemocracyClub/EveryElection/commit/4886869841d48e0426065bc8a106c8c561949cf4
and
https://github.com/DemocracyClub/EveryElection/commit/d49959b2270777ca25302836fa8e2fce52e3528b
forward and just pushing those to master so that I could build `ami-0c8d319e3a044e7da`
..which I believe completes this PR 🤞

every_election/settings/base.py
chris48s, line 1: Just to summarise what I have done here and why..

The difference is that the new VM is set to "IMDSv2: Required", whereas the previous one was "IMDSv2: Optional". I suspect this is probably a result of basing the dc_base_image on a newer first-party AMI and we'll have to solve this everywhere.

The upshot of this is that http://169.254.169.254/latest/meta-data/local-ipv4 now requires auth and returns a `400 Bad Request` if you don't pass a token.

One of the things I have done here is updated the code to get a token and use it.

Another thing I have done is updated the code to `pass` on `requests.exceptions.Timeout` or `requests.exceptions.ConnectionError` but throw other errors. One of the reasons I spent ages being confused about this was because we just passed over the `400 Bad Request`. If it had just failed loudly while bootstrapping the app I might have had an easier time debugging it.

This is a bit less bad than what we were doing before. Tbh, I actually think we should put stuff like this in a dedicated `production.py` but I think I'm not doing to do that right now in this PR.

chris48s: ..or if we actively want to avoid a separate `production.py`, another option would be to control whether we do this or not via an env var

.circleci/config.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           name: Run tests
           command: |
             . .venv/bin/activate
-            pytest --ruff --cov-report= --cov=every_election --junitxml=test-results/junit.xml
```
symroe, line 76: Minor point relating to this that we can punt until later...I like to lint early on before running tests so we can fail faster. I think in other places we're running lint (both `ruff check` and `ruff format` as soon as they're installed.

Not needed in this PR, as you've not changed this behaviour, but it would be nice to agree on a standard way to do this as part of aligning all CI projects.

every_election/settings/base.py
symroe: I think you can improve this with a minor change to check for `DC_ENVIRONMENT` in `get_ec2_ip`. This will only be set in a deployed environment

symroe (approved):
Couple of minor points that I'm ok with you ignoring. Otherwise this is good to go :+1:

.circleci/config.yml
chris48s: Yes I also think running the fastest steps of the build is usually a good idea.

I reckon the sensible time to consider this is when we are thinking about makefiles for common tasks on projects.

every_election/settings/base.py
chris48s: I was wondering if there was an env var we could rely on being set in every case.

I just SSHd in and checked. As far as I can see, `DC_ENVIRONMENT` is not set in dev or staging. That is a variable that is set in the CircleCI builds, but not in parameter store. So I think we can rely on that at build time, but not at runtime.

So these checks:

- https://github.com/DemocracyClub/EveryElection/blob/d49959b2270777ca25302836fa8e2fce52e3528b/cdk_app.py#L18
- https://github.com/DemocracyClub/EveryElection/blob/d49959b2270777ca25302836fa8e2fce52e3528b/deployscripts/create_deployment_group.py#L119

are safe

but I think this stuff is not doing what you think it is:

- https://github.com/DemocracyClub/EveryElection/blob/d49959b2270777ca25302836fa8e2fce52e3528b/every_election/settings/base.py#L293
- https://github.com/DemocracyClub/EveryElection/blob/d49959b2270777ca25302836fa8e2fce52e3528b/every_election/settings/base.py#L317

symroe: Ah this is interesting. Could you add those params please? Either in CDK or manually? In general we should have these set in the parameter store for each account...I'll do an audit

chris48s: https://github.com/DemocracyClub/EveryElection/pull/2198/commits/73d91e34bc6d3f0379fe50c42b29fee69562bc8f

PR #2199, opened 2024-07-25
get the build passing

chris48s:
Over on https://github.com/DemocracyClub/EveryElection/pull/2198 I find myself shaving various yaks unrelated to the original task.

Lets make a new branch and get to a stage where we've got a green CI build first.

Then I'll base the new work on top of that.

### DIFF
--- a/every_election/apps/core/tests/test_urls.py
+++ b/every_election/apps/core/tests/test_urls.py
@@ -6,7 +6,12 @@ from django_extensions.management.commands.show_urls import Command
 
 class UrlTests(TestCase):
     def is_exception(self, url):
-        exceptions = [".txt", ".ics", "/admin/<url>"]
+        exceptions = [
+            ".txt",
+            ".ics",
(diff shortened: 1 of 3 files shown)


chris48s:
The first commit here just proves our starting point is: If we run a build from what is on the top of `master`, it doesn't work.

chris48s:
OK, so in the end the changes needed were:

Safety CI was reporting security issues with:
- requests
- urllib3
- django rest framework
Upgraded all those. Bumping boto3 and botocore was necessary as the versions we were on was pinning us to an old version of urllib3.

Upgrading django rest framework required a change to the trailing slashes test case.

Our code style checks were failing due to breaking changes in ruff/pytest-ruff. For now, I've pinned pytest-ruff to the last known good version. We can upgrade it in another PR.

This is the minimum changeset needed to get a passing build. This is a baseline we can iterate from.

symroe (approved):

PR #2200, opened 2024-07-30
remove per-instance crontab entries

chris48s:
Right. So I did some digging.
All of these commands actually are running on the EE prod server(s).
EE has been restoring itself from a copy of its own backup every night :scream_cat: 

I don't think it was the intent that these jobs should have been running on WDIV servers running their own EE instance. We definitely wouldn't have wanted them running `snoop` for example, and `sync_elections` is there as an _alternative_ to restoring from backup every night.

Also a bit worried that we have no log aggregation and not really any visibility on these cron jobs, but I'm not going to do anything about those problems here.

This is one of those PRs that asks a lot of questions as well as answering some, so I'm going to leave some comments inline.

### DIFF
--- a/cdk_stacks/stacks/command_runner.py
+++ b/cdk_stacks/stacks/command_runner.py
@@ -57,11 +57,11 @@ class EEOncePerTagCommandRunner(Stack):
             )
 
             # New elections
-            # self.add_job(
-            #     "snoop",
-            #     "cron(30 * * * ? *)",
-            #     "output-on-error ee-manage-py-command snoop",
-            # )
+            self.add_job(
+                "snoop",
+                "cron(30 * * * ? *)",
+                "output-on-error ee-manage-py-command snoop",
+            )
 
             # Generate map layers and sync to S3
             self.add_job(
--- a/cdk_stacks/stacks/command_runner.py
+++ b/cdk_stacks/stacks/command_runner.py
@@ -74,7 +74,7 @@ class EEOncePerTagCommandRunner(Stack):
             self.add_job(
                 "add_nuts1_tags",
                 "cron(15 2 * * ? *)",
-                """output-on-error ee-manage-py-command add_tags -u "https://s3.eu-west-2.amazonaws.com/ee.public.data/ons-cache/NUTS_Level_1_(January_2018)_Boundaries.gpkg" --fields '{"NUTS118NM": "value", "NUTS118CD": "key"}' --tag-name NUTS1 --is-gpkg""",
+                "output-on-error /var/www/every_election/repo/serverscripts/add_nuts1.sh",
             )
 
             # Scrape LGBCE
--- a/deployscripts/crontab
+++ b/deployscripts/crontab
@@ -1,7 +1,4 @@
 # WARNING: Commands in this file will run on *every* instance in *every* environment
-# HOWEVER: these commands will NOT run on instances booted from the EE AMI
+# HOWEVER: these commands will NOT run on instances
+# booted from the EE AMI (i.e: WDIV installs running a "local" EE)
 # If you want to have a command run only once per env / once globally, use SSM Run Command
-0 2 * * * every_election cd /var/www/every_election/repo && ./serverscripts/rebuild_local_db.sh
-30 2 * * * every_election cd /var/www/every_election/repo && ./serverscripts/add_nuts1.sh
-0 1 * * * every_election ee-manage-py-command dumpdata elections election_snooper --indent 4 -o /var/www/every_election/repo/every_election/data/elections.json
-0 0 * * * every_election ee-manage-py-command snoop
--- a/appspec.yml
+++ b/appspec.yml
@@ -82,9 +82,6 @@ hooks:
     - location: deployscripts/write_envfile.sh
       timeout: 300
       runas: root
-    - location: serverscripts/rebuild_local_db.sh
-      timeout: 900
-      runas: every_election
     - location: deployscripts/install_crontab.sh
       timeout: 900
       runas: root
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -2,7 +2,7 @@ beautifulsoup4==4.12.1
 boto3==1.34.148
 botocore==1.34.148
 commitment==3.0.1
-Django==4.2.11
+Django==4.2.15
 django-cors-headers==4.2.0
 django-extensions==3.2.3
 django-filter==24.2


deployscripts/crontab
```diff
@@ -1,7 +1,4 @@
 # WARNING: Commands in this file will run on *every* instance in *every* environment
-# HOWEVER: these commands will NOT run on instances booted from the EE AMI
+# HOWEVER: these commands will NOT run on instances
+# booted from the EE AMI (i.e: WDIV installs running a "local" EE)
```
chris48s, line 3: I've attempted to clarify things with this comment. Does this seem better?

cdk_stacks/stacks/command_runner.py
```diff
@@ -91,6 +91,14 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
                 f"output-on-error ee-manage-py-command export_ballots_as_wkt_csv --bucket 'ee.data-cache.{dc_environment}' --prefix 'ballots-with-wkt'",
             )
 
+        if dc_environment == "staging":
+            # Restore staging from prod backup
+            self.add_job(
+                "restore_staging_from_s3",
+                "cron(0 2 * * ? *)",
```
chris48s, line 98: If this was in a crontab, I would write

```
0 2 * * *
```

The syntax used elsewhere for these jobs seems to imply this should be

```
0 2 * * ? *
```

Is there any documentation I can read to understand this?

cdk_stacks/stacks/command_runner.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
             self.add_job(
                 "add_nuts1_tags",
                 "cron(15 2 * * ? *)",
-                """output-on-error ee-manage-py-command add_tags -u "https://s3.eu-west-2.amazonaws.com/ee.public.data/ons-cache/NUTS_Level_1_(January_2018)_Boundaries.gpkg" --fields '{"NUTS118NM": "value", "NUTS118CD": "key"}' --tag-name NUTS1 --is-gpkg""",
+                "output-on-error /var/www/every_election/repo/serverscripts/add_nuts1.sh",
```
chris48s, line 77: This is basically runing the same command that is in this shell script.
Might as well use the shell script?

deployscripts/crontab
```diff
@@ -1,7 +1,4 @@
...
 # If you want to have a command run only once per env / once globally, use SSM Run Command
-0 2 * * * every_election cd /var/www/every_election/repo && ./serverscripts/rebuild_local_db.sh
-30 2 * * * every_election cd /var/www/every_election/repo && ./serverscripts/add_nuts1.sh
-0 1 * * * every_election ee-manage-py-command dumpdata elections election_snooper --indent 4 -o /var/www/every_election/repo/every_election/data/elections.json
```
chris48s, line 6: This `dumpdata` command just doesn't need to run any more. Since https://github.com/DemocracyClub/EveryElection/pull/1991 it is not relevant. I will remove some other things relating to this, but that is another PR. Here I am just removing this command.

cdk_stacks/stacks/command_runner.py
```diff
@@ -91,6 +91,14 @@ def __init__(self, scope: Construct, construct_id: str, **kwargs) -> None:
                 f"output-on-error ee-manage-py-command export_ballots_as_wkt_csv --bucket 'ee.data-cache.{dc_environment}' --prefix 'ballots-with-wkt'",
             )
 
+        if dc_environment == "staging":
```
chris48s, line 94: This looks right, but I've not actually tested any changes in this file. Is there any meaningful testing we can do on this other than deploy + monitor it?

chris48s:
The other big worry I have here is

https://github.com/DemocracyClub/EveryElection/blob/9caa6c31767d3aad6299c0e81a185d3596da1273/appspec.yml#L85-L87

When are these command run? Are we runing this on every deploy, including to production??
If it helps, the original commit doing all this was https://github.com/DemocracyClub/EveryElection/commit/e48dd917e45517800bb3685d2b3654a1eff88990

deployscripts/crontab
chris48s: PR open at https://github.com/DemocracyClub/EveryElection/pull/2201

cdk_stacks/stacks/command_runner.py
symroe: [Yes](https://docs.aws.amazon.com/systems-manager/latest/userguide/reference-cron-and-rate-expressions.html)

I'm sure there's some logic to this, but AWS deciding to use a _slightly_ different scheduling syntax than UNIX CRON is baffling, and has lead to problems in the past (e.g you can't validate using crontab.guru)

deployscripts/crontab
symroe: This isn't quite right: the sync elections management command will only sync...elections. That is, if we change a division set, add organisations, or do more or less anything else with the data model it's not going to update. 

We still restore from a DB back up nightly on the embedded instances. `sync_elections` is really for quick syncing of new elections, and e.g cancellations.

Also, yes, there is a problem if the data model in EE prod changes before the onboard EEs have had the code updated. That's such a rare case though, and one that doesn't happen just before elections, so based on our current traffic assumptions it's a risk we're ok with

cdk_stacks/stacks/command_runner.py
symroe: The string value is enforced [here](https://github.com/DemocracyClub/EveryElection/blob/master/cdk_app.py#L28). I think it's reasonable to trust that the string value is correct, but I guess we could pull `valid_environments` out and share the values somehow.

symroe:
> The other big worry I have here is
> 
> https://github.com/DemocracyClub/EveryElection/blob/9caa6c31767d3aad6299c0e81a185d3596da1273/appspec.yml#L85-L87
> 
> When are these command run? Are we runing this on every deploy, including to production?? If it helps, the original commit doing all this was [e48dd91](https://github.com/DemocracyClub/EveryElection/commit/e48dd917e45517800bb3685d2b3654a1eff88990)

Thinking about this more, I expect this was all caused by us copying the WDIV CDK model. Specifically, in WDIV we use replication, and here we don't. 

When using replication, you want to do something with the local DB before the instance serves traffic. When using a single remote postgres, you don't. I think I must have been confused between the two models.

deployscripts/crontab
chris48s: OK. Thanks for the clarification.
Given none of the commands in here run on embedded instances, the upshot for this PR is no change though, right?

cdk_stacks/stacks/command_runner.py
chris48s: I was thinking about the changes to this file slightly more widely tbh. Just had to put the comment _somewhere_

chris48s:
>  I must have been confused between the two models.

OK.. so upshot for this PR is: remove that line from `appspec.yml` then?

deployscripts/crontab
symroe: Sorry yes that's correct

cdk_stacks/stacks/command_runner.py
symroe: Right. Yeah, not much else we can do apart from deploy and monitor I don't think. I also don't think we get notified of failed `run_command` jobs...I saw some when you looked the other day that had failed, for example

chris48s:
Right.

Just for future reference: This was kind of less bad than we first thought.
(For reasons) as well as having a shared RDS instance the EE VMs also have a local postgres install. The `rebuild_local_db` script was running against the local Postgres install, not the shared RDS.

---

So I removed that line from `appspec.yml`. My next thought was that this also makes

https://github.com/DemocracyClub/EveryElection/blob/38ad35a793f5600511c1c94630e7570e202f0ece/cdk_stacks/stacks/command_runner.py#L94-L99

wrong because we really want staging to run that against the staging RDS, not the local DB.

Then I realised there is no staging RDS (or at least no staging DB connection credentials). Probably no dev RDS either.
So maybe this is actually sort of _right_
I think the next thing is: Do we think it is right that dev/staging are using a local DB not RDS?

chris48s:
I reckon removing the command from `appspec.yml` is why the deploy has completely hung as well.

Its probably getting a 500 from the healthcheck endpoint or whatever because the (local) DB is empty

symroe (approved):

chris48s:
I'm going to deploy this one first to make sure the staging RDS setup is fine. Then I will do https://github.com/DemocracyClub/EveryElection/pull/2198 second once I've checked that

PR #2202, opened 2024-07-30
use npm ci instead of npm install

chris48s:
This project uses a lockfile so should use [npm ci](https://docs.npmjs.com/cli/v10/commands/npm-ci) for deployment

### DIFF
--- a/cdk_stacks/components/install_app.yml
+++ b/cdk_stacks/components/install_app.yml
@@ -49,7 +49,7 @@ phases:
             - pip install -U pip
             - pip install -r requirements/production.txt
             - pip install gevent
-            - npm install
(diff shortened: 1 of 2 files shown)


symroe (approved):

symroe:
TIL. Can you document this somewhere, like the dev handbook? This is something that would be useful to use everywhere we use NPM

chris48s:
If the thing you want to do is install what's in your lockfile, this is what we should do everywhere - yes.

Where's the dev handbook? :see_no_evil:

chris48s:
There wasn't really a sensible place to fit this into, but I started a section called "Node and NPM" and put it there.

PR #2203, opened 2024-07-30
update install instructions

chris48s:
Install instructions were missing a couple of steps:

1. We need to install JS packages
2. On this project, it seems to be necessary to manually run `collectstatic` every time we change frontend assets.

In general, I'd like to a) have this stuff more scripted b) get manage.py monitoring for changes to frontend assets

but for now lets at least make the docs right

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -19,12 +19,6 @@ export LIBRARY_PATH=$LIBRARY_PATH:/opt/homebrew/opt/openssl/lib
 npm ci
 ```
 
-## Collectstatic
-
-```
-python manage.py collectstatic
-```
-
 ## Set up database
 By default, EveryElection uses PostgreSQL with the PostGIS extension. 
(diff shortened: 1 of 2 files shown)


symroe:
Out of interest, are you running the project with `DEBUG=True`?

symroe (approved):

chris48s:
Turns out.. I am not.
`DEBUG=True` seems like something that should be in https://github.com/DemocracyClub/EveryElection/blob/master/every_election/settings/local.example.py right?

Having done a quick test, that still doesn't seem to re-compile static files on edit, but I'm not sure if it alleviates the need for an initial `collectstatic`. Let me circle back and check before we merge this one..

chris48s:
Right. Having `DEBUG = True` on does a collectstatic for you on init. I don't think it hot reloads on changes to static files, but this is more expected.

PR #2210, opened 2024-08-15
add additional CI checks

chris48s:


.circleci/config.yml
```diff
@@ -70,6 +70,10 @@ jobs:
             python manage.py --version
             python manage.py check
             python manage.py makemigrations --check
+            pip check
+            ruff format . --check
+            ruff check .
+            djhtml every_election/ --check
```
chris48s, line 76: Note `djhtml every_election/` not `djhtml .`
The reason I'm doing this is to stop djhtml from trying to format files inside `.venv`
AFAIK djhtml doesn't have a config file we can put this in.

Arguably, we shouldn't run this on the html files in `every_election/apps/organisations/boundaries/boundary_bot/tests/fixtures` either but I think its probably best to just make this as simple as possible.

.circleci/config.yml
```diff
@@ -70,6 +70,10 @@ jobs:
             python manage.py --version
             python manage.py check
             python manage.py makemigrations --check
+            pip check
+            ruff format . --check
+            ruff check .
+            git ls-files '*.html' | xargs djhtml --check
```
chris48s, line 76: Note the invocation `git ls-files '*.html' | xargs djhtml --check` here rather than `djhtml .`

This passes only html files that are tracked by git. The reason I'm doing this is to stop `djhtml` from trying to format or check files inside `.venv`, `static`, `node_modules`, etc.

djhtml doesn't have a config file we can put this in. It also doesn't respect `.gitignore` files.

This is a sufficiently fiddly command it would be nice to abstract it behind `make format`, or whatever.

Arguably, we shouldn't run this on the html files in `every_election/apps/organisations/boundaries/boundary_bot/tests/fixtures` either but I think its probably best to just make invoking this as simple as possible.

symroe: Are you keeping a list of "things we might want in a common makefile" somewhere?

symroe (approved):

.circleci/config.yml
chris48s: Only in my head :see_no_evil:

PR #2225, opened 2024-09-04
Fix LGBCE scraper

VirginiaDooley:
co-authoured with @awdem 

This PR makes some changes to how the LGCE scraper works and should get it to run again. 

Before, the scraper first tried to `legislation_title` and `slug`  to determine if a review existed in our database. 

This PR changes it to use the `legislation_year`  in place of `legislation_title`. We extract `legislation_year` from the `legislation_url` of each scraped record, when there is a url.

TESTING:

For specific pytest changes/additions:
```
pytest every_election/apps/organisations/boundaries/boundary_bot/tests/test_scraper.py
pytest every_election/apps/organisations/boundaries/boundary_bot/tests/test_validation.py
```

To run the scraper, make sure you've restored your local database to a recent backup then run:
```
# The following commands fix some issues in the database that were causing the scraper to fail 
python manage.py remove_duplicate_reviews
python manage.py update_reviews_with_bad_links
# Run the scraper:
python manage.py scrape_lgbce
```
The scraper should run to completion without raising any errors and end with a list of slack messages and github PRs to create (it won't actually create them).

### DIFF
--- a/every_election/apps/organisations/boundaries/boundary_bot/scraper.py
+++ b/every_election/apps/organisations/boundaries/boundary_bot/scraper.py
@@ -70,6 +70,7 @@ class LgbceScraper:
             "greater-london",
             "north-yorkshire",
             "somerset",
+            "north-warwickshire",  # LGBCE site links to webarchive version of legislation rather than legislation.gov.uk
         ]
 
     def scrape_index(self):
--- a/every_election/apps/organisations/boundaries/boundary_bot/scraper.py
+++ b/every_election/apps/organisations/boundaries/boundary_bot/scraper.py
@@ -1,4 +1,5 @@
 import pprint
+import re
 
 import lxml.html
 import requests
@@ -181,6 +182,14 @@ class LgbceScraper:
         return OrganisationBoundaryReview.objects.filter(
             organisation=org, slug=record["slug"]
         ).exclude(status=ReviewStatus.COMPLETED)
+
+    def clean_legislation_url(self, url):
+        url = url.replace("/id/", "/")
+        url = re.search(
+            r"(?:www\.)?legislation.gov.uk/(wsi|ukdsi|uksi|ssi)/\d+/\d+",
+            url,
+        ).group()
+        return f"https://{url}"
 
     def validate(self):
         # perform some consistency checks
--- a/every_election/apps/organisations/boundaries/boundary_bot/scraper.py
+++ b/every_election/apps/organisations/boundaries/boundary_bot/scraper.py
@@ -163,17 +163,21 @@ class LgbceScraper:
         return org
 
     def get_review_from_db(self, record):
-        if record["legislation_title"]:
+        if record["legislation_url"]:
+            legislation_year = self.get_legislation_year(
+                record["legislation_url"]
+            )
+
             try:
                 result = OrganisationBoundaryReview.objects.filter(
-                    legislation_title=record["legislation_title"],
+                    legislation_url__contains=f"/{legislation_year}/",
                     slug=record["slug"],
                 )
                 if len(result) == 1:
                     return result
                 if len(result) > 1:
                     raise ScraperException(
-                        f"More than one review found with same legislation_title: {record['legislation_title']}"
+                        f"More than one review found for {record['slug']} with year {legislation_year}",
                     )
             except OrganisationBoundaryReview.DoesNotExist:
                 pass
--- a/every_election/apps/organisations/boundaries/boundary_bot/scraper.py
+++ b/every_election/apps/organisations/boundaries/boundary_bot/scraper.py
@@ -1,5 +1,6 @@
 import pprint
 import re
+from urllib.parse import urlparse
 
 import lxml.html
 import requests
@@ -190,6 +191,16 @@ class LgbceScraper:
             url,
         ).group()
         return f"https://{url}"
+
+    def get_legislation_year(self, url):
+        try:
+            cleaned_url = self.clean_legislation_url(url)
+        except AttributeError:
+            raise ScraperException(f"Failed to clean legislation URL: {url}")
+
+        urlparse_result = urlparse(cleaned_url)
+        path = urlparse_result.path
+        return path.split("/")[2]
 
     def validate(self):
         # perform some consistency checks
--- a/every_election/apps/organisations/boundaries/boundary_bot/tests/test_scraper.py
+++ b/every_election/apps/organisations/boundaries/boundary_bot/tests/test_scraper.py
@@ -227,3 +227,57 @@ class TestScraperSaves(TestCase):
                 slug="allerdale"
             ).legislation_title,
         )
+
+
+class TestGetLegislationYear(TestCase):
+    def setUp(self):
+        self.scraper = LgbceScraper(False, False)
+
+    def test_get_legislation_year_uksi_url(self):
+        legislation_url = "https://www.legislation.gov.uk/uksi/2023/1023/made"
+        year = self.scraper.get_legislation_year(legislation_url)
+        self.assertEqual(year, "2023")
+
+    def test_get_legislation_year_uksi_url_with_id(self):
+        legislation_url = "http://www.legislation.gov.uk/id/uksi/2023/732"
+        year = self.scraper.get_legislation_year(legislation_url)
+        self.assertEqual(year, "2023")
+
+    def test_get_legislation_year_no_subdomain(self):
+        legislation_url = "legislation.gov.uk/id/uksi/2023/732"
+        year = self.scraper.get_legislation_year(legislation_url)
+        self.assertEqual(year, "2023")
+
+    def test_get_legislation_year_ukdsi_url(self):
+        legislation_url = (
+            "https://www.legislation.gov.uk/ukdsi/2024/9780348262735/contents"
+        )
+        year = self.scraper.get_legislation_year(legislation_url)
+        self.assertEqual(year, "2024")
+
+    def test_get_legislation_year_wsi_url(self):
+        legislation_url = (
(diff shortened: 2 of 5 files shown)


every_election/apps/organisations/boundaries/boundary_bot/tests/test_validation.py
```diff
@@ -79,6 +79,9 @@ def test_backwards_status_move(self):
             "legislation_title"
         ] = "The Allerdale Electoral Change order"
         scraper.data["allerdale"]["latest_event"] = "Effective date"
+        scraper.data["allerdale"][
+            "legislation_url"
+        ] = "https://www.legislation.gov.uk/uksi/2017/1067/contents/made"
```
awdem, line 84: This test was failing because the 'scraped' record was being mocked with no legislation link despite having status `COMPLETED`.  When the record's status changes back to `CURRENT`, the scraper would see the record as new instead of updating the completed one. 

This is because in `get_review_from_db`, if we don't have a `legislation_url`, then we search by `org` and `slug` and exclude `COMPLETED` reviews, so we don't find the 'completed' review that has had its status backslide even though it was in our database.

It's a weird edge case - I assume a completed boundary review should always have a legislation link so we would catch a backwards move like this?

every_election/apps/organisations/boundaries/boundary_bot/scraper.py
```diff
@@ -182,6 +187,26 @@ def get_review_from_db(self, record):
             organisation=org, slug=record["slug"]
         ).exclude(status=ReviewStatus.COMPLETED)
 
+    def clean_legislation_url(self, url):
+        url = url.replace("/id/", "/")
+        url = re.search(
+            r"www.legislation.gov.uk/(wsi|ukdsi|uksi|ssi)/\d+/\d+",
```
awdem, line 193: I don't know if this method actually needs to handle urls with `wsi` or `ssi`, because it's only scraping English legislation? I kept them in because I wasn't sure.

every_election/apps/organisations/boundaries/boundary_bot/scraper.py
```diff
@@ -161,17 +164,19 @@ def get_org_from_reg_code(self, register_code):
...
-                    legislation_title=record["legislation_title"],
+                    legislation_url__contains=f"/{legislation_year}/",
                     slug=record["slug"],
                 )
                 if len(result) == 1:
                     return result
                 if len(result) > 1:
                     raise ScraperException(
-                        f"More than one review found with same legislation_title: {record['legislation_title']}"
+                        f"More than one review found for {record['legislation_title']} with same year: \n {str(record)}",
```
chris48s, line 179: The duplicate here is on year and slug, so lets say that in the error.

```suggestion
                        f"More than one review found for {record['slug']} with year {legislation_year}",
```

I don't think we need to dump the whole record

every_election/apps/organisations/boundaries/boundary_bot/scraper.py
chris48s: I don't think this does any harm :+1:

every_election/apps/organisations/boundaries/boundary_bot/scraper.py
```diff
@@ -182,6 +187,26 @@ def get_review_from_db(self, record):
...
 
+    def clean_legislation_url(self, url):
+        url = url.replace("/id/", "/")
+        url = re.search(
+            r"www.legislation.gov.uk/(wsi|ukdsi|uksi|ssi)/\d+/\d+",
+            url,
+        ).group()
+        return f"https://{url}"
+
+    def get_legislation_year(self, record):
```
chris48s, line 198: This method is only really operating on a URL. Instead of passing it the whole record dic, can we make this method take just a url as the input param.
The one place you're using the whole `record` is in the errror message, but we don't really need to dump the whole dict there. Just `f"Failed to clean legislation URL: {url}"` would do the job there.
You'll also need to update call site and tests.

every_election/apps/organisations/boundaries/boundary_bot/scraper.py
```diff
@@ -182,6 +187,26 @@ def get_review_from_db(self, record):
...
+        try:
+            cleaned_url = self.clean_legislation_url(record["legislation_url"])
+        except AttributeError:
+            raise ScraperException(
+                "Failed to clean legislation URL: \n%s" % str(record)
+            )
```
chris48s, line 204: any particular reason for using `%` rather than a f-string here?

every_election/apps/organisations/boundaries/boundary_bot/scraper.py
chris48s, line 193: Can we make this tolerate links both with/without `www.` here (and add a test case for without www)

every_election/apps/organisations/boundaries/management/commands/update_reviews_with_bad_links.py
```diff
@@ -0,0 +1,28 @@
...
+        review = OrganisationBoundaryReview.objects.get(id=review_id)
+        old_link = review.legislation_url
+        review.legislation_url = correct_legislation_url
+        review.save()
+        print(
+            f"Updated legislation link for {review.generic_title} from",
+            old_link,
+            "to",
+            review.legislation_url,
+        )
```
chris48s, line 28: Can we use `self.stdout.write()` instead of `print()` here.

every_election/apps/organisations/boundaries/boundary_bot/scraper.py
```diff
@@ -68,6 +70,7 @@ def __init__(self, BOOTSTRAP_MODE, SEND_NOTIFICATIONS):
             "greater-london",
             "north-yorkshire",
             "somerset",
```
chris48s, line 72: Did we get anywhere with these other ignores, or are we leaving them to another PR for day? Its fine if we are.. just wondering

every_election/apps/organisations/boundaries/management/commands/remove_duplicate_reviews.py
```diff
@@ -0,0 +1,39 @@
+from django.core.management import BaseCommand
+from organisations.models import OrganisationBoundaryReview
+
+
+class Command(BaseCommand):
+    help = "removes a specific set of duplicated boundary reviews from database"
+
+    def handle(self, *args, **options):
+        ids = [
```
chris48s, line 9: Can you add a comment here explaining where this list of magic numbers came from.
e.g: explain we had a load of duplicates and these are all the duplicate IDs that weren't attached to a divisionset, or whatever the rationale is.

every_election/apps/organisations/boundaries/management/commands/remove_duplicate_reviews.py
```diff
@@ -0,0 +1,39 @@
+from django.core.management import BaseCommand
+from organisations.models import OrganisationBoundaryReview
+
+
+class Command(BaseCommand):
+    help = "removes a specific set of duplicated boundary reviews from database"
```
chris48s, line 6: With both of these management commands, it is good that we've done them as commands so we can work on them locally, test in a reproducible way and then run the same process in prod. They're not really commands we want to run lots of times though. In the help, can we make it clear these are one shot operations rather than re-usable commands we're going to want again in future. e.g:

```
help = "One-off command to remove duplicated boundary reviews from database"
```

```
help = "One-off command to update boundary reviews with bad legislation links"
```

every_election/apps/organisations/boundaries/boundary_bot/tests/test_scraper.py
```diff
@@ -48,43 +49,47 @@ def test_get_review_from_db(self):
...
+        complete_qs = self.scraper.get_review_from_db(
             {
                 "register_code": org1.official_identifier,
                 "legislation_title": "completed title",
+                "legislation_url": "https://www.legislation.gov.uk/uksi/2023/1023/made",
                 "slug": org1.slug,
             }
         )
-        incomlete_qs = self.scraper.get_review_from_db(
+        incomplete_qs = self.scraper.get_review_from_db(
```
chris48s, line 60: :+1: leave things a little better than we found them

every_election/apps/organisations/boundaries/boundary_bot/tests/test_validation.py
chris48s: Yeah, so basically what we've done here is just set up the minimum subset of data needed for the test rather than a completely realistic record. You've updated it correctly :+1:

every_election/apps/organisations/boundaries/boundary_bot/scraper.py
awdem: Yeah, I spent a bit of time trying to figure out exactly why they were breaking the scraper, but I decided to move on for times sake.

chris48s: OK. Lets stop pulling the thread for this PR. At some point areas in that list will have a review that we need to pick up though, so I think it is worth spending a bit of time to work out what the problems are here and how we might fix them (if we can). Can you write it up as a ticket in Asana and assign it to yourself.

chris48s:
Next jobs on this

1. @awdem : Squash fixups
2. Merge/deploy
3. @chris48s or @GeoWill : SSH into prod and run 2x management commands
4. Wait till the scraper runs tomorrow - if all goes well it should run to completion and push some notifications :crossed_fingers: 
5. @awdem : Create Asana ticket: https://github.com/DemocracyClub/EveryElection/pull/2225#discussion_r1763143501

every_election/apps/organisations/boundaries/boundary_bot/scraper.py
awdem: Yup, will do.

PR #2255, opened 2024-09-26
Basic auth on dev and stage

awdem:
See [Asana Card](https://app.asana.com/0/1208197964378161/1208294218492449/f)

Bumps dc utils to 7.0.0 and adds DC BasicAuthMiddleware with the following url patterns excluded:

```
/_status_check/
/api
/api/*
```

I tested it by running the dev and staging servers to make sure there was auth.

---
  - https://app.asana.com/0/0/1208397645646728

### DIFF
--- a/every_election/settings/base.py
+++ b/every_election/settings/base.py
@@ -137,6 +137,7 @@ MIDDLEWARE = [
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.contrib.messages.middleware.MessageMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
+    "dc_utils.middleware.BasicAuthMiddleware",
 ]
 
 ROOT_URLCONF = "every_election.urls"
@@ -289,9 +290,12 @@ AWS_STORAGE_BUCKET_NAME = NOTICE_OF_ELECTION_BUCKET
 AWS_S3_FILE_OVERWRITE = True
 AWS_DEFAULT_ACL = None
 
-# Disable Basic Auth by default
-# We only want to use this on staging deploys
-BASICAUTH_DISABLE = True
+# Allowlist of URLs that should be ignored by DC BasicAuthMiddleware
+BASIC_AUTH_ALLOWLIST = [
+    "/",  # load balancer health check
+    "/api",
+    "/api/*",
+]
 
 # elections where polling day is in the range
 # (NOW - CURRENT_PAST_DAYS) - (NOW + CURRENT_FUTURE_DAYS)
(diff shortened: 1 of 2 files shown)


every_election/settings/base.py
```diff
@@ -289,9 +290,12 @@ def get_ec2_ip():
 AWS_S3_FILE_OVERWRITE = True
 AWS_DEFAULT_ACL = None
 
-# Disable Basic Auth by default
-# We only want to use this on staging deploys
-BASICAUTH_DISABLE = True
```
awdem, line 294: This is residual code (see https://github.com/DemocracyClub/dc_django_utils/pull/86#issuecomment-2276132107) so I've removed it. Django basic auth isn't even `requirements.txt` anymore so I assume this variable wasn't doing anything?

every_election/settings/base.py
```diff
@@ -289,9 +290,12 @@ def get_ec2_ip():
...
+# Allowlist of URLs that should be ignored by DC Basic Auth on dev and stage
+BASIC_AUTH_ALLOWLIST = [
+    "/_status_check/",
```
awdem, line 295: I copied `"/_status_check/"` from: https://github.com/DemocracyClub/WhoCanIVoteFor/blob/7b5eaade4670d83b60ecbcb584a83b8bd458a7b4/wcivf/settings/base.py#L335

But I wasn't sure how to test it locally unlike the other two patterns. I tried `http://localhost:8000/_status_check/` but that is not a valid url. What url does the load balancer ping when it does its health check?

symroe: This is defined in the CDK stack:

https://github.com/DemocracyClub/EveryElection/blob/8109f2e87e47ec984efb641f620e4fae05267f8a/cdk_stacks/stacks/code_deploy.py#L105

So it's looking at `/`

awdem: Ah I see, thanks. I'll correct it.

every_election/settings/base.py
```diff
@@ -289,9 +290,12 @@ def get_ec2_ip():
 AWS_S3_FILE_OVERWRITE = True
 AWS_DEFAULT_ACL = None
 
-# Disable Basic Auth by default
-# We only want to use this on staging deploys
-BASICAUTH_DISABLE = True
+# Allowlist of URLs that should be ignored by DC BasicAuthMiddleware
+BASIC_AUTH_ALLOWLIST = [
+    "/",  # load balancer health check
```
symroe, line 295: So this is going to work at a technical level, but of course it does mean that the home page on non-prod sites will be exposed. I don't thin that's a big deal for this project, as the home page does very little and isn't going to have a lot of content on it, but in other projects where this might matter then I think we should change the status check location, or add a new URL. 

Not something for this PR, but something to consider in others like it.

symroe (approved):

PR #2282, opened 2024-10-24
Migrate to UV

chris48s:
Refs
- https://app.asana.com/0/1204880927741389/1208507992823187/f
- https://app.asana.com/0/1207789998884860/1208511245881286/f
- https://app.asana.com/0/0/1208511245881289/f

The core point of this PR is to migrate us from pip to uv.

However, I have scope-creeped (scope-crept?) this a bit to cover several other related tasks which made sense to address at the same time. So in reality, the things happening in this PR are:

## 1. UV
Migrate from pip to UV, including related CI and deployment changes.

## 2. Audit packages
Instead of just blindly re-creating our existing pip setup, I've taken the opportunity to do a bit of an audit of our packages. This includes a bit of fiddling with the optional groups we have but more importantly I have removed a bunch of packages that we no longer use, no longer need or were there to ensure we install a compatible version of a second-order package (which is now better achieved by uv's resolver/lockfile).

As we go through this process on other repos I would propose we do the same. Writing a new manifest seems like a good opportunity to review what we're packages we are pulling in and why.

Most of the action on this is in https://github.com/DemocracyClub/EveryElection/pull/2282/commits/6d5d0d10addc24f3c143b49a75b5219db63794b1 and https://github.com/DemocracyClub/EveryElection/pull/2282/commits/7ad181a819f1e1e247b6b39026910f505560e4de

## 3. Dependabot
I've done a bit of a review of our dependabot config

Because dependabot [doesn't support uv.lock](https://github.com/dependabot/dependabot-core/issues/10478), I've written a GitHub action which will run `uv lock` and push the lockfile on any PRs submitted by dependabot.

I don't really know how this is going to work out yet.

I do know this will come with the following tradeoffs:

1. `@dependabot rebase` won't work because every PR will contain additional commits not by dependabot. I'm hoping we can rely on `@dependabot recreate`
2. Dependabot will only know about our first order dependencies, but not our transitive dependencies. This means that Dependabot can bump things that appear in our manifest but not things that only appear in our lockfile, so it won't be able to offer us bumps for security issues affecting transitive dependencies.

I don't really know how this is going to work out in practice - are there also other problems I haven't thought of?

I think my proposal for this is: I've intentionally left some packages out of date in this PR. I reckon we merge this, and then I'll try updating them with dependabot. We'll see if what I've done works, what the workflow looks like, where the pain points are, and find out what I haven't considered. Ideally I would quite like it if we could avoid having to migrate to renovate immediately on every repo where we adopt this. Renovate might solve several other problems for us, but being able to keep the scope of each PR to just "migrate from one package manager to another" to start with would be nice.

---

Final remarks:
Although one of the things we're doing here is doing this work for EE, this also hopefully lays down a template for how we're going to do it elsewhere. So it seems like a good place to bikeshed the details. Lets nail the blueprint here before we do it everywhere.

I've tried to break this down into sensible commits to review so you can see what I'm doing at each stage. There's a little bit of going round the houses on how to install UV where I did one thing and then another. In general, going commit by commit is probably helpful though.


.circleci/config.yml
```diff
@@ -70,7 +63,8 @@ jobs:
             python manage.py --version
             python manage.py check
             python manage.py makemigrations --check
-            pip check
+            uv lock --locked
```
chris48s, line 66: `uv lock --locked` is a bit of an odd command, but this is the equivalent of your `pipenv verify` or `poetry check`. Basically it will fail the build if we've updated `pyproject.toml` without updating the lockfile.

.circleci/config.yml
```diff
@@ -107,20 +102,20 @@ jobs:
...
     - run:
         name: Install CDK Python dependencies
         command: |
-          pyenv local $(ls -1 /opt/circleci/.pyenv/versions | grep 3.12)
-          python -m venv .venv
+          deployscripts/install_uv.sh
+          uv venv --python 3.12
           . .venv/bin/activate
-          pip install -r requirements/cdk.txt
+          uv sync --extra cdk --no-dev
```
chris48s, line 112: As discussed, `uv sync --extra cdk --no-dev` here will install all the main packages as well as the cdk extras.
Nothing _bad_ will happen. We're just installing more stuff than we really need to.

Upstream feature request: https://github.com/astral-sh/uv/issues/8292

.circleci/config.yml
```diff
@@ -85,6 +79,7 @@ jobs:
           name: Submit coverage
           command: |
             . .venv/bin/activate
+            uv pip install coveralls
```
chris48s, line 82: I considered moving this to a package in our manifest, but I can't really be bothered. I'm not sure how to meaningfully review a bump, for example.
Lets just stick with installing the latest version in CI and deal with it if it breaks.

.github/dependabot.yml
```diff
@@ -3,14 +3,20 @@ updates:
...
-  open-pull-requests-limit: 10
+    interval: "daily"
+  groups:
+    cdk-libs:
+      patterns:
+        - "aws-cdk*"
+        - "constructs"
+    boto-libs:
+      patterns:
+        - "boto*"
```
chris48s, line 14: I've taken some inspiration from WDIV on the groupings here.

Sadly dependabot doesn't allow us to do things like have a different schedule for different groups or have different settings depending on whether the packages are in

- `project.dependencies`
- `project.optional-dependencies.cdk`
- `project.optional-dependencies.production`
- `tool.dev-dependencies`

This is stuff you do get with renovate, but as noted elsewhere I'm kind of trying to avoid shaving that yak right now unless we have to.

.github/workflows/uv-lock.yml
```diff
@@ -0,0 +1,29 @@
+name: UV Lock
```
chris48s, line 1: Cards on the table: I have not run this code.
There isn't really an easy way to test this until we merge this PR.
Obviously if there's anything that looks drastically wrong here, flag it.
Otherwise, I'm going to suggest we merge this try it out and fix it if I broke everything.

INSTALL.md
```diff
@@ -3,11 +3,13 @@ EveryElection requires Python 3.x and Postgres.
 
 ## Install Python dependencies
 
+Every Election uses [uv](https://docs.astral.sh/uv/) to manage python packages.
```
chris48s, line 6: Because we are :sunglasses: the cool kids

cdk_settings.json
```diff
@@ -1,4 +1,4 @@
 {
   "base_ami_id": "ami-0c40c7ef04b5ab961",
-  "recipe_version": "0.0.35"
+  "recipe_version": "0.0.41"
```
chris48s, line 3: Guess how many times I had to push to get the image builder thing right

deployscripts/install_uv.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -xeE
+
+UV_CONSTRAINT=">=0.4.0,<0.5.0"
+
+if [ "$CI" = "true" ]; then
+    pip install uv"$UV_CONSTRAINT"
+else
+    sudo PIP_BREAK_SYSTEM_PACKAGES=1 pip install uv"$UV_CONSTRAINT"
+fi
```
chris48s, line 10: There are a couple of things to talk about in this file:

1. Based on my experience with ruff, I think we probably don't want to just always install the latest uv, at least at this stage. Pinning a constraint and accepting we're going to need to bump it is probably a better call while this is still zero-versioned and offers no stability guarantees IMO. `>=0.4.0,<0.5.0` seems pragmatic at this stage. Simultaneously, we'll probably want to bump it every so often. That's one of the costs we absorb of being early adopters. I abstracted this into this slightly clumsy script which I am calling in lots of places so that we've only got one place we have to bump the constraint. This is the reason why I'm not using the astral-sh/setup-uv action in `.github/workflows/uv-lock.yml`.

2. One of the problems I was trying to solve here is that on the EC2 instances I wanted to install uv in such a way that regardless of whether we `aws ec2-instance-connect ssh --os-user every_election` or `aws ec2-instance-connect ssh --os-user ubuntu`, we've got `uv` installed and on the path. The simplest way to do that was smoosh it into the global python env with pip, rather than curl-bash it. There's more detail on this in the commit message for https://github.com/DemocracyClub/EveryElection/commit/ea89badaf4e6d13457ed717ecf831c289a2a7223

pyproject.toml
```diff
@@ -1,5 +1,84 @@
-[tool.black]
-line-length = 80
+[project]
+name = "EveryElection"
+version = "0.0.0"
+description = "For recording every election in the UK"
+readme = "README.md"
+requires-python = "==3.12.*"
+dependencies = [
+    "beautifulsoup4==4.12.1",
```
chris48s, line 8: I reckon as a convention going forward: For applications, manifest should contain only pins not ranges. Obviously you want ranges for libraries.

cdk_stacks/components/install_app.yml
```diff
@@ -43,12 +43,11 @@ phases:
             - git checkout {{ git_branch }}
             - git pull --rebase origin {{ git_branch }}
             - git branch
-            - python3 -m venv .venv
+            - ./deployscripts/install_uv.sh
```
chris48s, line 46: As we said on the call, we'll probably want to build a new EE base image with uv in it, but while other apps are not yet migrated to UV lets not do that

.circleci/config.yml
```diff
@@ -60,7 +55,7 @@ jobs:
           paths:
             - ./.venv
             - ./node_modules
```
symroe, line 55: Adding `$HOME/.cache/uv` here will mean that `uv` doesn't need to re-download packages between runs.

cdk_stacks/components/add_user.yml
```diff
@@ -1,7 +1,7 @@
 name: AddUser
 description: Adds a system user
 schemaVersion: 1.0
-component_version: 0.0.4
+component_version: 0.0.5
```
symroe, line 4: What changed here?

symroe (commented):
I commented on commits as I went so missed some refactors that went on later. I've left these comments in case they're useful, but I think you can ignore some due to you having had the same thought as me :)

.circleci/config.yml
chris48s: Well `uv run --with coveralls coveralls` worked, but first I copied and pasted what you said and then spent ages being confused about why the coverage report wasn't submitting :D

.circleci/config.yml
```diff
@@ -23,7 +23,7 @@ jobs:
           node-version: '18'
 
       - restore_cache:
-          key: v3-dependencies-{{ checksum "requirements/base.txt" }}-{{ checksum "requirements/testing.txt" }}-{{ checksum "requirements/cdk.txt" }}
+          key: v3-dependencies-{{ checksum "pyproject.toml" }}-{{ checksum "uv.lock" }}
```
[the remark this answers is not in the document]
chris48s: Yeah good question. I went back and forth on this in my own head.

My thinking on this was that because dependabot can't bump `uv.lock` it will try to push commits that change `pyproject.toml` but don't change the lockfile this might be sensible. Maybe it is irrelevant though. `uv lock --locked` should ensure we just fail the build early doors on those commits anyway.

I've removed it in https://github.com/DemocracyClub/EveryElection/pull/2282/commits/46d6de18b423d2409196adc6263a8447621649ce

.circleci/config.yml
```diff
@@ -35,22 +35,17 @@ jobs:
           name: Install app dependencies
           command: |
             sudo apt update && sudo apt install -y gdal-bin python3-gdal python3-dev
-            pyenv local $(ls ~/.pyenv/versions/)
-            python -m venv .venv
+            curl -LsSf https://astral.sh/uv/0.4.23/install.sh | sh
```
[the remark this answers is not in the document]
chris48s: I think this comment is irrelevant given the changes I made in subsequent commits 1e57ca19b4acd9354c789177d12689d00dc2b363 and ea89badaf4e6d13457ed717ecf831c289a2a7223.
Do you agree?

.circleci/config.yml
```diff
@@ -35,22 +35,17 @@ jobs:
...
+            echo 'export PATH=$HOME/.cargo/bin:$PATH' >> $BASH_ENV
+            source $BASH_ENV
+            uv venv --python 3.12
```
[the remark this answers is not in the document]
chris48s: I think we actually have opposite objectives here.

My thinking here was: Previously we were doing all sorts of pyenv faffing to make sure we were using the right global python version:

- https://github.com/DemocracyClub/EveryElection/blob/0decd2dddd2d2e4c09c23e70d6ca74eac1c2ee57/.circleci/config.yml#L38
- https://github.com/DemocracyClub/EveryElection/blob/0decd2dddd2d2e4c09c23e70d6ca74eac1c2ee57/.circleci/config.yml#L114
- https://github.com/DemocracyClub/EveryElection/blob/0decd2dddd2d2e4c09c23e70d6ca74eac1c2ee57/.circleci/config.yml#L163
- https://github.com/DemocracyClub/EveryElection/blob/0decd2dddd2d2e4c09c23e70d6ca74eac1c2ee57/.circleci/config.yml#L199

I did at one point work out what we were trying to do there, but I've forgotten again.

My thinking here is: Lets bin all that and invoke `uv venv --python 3.12`. If uv can use a version of python that is already installed, cool. If it needs to go and download one, also fine. Either way, we're running the right version. That is one of the core value propositions of UV imo. Tbh we could even consider switching the base image for the build to one that doesn't have python on it at all.

I don't mind trying what you suggested, but also I'd like to understand what advantage that gives us, or what problem what I've suggested causes.

Happy to just talk about this one on a call or whatever.

.circleci/config.yml
chris48s: done in https://github.com/DemocracyClub/EveryElection/pull/2282/commits/9d49161f601919fbe61c0b7b83a56af551b58c23

.circleci/config.yml
chris48s: https://github.com/DemocracyClub/EveryElection/pull/2282/commits/889c102c56fbc7183bf1bd91d55f056a61da3105

.circleci/config.yml
chris48s: OK, so there's a lot going on in https://github.com/DemocracyClub/EveryElection/pull/2282/commits/c5b78bef84e266632c1e728af3a8250b54e0ee7f .

As of UV 0.4.27
https://github.com/astral-sh/uv/releases/tag/0.4.27

UV now has `dependency-groups`: https://docs.astral.sh/uv/concepts/dependencies/#dependency-groups
This is a distinct feature from `optional-dependencies`: https://docs.astral.sh/uv/concepts/dependencies/#optional-dependencies

..although there is obviously a lot of overlap between these two concepts.
I think if you're using uv as an application package manager, they basically just reduce to 2 overlapping ways to accomplish almost the same thing.
If you're using uv to build packages, `optional-dependencies` allow you to define optional groups that are "public" to your user so you can `pip install package[optional]`, whereas `dependency-groups` are "private".

Switching from `optional-dependencies` to `dependency-groups` means we can now use the `--only-group` flag to just install CDK dependencies, which is good. There's no equivilent of the `--all-extras` flag though, so if we want everything we have to either do `uv sync --group cdk --group production` or add everything to our [default groups](https://docs.astral.sh/uv/concepts/dependencies/#default-groups) and exclude the ones we don't want when we don't want them. I think what I've done here is probably reasonable.

The `dev-dependencies` feature that was current when I wrote this PR ~a week ago is now "legacy"
https://docs.astral.sh/uv/concepts/dependencies/#legacy-dev-dependencies

I think this is a pretty good microcosm of what we're getting involved in here. Both in terms of the rough and the smooth (or the ruff and the smooth, if you prefer :badpundog: ).

On the one hand, uv is moving quickly :rocket: A feature we wanted ended up in a release within a week or so and it was already something they were looking at before they triaged the issue.
On the other hand, uv is moving quickly :scream: A feature that was current a week ago is now considered legacy as of a patch release and there are several overlapping ways to do quite similar things.

I suspect this will all shake out by the time they're offering any sort of stability but I reckon there will be some more churn between now and then. I will also note that this whole thing is basically something poetry went through over a longer timeframe before they really got it right.

..anyhooo, in https://github.com/DemocracyClub/EveryElection/pull/2282/commits/c5b78bef84e266632c1e728af3a8250b54e0ee7f I have

- said we need uv>=0.4.27
- converted our `optional-dependencies` and `dev-dependencies` 
- updated the relevant `uv sync` commands

chris48s:
Right. I've pushed some commits and replied to some comments.. back to you

.circleci/config.yml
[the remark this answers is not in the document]
symroe: I guess the only other consideration here is the python version. At the moment we tend to pump the `v[n]` prefix when we bump python versions, but including `pyproject.toml` would mean we bust the cash a lot more often. Not a big deal, I think what you currently have is reasonable :+1:

.circleci/config.yml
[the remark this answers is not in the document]
symroe: Yes I think so too, but I kept it in on the off-chance you'd not need some of the env vars / flags I pointed out :+1:

.circleci/config.yml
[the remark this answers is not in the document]
symroe: I basically want to ensure we don't drift between [Circle]CI image version and the Python version in the venv. 

At the moment the build isn't going to fail if we use `cimg/python3.8` and `uv venv --python 3.12`, but it _will_ download a python (maybe for every CI run, depending on caching). If someone doesn't know this they might see the image version and assume that's what we're running. When we move to Python3.13 we might just bump the image, etc, etc.

I would like to be in a position where we either only specify the python version in one place (`uv`) or where we prevent `uv` from installing a version of Python that's not installed.

The `pyenv` faffing was due to Pyenv being too picky. If `3.12` is in `python-version` but you have `3.12.4` installed, it complains. You can't (or couldn't at the time) specify `3.12.*` or anything like that. I don't think `uv` has this "feature" so it's moot here.

symroe:
Great, I've replied where I think it might be useful, but basically this is good to go.

I can see we're going to need to stay on top of `uv` changes quite a bit, especially when this pattern is in more than one project. 

Would you mind writing a "how to even UV at DC" document, where you pull out a load of these decisions? I'd like a document we can use when working across DC that explains the state of play and how we're using it. Does that sound useful to you?

symroe (approved):

PR #2304, opened 2024-11-05
prune uv cache before persisting

chris48s:
As recommended by https://docs.astral.sh/uv/concepts/cache/#caching-in-continuous-integration

Run `uv cache prune --ci` before we perisist `$HOME/.cache/uv` to cache


symroe (approved):
Not 100% sure, but you might want to bump the cache version key as well, otherwise anything in the current cache will hang around. It might not matter, but I think thisis how it works

PR #2307, opened 2024-11-11
raise if we get error fetching lookup_name_to_registry

chris48s:
Refs https://democracy-club-gp.sentry.io/issues/6056737616

We hit this error running Boundary Bot last night.
I tried running locally to see if I could reproduce it, but no dice.
My guess is this will just work tomorrow.

My theory about what happened is: We got some kind of transient error fetching the file from github (which we ignored). The error response contained a non-200 OK status code and a response body which then did parse into a "valid" CSV file but didn't match our expected format.

I propose we add a `r.raise_for_status()` after we fetch the file. This would make that class of error more explicit if it happened again and it is good practice anyway. If we see this error again, we can review.

Also, while I was working on that: Quick drive by fix for the help text on the management command. Looks like a copy & paste error.


symroe (approved):

symroe:
Sentry context shows we got a `Forbidden` 
image

chris48s:
Good find :+1: I didn't spot that in Sentry but that confirms my hunch

PR #2313, opened 2024-11-14
Add management command to update `requires_voter_id` property

chris48s:
Refs: https://app.asana.com/0/1204880927741389/1208687275764621/f

As agreed, I'm making 2 changes in this PR:

1. Enable editing `requires_voter_id` in the django admin
2. Add management command to update `requires_voter_id` property

That said, cards on the table: This should have been a really simple PR, but I have got a bit distracted by some architecture astronauting here and made it a bit more complicated. I promise it is good architecture astronauting though!

In terms of solving the original problem, once this is deployed I'll

- SSH in
- run `python manage.py refresh_voter_id local.city-of-london.2025-03-20`

and then the next step will be to work out whether that just magically syncs through to everywhere else or not.

### DIFF
--- a/every_election/apps/elections/models.py
+++ b/every_election/apps/elections/models.py
@@ -9,7 +9,7 @@ from django.contrib.gis.db.models.functions import Distance
 from django.core.exceptions import ObjectDoesNotExist, ValidationError
 from django.core.files import File
 from django.db import models, transaction
-from django.db.models import JSONField
+from django.db.models import JSONField, Q
 from django.db.models.fields.related_descriptors import (
     create_reverse_many_to_one_manager,
 )
@@ -157,6 +157,12 @@ class Election(TimeStampedModel):
         ],
     )
 
+    def _get_manager_obj(self, manager):
+        for m in self._meta.managers:
+            if m.name == manager or m == manager:
+                return m
+        raise ValueError("Unknown manager {}".format(manager))
+
     def get_children(self, manager):
         """
         This method allows us to call with a manger instance or a string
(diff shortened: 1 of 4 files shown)


every_election/apps/elections/models.py
```diff
@@ -165,15 +171,39 @@ def get_children(self, manager):
...
-            if m.name == manager or m == manager:
-                child_manager_cls = create_reverse_many_to_one_manager(
-                    m.__class__, self._meta.get_field("_children_qs")
+        manager = self._get_manager_obj(manager)
+        child_manager_cls = create_reverse_many_to_one_manager(
+            manager.__class__, self._meta.get_field("_children_qs")
+        )
+        return child_manager_cls(self)
+
+    def get_descendents(self, manager, inclusive=False):
```
chris48s, line 182: So when I came to this task, what I wanted to do was start with an election object and then fetch anything under it in the tree, without caring where I was starting in the tree. I was hoping there would be a method that does this already, but no dice, so I decided to write one.

One of the cool things I've been exposed to through some years of working on wagtail is that their page tree query API is really nice to use and very well thought out. On a page object, you get

```py
page.get_parent()
page.get_children()
page.get_ancestors(inclusive=False) # everything above page in the tree optionally including/excluding page itself
page.get_descendants(inclusive=False) # everything below page in the tree optionally including/excluding page itself
page.get_siblings(inclusive=True) # everything on the same level as page in the tree optionally including/excluding page itself
```

and a nice thing about this is that aside from `.get_parent()` (which returns a single object) all of them return a Django Queryset so you can just chain a `.filter()` or whatever on the end of that. Anyway, my point is: I think we should completely rip off their API for EE. Although our tree is only 3 levels deep, being able to navigate the Election object hierarchy like this would be totally sweet.

Anyway, step one.. I needed `Election.get_descendants()` so I've implemented it here. Lets do more of this though. I think I also want to change `get_children()` to return a queryset instead of a manger for consistency, but that's a refactor for another day.

every_election/apps/elections/models.py
```diff
@@ -157,6 +157,12 @@ class Election(TimeStampedModel):
         ],
     )
 
+    def _get_manager_obj(self, manager):
```
chris48s, line 160: Having written this, I wish Chris from the past hadn't tried to be clever here. If this just took a manager object, this would have been simpler. Ho hum.

every_election/apps/elections/models.py
chris48s: OK, so I've really gone round the houses on the implementation of this function.

My first method using `.union()` in 96c3205 is just a conceptually hideous query. Lets not do that.

My second go (building up a list of IDs and doing an `IN` query) in b654872 is conceptually nicer.

Both of these are not very efficient in the case where we have a lot of org groups though because we have to individually call `get_children()` for each one. You could easily end up making hundreds of round-trips to the DB this way.

My third attempt in e910118 is way more efficient because it uses the structure of the election IDs to just make one query that gets all the descendents. This is way more efficient, but it totally ignores the actual `ForeignKey` relationships between records and just makes assumptions based on the known pattern of the IDs.

I think on balance, option 3 is probably still the best.

Thoughts? Also, can you see anything in e910118 I have not thought of?

every_election/apps/elections/models.py
```diff
@@ -165,15 +171,45 @@ def get_children(self, manager):
...
+
+        id_parts = self.election_id.split(".")
+        head = id_parts[:-1]
+        tail = id_parts.pop(-1)
+
+        prefix = f"{'.'.join(head)}."
+        suffix = f".{tail}"
+
+        return queryset.filter(
+            Q(election_id__startswith=prefix) & Q(election_id__endswith=suffix)
```
symroe, line 211: Quick drive by before a longer review: I'd use `election_date` here rather than `__endswith`.

every_election/apps/elections/models.py
```diff
@@ -195,15 +195,13 @@ def get_descendents(self, manager, inclusive=False):
...
 
         if self.identifier_type != "ballot":
             for child in self.get_children(manager).all():
-                queryset = queryset.union(
-                    child.get_descendents(manager, inclusive=True)
-                )
+                ids.update(child.get_descendents(manager, inclusive=True))
 
-        return queryset
+        return manager.filter(election_id__in=ids)
```
symroe: Could this also be a `startswith` + date filter to save using an `IN` query?

chris48s: This line isn't in the final diff. It is from an outdated commit.
As I said, I went round the houses a bit on this.
I will squash the history into something sensible before I merge this.

symroe (approved):

Issue #2318, opened 2024-11-20
Boundary Bot doesn't scrape legislation links or titles for ECOs before their effective date


awdem:
Boundary Bot doesn't scrape legislation links or titles for ECOs before their effective date. 

The relevant code is:
https://github.com/DemocracyClub/EveryElection/blob/77adf175a4b80e0a5841764e35319049e6eda504/every_election/apps/organisations/boundaries/boundary_bot/spider.py#L90-L134

This method in the scrapy spider gets the ECO link and title from containers like this one on the LGBCE review detail pages:
![image](https://github.com/user-attachments/assets/f761573c-b61f-498e-bca8-1470ba6b479e)

However, reviews that are made or laid, but haven't reached their effective date yet, have containers like this:
![image](https://github.com/user-attachments/assets/a8e7f257-d8ba-4eac-b0aa-552d752dff3a)
This container doesn't have the legislation title in it, so `get_eco_title_and_link` won't see it as a relevant review and won't return it. 

I'm currently processing ECOs still in 'draft' and I have to manually enter the title and legislation links. If we're going to continue to process draft ECOs, then we might want to adapt this method to handle cases like the above.

awdem:
> Do you mean it doesn't scrape **before the effective date** or it doesn't scrape **before the legislation has been made** (and a link to the made legislation added to the page)? I thought it was the latter, but could be wrong.

I mean the latter, sorry. I will correct the description/title

chris48s:
Given that importing draft ECOs is something we should be doing infrequently, a little bit of friction there might actually be a good thing tbh. i.e: having to go get the URL and paste it in the admin might gives us a bit of a "yes I really do mean to be importing draft legislation" in the handful of cases we want to do this.

awdem:
Do we want to have links to draft legislation anyway though? Since we have records in our db for draft ECOs, but we don't have their names or legislation links?

PR #2319, opened 2024-11-20
Hotfix/fixes to allow ukdsi links in eco processing

awdem:
I just processed the first ECO that is laid but still in 'draft' and I had to make a few changes to in order to use the handy `write_to_s3` button in the admin console:

- The `OrganizationBoundaryReview` model can now clean links that contain `ukdsi` as a subdomain
- Draft legislation pages hide their xml links differently so I've changed `get_xml_link_from_eco_url ` to handle that. I've not done a comprehensive review of xml links for draft legislation, but I'm hoping that they're won't be weirder edge cases that come up as I do more processing - we'll see.
- I've adapted existing tests to cover these cases.

### DIFF
--- a/every_election/apps/organisations/boundaries/lgbce_review_helper.py
+++ b/every_election/apps/organisations/boundaries/lgbce_review_helper.py
@@ -196,7 +196,10 @@ class LGBCEReviewHelper:
 
         parts = ["schedule/1", "schedule", "schedules"]
         for part in parts:
-            xml_link = f"{eco_url}/{part}/made/data.xml"
+            xml_resource_path = (
+                "/made/data.xml" if "ukdsi" not in eco_url else "/data.xml"
+            )
+            xml_link = f"{eco_url}/{part}{xml_resource_path}"
             response = requests.head(xml_link, allow_redirects=True)
             if response.status_code == 200:
                 return xml_link
--- a/every_election/apps/organisations/boundaries/tests/test_lgbce_review_helper.py
+++ b/every_election/apps/organisations/boundaries/tests/test_lgbce_review_helper.py
@@ -337,6 +337,102 @@ class TestLGBCEReviewHelper(TestCase):
                 ),
             )
 
+    def test_get_xml_link_from_ukdsi_eco_url_a(self):
+        def get_responses(link, **kwargs):
+            """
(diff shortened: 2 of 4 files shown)


GeoWill:
This will mean we have draft urls in the db. Not a huge problem since they link to the made legislation, but seems a bit of a shame if we do a lot like this. I guess if we manually fix them as the ECOs get made, it's an opportunity to double check nothing changed. Will Boundary Bot still alert us when the review changes to 'Completed' on the LGBCE site?

awdem:
> This will mean we have draft urls in the db. Not a huge problem since they link to the made legislation, but seems a bit of a shame if we do a lot like this. I guess if we manually fix them as the ECOs get made, it's an opportunity to double check nothing changed. Will Boundary Bot still alert us when the review changes to 'Completed' on the LGBCE site?

Yeah it should still alert us. And I believe that as long as the legislation year stays the same in the url then boundary bot **should* * update the existing db record's draft url with the made legislation (as long as the LGBCE site has updated that is).

Also, I think Boundary Bot doesn't scrape the legislation urls of ECOS that are not marked as made yet on the LBCE site (see https://github.com/DemocracyClub/EveryElection/issues/2318). When I made these changes, I was working with a draft url that I manually added to the record in admin. 

We could possibly add a warning on to  the write_to_s3  admin page that you're about to write to s3 with a draft legislation link?

GeoWill:
Yeah, I think making it clear that's what's going to happen, and then having a way to track it is good. I think maybe asana is the place to track this. Perhaps we can have a 'imported with draft legislation' label, and then a reminder on the EE admin page to go update the card?

awdem:
> Yeah, I think making it clear that's what's going to happen, and then having a way to track it is good. I think maybe asana is the place to track this. Perhaps we can have a 'imported with draft legislation' label, and then a reminder on the EE admin page to go update the card?

Yeah Asana would be good! The only other reason it might be nice to have it on the admin page is that in the past the LGBCE has left a draft link in place of a made legislation link for completed reviews and it's easy to miss that when you're processing ECOs.

awdem:
Closing this as I've realized that this is only to make the write_to_s3 button work for draft links, which is handy locally but doesn't matter for processing on prod!

PR #2331, opened 2024-11-28
prevent elections being created when divisions have invalid territory

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208865695360845/f

Currently, if a division has a missing territory code, we ignore this and silently let the user create data which is subtly wrong and will lead to us telling end users they don't need ID when they do.

Controversial opinion: This is a _bad_ thing.

The place I would like us to get to is: territory_code is a required field, and we populate it as part of the process of importing new divisions. That's quite a big job though, or at least its more than I can quickly knock out in an afternoon.

The next best thing we can do is what I've done in this PR: if someone tries to create an election ID involving a division with a missing territory_code, throw an exception. Stop them doing it. Log something to sentry. Make a dev go and fix the bad data.

### DIFF
--- a/every_election/apps/elections/forms.py
+++ b/every_election/apps/elections/forms.py
@@ -225,6 +225,10 @@ class DivsFormset(forms.BaseFormSet):
                 )
 
                 for div in divisions_qs:
+                    if not div.territory_code:
+                        raise ValueError(
+                            f"Division {div} is missing territory code"
+                        )
                     kwargs["initial"].append(
                         {"division_name": div.name, "group": div.group}
                     )
--- a/every_election/apps/elections/tests/test_utils.py
+++ b/every_election/apps/elections/tests/test_utils.py
@@ -9,7 +9,7 @@ class TestGetVoterIdRequirements:
 
     @pytest.mark.parametrize(
         "territory_code,expected_result",
-        [("ENG", "EA-2022"), ("NIR", "EFA-2002"), ("", None), (None, None)],
+        [("ENG", "EA-2022"), ("NIR", "EFA-2002")],
     )
     def test_get_voter_id_requirement_expected_territory_codes(
         self, territory_code, expected_result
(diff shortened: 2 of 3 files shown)


chris48s:
Also failing due to safety
Again, imagine https://github.com/DemocracyClub/EveryElection/pull/2329/commits/23d42b50baa924c48784c34f863c25773fe3f38e was on this branch

every_election/apps/elections/forms.py
```diff
@@ -225,6 +225,10 @@ def __init__(self, *args, **kwargs):
                 )
 
                 for div in divisions_qs:
+                    if not div.territory_code:
+                        raise ValueError(
+                            f"Division {div} is missing territory code"
```
symroe, line 230: I don't like causing a 500 for end users who have already entered data into the ID creator.

We should throw a sentry warning if someone hits this, but not 500 for them. We can even show the user a message, but we should allow them to create elections for other divisions.

Imagine a situation where someone has selected all 100 organisations that have elections. One division throws a 500 and the end user has no idea why. 

More generally, I'd like to report on / catch this stuff before it gets to users. We have a `DivisionProblem` model that we can report into. We can print a big message to stdout whenever we save, we can email us on a cron every day, etc, etc

chris48s: OK. I think if we're not going to merge this as a stop-gap then the next place to spend time is making `territory_code` a mandatory field and populating it at division create time. I don't think there's much point in spending the time on more elaborate stop gaps while still allowing invalid data to be created.

symroe: I think just removing the field and showing a message is more or less the same as this PR and provides that stop gap.

I had another thought after writing this: really we should think of this project as making data for export. We should add validation to the _output_ rather than (or as well as) the _input_. So: we can raise errors in the API layer when other services try to read this, or generally validate that we're exporting the right thing

chris48s:
Instead of doing this, we're going to do https://github.com/DemocracyClub/EveryElection/pull/2345 instead

PR #2333, opened 2024-12-02
revise how set_division_territory_code cmd works

chris48s:
OK then. So quick summary of the problem first.

First off, there are 12 divisions already in our DB that already have the wrong `territory_code` on them, for wahtever reason.
These are all wards in Monmouthshire that are incorrectly set as being in England or wards in Cumberland that are incorrectly set as being in Scotland.

Knowing this, I had a look to see if we had any elections already created involving those divisions. Usefully, there weren't so we don't actually have any elections already created with the wrong ID requirements on them. So that's good :relaxed: 

Then secondly, I tried running this command against a dump from production and it assigned a number of wards in Shropshire to Wales and some in Northumberland to Scotland.

So I think my plan is:

1. Fix the command (more on this in a sec)

2. Go through and manually edit the affected records in EE and set the territory code to empty string in the admin

    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30848/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30829/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30830/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30833/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30835/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30839/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30810/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30843/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30844/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30845/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/30815/change/
    - https://elections.democracyclub.org.uk/admin/organisations/organisationdivision/31393/change/

3. Once we've deployed a fixed version of the command, SSH into prod and run

```
python manage.py set_division_territory_code
python manage.py refresh_voter_id local.2025-05-01
```

That will:

- Fix the existing Monmouthsire/Cumberland wards that we un-set in step 2
- Assign `territory_code`s to all the divisions that are missing one
- Fix the ID requirements on all the elections we just created

..and then the next job is making sure those fixed ID requirements get synced to all the right places.. but lets not worry about that for now.

Meanwhile back at the point of this PR:

The existing `set_division_territory_code` command clearly has at least one bug.
Additionally, I found it is quite a difficult command do debug/reason about. Using a lot of `update()` statements is good for performance, but it makes it hard to understand which records have been modified by which rules.
Also, I think the previous approach is missing quite an important trick, so rather than try to "fix" what we had, I've taken a bit of a different approach. In this PR I am re-imagining this command in several ways:
1. Assign the correct territory codes to the correct divisions using some less ambiguous rules
2. Make the command explain exactly which values are being changed, and why. This also allows running with a `--dry-run` flag to explain what would happen without writing anything.
3. Say "I don't know" in preference to a possibly incorrect guess

This version doesn't do any kind of "this point is in that polygon" or "this polygon overlaps that polygon" type logic but should still cover most of the cases we need to deal with.

In terms of reviewing it, it is probably easier just to think of it as a new command than to look at the changes as a 'diff'

### DIFF
--- a/every_election/apps/organisations/management/commands/set_division_territory_code.py
+++ b/every_election/apps/organisations/management/commands/set_division_territory_code.py
@@ -1,7 +1,5 @@
-from typing import List
-
 from django.core.management.base import BaseCommand
-from organisations.models import Organisation, OrganisationDivision
+from organisations.models import OrganisationDivision
 
 GSS_TO_NATION = {
     "W": "WLS",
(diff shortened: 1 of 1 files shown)


every_election/apps/organisations/management/commands/set_division_territory_code.py
```diff
@@ -10,144 +8,114 @@
...
-        # Step 4: Loop over all remaining divisions and find a parent with a known code
-        self.stdout.write("Assigning from parents")
-        self.assign_from_parents()
+        messages = {
+            "ced": [],
+            "org_type": [],
+            "gss": [],
+            "parent_gss": [],
+            "fail": [],
+        }
```
chris48s, line 46: I'm collecting up the log messages like this so we group them by rule when we output them

every_election/apps/organisations/management/commands/set_division_territory_code.py
```diff
@@ -10,144 +8,114 @@
...
+                gss_head = org_geo.gss[0]
+                territory_code = GSS_TO_NATION[gss_head]
+                div.territory_code = territory_code
+                messages["gss"].append(
+                    f"Division {div.official_identifier}: Setting territory_code='{territory_code}'. Reason: Division is a child of {div.organisation.official_name}. Parent GSS code '{org_geo.gss}' starts with '{gss_head}'"
+                )
+            else:
+                messages["fail"].append(
+                    f"Division {div.official_identifier}: Not setting a Territory code."
+                )
```
chris48s, line 86: This is the most important thing I am doing differently in this PR.

I am using the fact that a division is a child of an organisation and basically every `OrganisationGeography` has a GSS code on it. This wasn't something we looked at in the previous model.

Right now there are only 3 `OrganisationGeography` records in EE with a missing GSS code and they're all combined authorities (which have no sub-divisions).

The situations where this will fail to give us a territory are:
1. We will end up with `OrganisationGeography` records that have a missing a GSS code when we have had to union together some districts into a new unitary authority and we don't have a code yet.
2. New Westminster constituencies. In this case, the parent org is the whole UK.

In these two cases, I'd rather bail at this point and say "we don't know" and make it a manual job for us, rather than do something error prone and silently decide bits of Monmouthsire are in England now actually, or whatever.

More generally, I'd like us to move to setting this `territory_code` when we import new boundaries given it is now a load-bearing value.

symroe (approved):
This seems reasonable. I think at the time, we've no back-posted GSS codes for some years so GSS wasn't a route we could use. Now we have that for most divisions, it's reasonable to use that. 

I still think that NUTS1 codes are trustworthy and could be used as a backup. It would catch the last few cases, I expect.

chris48s:
So the most common case here will still be that we want to set `territory_code` on new divisions that don't have a GSS code yet (because they are yet to appear in a BoundaryLine release).

The place where we do have pretty much complete GSS code coverage (which we were ignoring before) is at the Organisation level.

PR #2342, opened 2024-12-12
add renovate bot config

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208746898270758/f

In this PR I have taken an intial stab at a config file for renovate based on my current understanding of how it works.

This is one of those things where I've tried to write something sensible but I've not really tested it in anger (beyond checking the file validates with `npx --yes --package renovate -- renovate-config-validator --strict` see https://docs.renovatebot.com/config-validation/ )

We'll have to merge it, enable renovate, and see what happens to really see if this is doing exactly what we want when the rubber hits the road.

I've chosen to use JSON5 for the config file, which is basically JSON except you are allowed comments and trailing commas. Comments are really the key thing here.

Next Steps will be:

- Merge this
- Close all the open dependabot PRs
- Enable renovate and see what happens (I have not done this yet)

Given this is the first time we're reviewing a renovate config file, I'll put a bunch of comments inline on the diff explaining why I have made the specific choices I have.


renovate.json5
```diff
@@ -0,0 +1,55 @@
+{
+  "extends": [
+    "config:recommended",
+    ":pinAllExceptPeerDependencies",
+  ],
```
chris48s, line 5: Start off by extending renovate's default config for web apps
https://docs.renovatebot.com/presets-config/#configjs-app

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+
+  "timezone": "Europe/London",
+
+  "dependencyDashboardApproval": false,
```
chris48s, line 9: Dependabot's default behaviour is to tell you about upgrades in a "dependency dashboard" issue and then you can manually tell it to open PRs. This tells it to just open PRs without us manually telling it to.

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+  "prHourlyLimit": 0,
+  "branchConcurrentLimit": 0,
```
chris48s, line 11: Allow renovate to open as many PRs as it wants

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+  "automerge": false,
```
chris48s, line 12: Don't auto-merge anything

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+  "rangeStrategy": "pin",
```
chris48s, line 13: If renovate finds a range in our `pyproject.toml` or `package.json` convert it to a pin. For applications, there's not really a good case for ranges. Obviously not this for package repos.

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+
+  "enabledManagers": [
+    "pep621",
+    "npm",
+  ],
```
chris48s, line 18: Tell renovate to monitor/bump:

- Python packages in our PEP-621-compliant `pyproject.toml`
- NPM packages in `package.json`

We're not using it on this project, but renovate also has a `pre-commit` manager, which we can use on projects where we do have that set up. We can probably do something clever in `packageRules` to get it to bump the packages that appear in both `.pre-commit-config.yaml` and `pyproject.toml` in the same PR to keep them in sync.

We may also want to use the `github-action` manager on some projects, but I haven't added it here as this PR deletes the one GH action we were using.

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+
+  "packageRules": [
```
chris48s, line 20: This is my initial stab at a set of package rules for us. I'm hoping this sort of setup across projects will give us a volume of notifications we can realistically stay on top of :crossed_fingers:

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+    {
+      "groupName": "boto",
+      "matchManagers": ["pep621"],
+      "matchPackageNames": ["/^boto.*$/"],
+      "schedule": ["* 0-3 1 * *"], // monthly
+    },
```
chris48s, line 26: Group boto-related packages into one PR. Only bump once per month

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+    {
+      "groupName": "cdk",
+      "matchManagers": ["pep621"],
+      "matchPackageNames": ["/^aws-cdk.*$/"],
+      "schedule": ["* 0-3 1 * *"], // monthly
+    },
```
chris48s, line 32: Group cdk-related packages into one PR. Only bump once per month

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+    {
+      "matchManagers": ["pep621"],
+      "matchDepTypes": ["project.dependencies"],
+      "schedule": ["before 4am on monday"], // weekly
+    },
+    {
+      "matchManagers": ["npm"],
+      "schedule": ["before 4am on monday"], // weekly
+    },
```
chris48s, line 41: Bump main packages once per week. Also check NPM packages once per week. We don't have very many of these.

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+    {
+      "matchManagers": ["pep621"],
+      "matchDepTypes": ["dependency-groups"],
+      "schedule": ["* 0-3 1 * *"], // monthly
+    },
```
chris48s, line 46: Only check for updates to the packages under dependency-groups.cdk/production/dev monthly.

Tbh, ideally I would have set this to check `dependency-groups.production` weekly with `project.dependencies` but at the moment renovate only allows you to treat all of your `dependency-groups` as one block. On the one hand this is a bit annoying on the other hand

- this is a massive improvement on what dependabot gives us
- it might be possible to implement this ourselves with a regex rule

but I don't think I'm going to get super bogged down in it right now

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+    {
+      // We don't want notifications about Django 5
+      // until 5.2 LTS is released in 2025
+      "matchManagers": ["pep621"],
+      "matchPackageNames": ["django"],
+      "allowedVersions": "<5.0",
+    }
```
chris48s, line 53: Ignore django 5 for the moment, but still offer us django 4.2.x upgrades

renovate.json5
```diff
@@ -0,0 +1,55 @@
...
+  ]
```
chris48s, line 54: The one exception to all this is that I want renovate to give us security updates as soon as they are available.

As far as I can tell from the docs, this is the default behaviour and takes precedence over all our `packageRules`. However, I think to confirm that I have this right, once I set up renovate I am going to explicitly downgrade django to 4.2.16, push it, and make sure renovate offers us a the bump to 4.2.17 fairly quickly.

chris48s:
Another question I haven't thought about here yet.
I don't know if renovate automatically knows how to bump:
https://github.com/DemocracyClub/EveryElection/blob/debe61fbb85750298cfe3d14b2ea29b84f9588e5/pyproject.toml#L37-L38
If not, I'm fairly convinced we can write some rules that teach it how. I reckon lets just see what happens to start with.

symroe (approved):

symroe:
Let's see what happens!

PR #2343, opened 2024-12-12
automatically update subdivided geographies when geography is saved

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208894626914263/f

A problem we hit recently was that we created a bunch of new geography objects but forgot to update the corresponding subdivided tables. In this PR, I am setting it up so that every time we save an Org/Division Geography record, we also save the corresponding subdivided record.

There's a couple of possible gotchas here.

1. What I've done here is quite crude. Every time we save a Org/Division Geography we just nuke the related subdivided records and rebuild them, even if the only thing that changed was the gss code or something. I don't think this is a massive problem because we don't edit these often and when we do it is usually via management commands instead of through the frontend, but if you've got any ideas about how we could do this in a smarter way I'd be interested.

2. Because this is done in a `.save()`, if we were to modify Org/Division Geography records with a `bulk_create`, `bulk_update` or raw SQL (basically anything that bypasses the model `.save()` method) then this code would not run. In general this codebase has quite a lot of knock-on effects that happen in model `.save()` methods so I don't think this is really adding any extra constraints here.

### DIFF
--- a/every_election/apps/organisations/models/divisions.py
+++ b/every_election/apps/organisations/models/divisions.py
@@ -2,6 +2,7 @@ import re
 
 from core.mixins import UpdateElectionsTimestampedModel
 from django.contrib.gis.db import models
+from django.db import connection, transaction
 from django.db.models import Q
 from django_extensions.db.models import TimeStampedModel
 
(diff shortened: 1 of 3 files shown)


GeoWill (approved):
I think this looks good. 

Don't need to address is now, but I notice we've now got three copies of:

```
INSERT INTO organisations_divisiongeographysubdivided (geography, division_geography_id)
SELECT st_subdivide(geography) as geography, id as division_geography_id
FROM organisations_divisiongeography dg
```

(and same for the organisation geographies). I think what you did makes sense because the where clause you use is different to in the subdivided classes' `POPULATE_MISSING_SQL`.
I guess pulling those snippets out at might be sensible, but they're currently all next to each other in the file, and, it doesn't feel that like they're going to change much, so safe to leave.  

If they were in different files I think making an interface to do the updates through the subdivided classes might be better.

chris48s:
Yeah I did think about whether to try and DRY this up at all. Tbh, I wonder if we even need the constants any more as the only place they are used is in the management command that we're hoping to not run any more. I think I'll raise this as an issue.

chris48s:
raised https://github.com/DemocracyClub/EveryElection/issues/2357

PR #2345, opened 2024-12-16
convert territory_code to required field

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208891863227030/f

The purpose of this PR is to convert `OrganisationDivision.territory_code` to a required field and make the other necessary updates around that.


every_election/apps/organisations/boundaries/lgbce_review_helper.py
```diff
@@ -301,6 +301,7 @@ def get_base_eco_row(self, review: OrganisationBoundaryReview):
             "Mapit Generation URI": "",
             "Organisation ID": review.organisation.official_identifier,
             "Organisation ID type": "local-authority-eng",
+            "Territory Code": "ENG",
```
chris48s, line 304: The clue is very much in the name on this one. If we're doing a LGBCE review, it is England.

every_election/apps/organisations/management/commands/import_divisionsets_from_csv.py
```diff
@@ -31,9 +31,9 @@
 
 class Command(ReadFromCSVMixin, BaseCommand):
     help = """Import from CSV at URL with the headers:
-        Start Date, End Date, Name, official_identifier,
-        seats_total, Boundary Commission Consultation URL, Legislation URL,
-        Short Title, Notes, Mapit Generation URI, Organisation ID"""
+        Start Date, End Date, Name, official_identifier, seats_total,
+        Boundary Commission Consultation URL, Legislation URL, Short Title,
+        Notes, Mapit Generation URI, Organisation ID, Territory Code"""
```
chris48s, line 36: Tbh, I reckon there are some fields in here that we really really don't need
but I reckon that is another issue for another day

every_election/apps/organisations/management/commands/ni_import_district_electoral_areas.py
```diff
@@ -72,6 +73,7 @@ def create_divisions(self, csv_data):
                 slug=slugify(line["District Electoral Area"]),
                 division_type="LGE",
                 seats_total=line["Number of councillors"],
+                territory_code=TerritoryCode.NIR,
```
chris48s, line 76: Again, clue's in the name for this management command

every_election/apps/organisations/management/commands/set_division_territory_code.py
chris48s, line 1: So this management command was.. short-lived.
I did consider if it makes sense nick the logic from this and populate `territory_code` on save if it is missing but.. I think it has done its job. We've fixed all the historic data. From this point forward, we assign divisions to a territory at the point we make/import them. We don't need to do anything clever at the point we generate the CSVs for LGBCE reviews because by definition they're all for areas in England. Any other changes are going to be both infrequent and relatively bespoke.

PR #2367, opened 2025-01-06
Fix/allow auth headers

awdem:
Dev and stage sites aren't passing the Authorization header to the CF cache and basic auth wasn't working so this fixes that. 

The `/id_creator/*` pattern doesn't use the CF cache, and I couldn't add the Authorization header to the existing custom origin request policy, so I changed it to a managed policy that allows all headers through to the origin.

---
  - https://app.asana.com/0/0/1208998091856562

### DIFF
--- a/cdk_stacks/stacks/code_deploy.py
+++ b/cdk_stacks/stacks/code_deploy.py
@@ -365,21 +365,7 @@ class EECodeDeployment(Stack):
                 "/id_creator/*": cloudfront.BehaviorOptions(
                     origin=app_origin,
                     allowed_methods=cloudfront.AllowedMethods.ALLOW_ALL,
-                    origin_request_policy=cloudfront.OriginRequestPolicy(
-                        self,
-                        "origin_request_policy_id_creator",
-                        cookie_behavior=cloudfront.OriginRequestCookieBehavior.all(),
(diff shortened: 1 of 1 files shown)


cdk_stacks/stacks/code_deploy.py
```diff
@@ -361,21 +365,7 @@ def create_cloudfront(self, alb: elbv2.ApplicationLoadBalancer):
...
-                            "Accept",
-                            "Accept-Language",
-                            "Cache-Control",
-                            "Referer",
-                        ),
-                        query_string_behavior=cloudfront.OriginRequestQueryStringBehavior.allow_list(
-                            "reset"
-                        ),
-                    ),
+                    origin_request_policy=cloudfront.OriginRequestPolicy.ALL_VIEWER,
```
chris48s, line 368: @symroe I don't think I have the background on this one.
Was the original intent of this that there were particular request headers we explicitly wanted to drop?

chris48s: I think based on our discussion in standup this is fine. The reason for excluding certain headers was about cache-ability. Given the point of this rule is not to cache we should be fine to just pass through all the headers.

chris48s (approved):

PR #2368, opened 2025-01-07
create CAs with source=custom

chris48s:
Very small fix.
BoundaryLine doesn't carry "official" CA boundaries, so union-ing together Local Auths will be the best we ever get. Set `source="custom"` on create so the reports won't flag this as an issue.
Once we set a GSS on them, they're "done"


awdem (approved):

awdem:
I've set the source for the upcoming CA and CCA (Greater Lincolnshire Combined County Authority and Hull and East Yorkshire Combined Authority) to custom manually in admin just for consistency.

Issue #570, opened 2019-07-11
Dealing with Community Governance Reviews


chris48s:
Local authorities may sometimes make minor 'technical' boundary changes in a Community Governance Review. These are problematic because they result in boundary changes defined in an Electoral Changes Order, but because the reviews are conducted at local level (without LGBCE's involvement), there is no provisional boundary set available before the first election using the new boundaries and none of our existing tooling for dealing with boundary reviews applies. An ECO resulting from a CGR will never change the number of wards or the number of councillors representing each ward, but they do change the internal boundaries.

This usually means we have to run the first election using the old (incorrect) boundaries, accepting that this means there will be some users who we're assigning to the wrong ward. Then we have to retrospectively import the new boundaries once the May BoundaryLine release is issued.

Within that structure, there are 2 distinct cases:

1. A Community Governance Review is conducted after an LGBCE electoral review has been concluded, but before the first election has been conducted using the new boundaries. e.g: [The Test Valley (Electoral Changes) Order 2018](http://www.legislation.gov.uk/uksi/2018/36/contents/made) was subsequently amended by [The Test Valley (Electoral Changes) Order 2019](http://www.legislation.gov.uk/uksi/2019/200/made) before the first election using the new boundaries in May 2019.
2. A Community Governance Review is conducted making changes to electoral boundaries which have already been used to conduct one or more elections. e.g: [The Newark & Sherwood (Electoral Changes) Order 2014](https://www.legislation.gov.uk/uksi/2014/1907/contents/made) defined a set of boundaries which were used to conduct local elections in 2015. These were subsequently amended by [The Newark and Sherwood (Electoral Changes) Order 2016](http://www.legislation.gov.uk/uksi/2016/855/contents/made). This defined changes applicable to the next scheduled election in May 2019.

The first of those two cases is easiest to deal with - it doesn't really need any special handling. Because no election has ever been held using the provisional boundaries, there's no need to retain a record of them. We can run our usual [process](https://github.com/DemocracyClub/EveryElection/wiki/Boundary-Reviews-(2)) for back-porting codes/boundaries from BoundaryLine over provisional boundaries from LGBCE. The provisional boundaries will be overwritten. The elections we already ran using the provisional boundaries will now be attached to the correct ones and we'll use the right boundaries moving forwards.

[...]

PR #2393, opened 2025-01-14
add management command to assign election group to divisionset

awdem:
This command is intended to make the process described in https://github.com/DemocracyClub/EveryElection/issues/570 slightly easier.

It takes any number of election_ids for election groups and tries assign the divisions of their child elections to the divisions in a  given divisionset_id.

### DIFF
--- a/every_election/apps/elections/management/commands/assign_election_to_divisionset.py
+++ b/every_election/apps/elections/management/commands/assign_election_to_divisionset.py
@@ -1,0 +1,83 @@
+"""
+This management command assigns elections from election groups to a given divisionset's divisions.
+
+This is usually used when we're processing Community Governance Reviews as decscribed in this issue:
+
+https://github.com/DemocracyClub/EveryElection/issues/570.
+
+python manage.py assign_election_to_divisionset <election_ids> <divisionset_id>
+
+example:
+python manage.py assign_election_to_divisionset local.broxtowe.2023-05-04 local.broxtowe.2024-05-02 794
+"""
+
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+from elections.models import Election
+from organisations.models import OrganisationDivision
+
+
+class Command(BaseCommand):
+    help = "Tries to assign elections from election groups to a given divisionset's divisions"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "election_group_ids",
+            action="store",
+            nargs="+",
+            help="space separated list of election_ids",
+        )
+        parser.add_argument(
+            "divset_id",
+            action="store",
+            type=int,
+            help="The divisionset_id of the divisionset",
+        )
+
+    def handle(self, *args, **options):
+        divset_id = options["divset_id"]
+        election_group_ids = options["election_group_ids"]
+
+        elections = self.get_elections(election_group_ids)
+
+        self.assign_elections_to_divisionset(divset_id, elections)
+
+    def assign_elections_to_divisionset(self, divset_id, elections):
+        # Find election division in the new divisionset and update the election
+        for e in elections:
+            try:
+                new_div = OrganisationDivision.objects.get(
+                    divisionset_id=divset_id,
+                    official_identifier=e.division.official_identifier,
+                )
+                e.division = new_div
+                e.division_geography = new_div.geography
+                self.stdout.write(
+                    f"Assigned {e.election_id} to "
+                    f"{new_div.name} ({new_div.official_identifier}) from divisionset {divset_id}"
+                )
+            except OrganisationDivision.DoesNotExist:
+                raise CommandError(
+                    f"Division with official_identifier {e.division.official_identifier} not found in divisionset {divset_id}."
+                )
+        self.stdout.write("Saving...")
+        # Save all elections in a single transaction
+        with transaction.atomic():
+            for e in elections:
+                e.save()
+        self.stdout.write("...Done")
+
+    def get_elections(self, election_group_ids):
+        election_groups = Election.public_objects.all().filter(
+            election_id__in=election_group_ids
+        )
+
+        if len(election_group_ids) != election_groups.count():
+            raise CommandError(
+                f"Expected {len(election_group_ids)} election groups with ids {election_group_ids} but found {election_groups.count()}"
+            )
+
+        elections = []
+        for eg in election_groups:
+            elections.extend(eg.get_children("public_objects").all())
+        return elections


every_election/apps/elections/management/commands/assign_election_to_divisionset.py
```diff
@@ -0,0 +1,74 @@
+"""
+This management command assigns elections from election groups to a given divisionset's divisions.
+
+This is usually used when a Community Governance Review came into force the same day as an election
+but we were unable to import its boundary changes until the next BoundaryLine release post-election.
```
chris48s, line 5: In the explanation of why we would use this, lets link to https://github.com/DemocracyClub/EveryElection/issues/570

in lieu of any better written up docs, that's about the best resource we have

every_election/apps/elections/management/commands/assign_election_to_divisionset.py
```diff
@@ -0,0 +1,74 @@
...
+        parser.add_argument(
+            "divset_id",
+            action="store",
+            help="The divisionset_id of the divisionset",
+        )
+
+    def handle(self, *args, **options):
+        divset_id = options.get("divset_id")
+        if not divset_id.isdigit():
+            raise CommandError("The divisionset_id must be a number.")
```
chris48s, line 37: Argparse can perform validation for us. If you declare this as
```py
parser.add_argument(
    "divset_id",
    required=True,
    type=int,
    help="The divisionset_id of the divisionset",
)
```

`type=int` will perform the validation when we parse the arguments and we can drop this check.
https://docs.python.org/3/library/argparse.html#type

every_election/apps/elections/management/commands/assign_election_to_divisionset.py
```diff
@@ -0,0 +1,74 @@
...
+
+        election_groups = options.get("election_groups", [])
+        if not election_groups:
+            raise CommandError("No election groups provided.")
```
chris48s, line 41: Again, if you just set `nargs='+'` on the `election_groups` argument, argparse will check that at least one is present and you can remove this manual check.
https://docs.python.org/3/library/argparse.html#nargs

every_election/apps/elections/management/commands/assign_election_to_divisionset.py
```diff
@@ -0,0 +1,74 @@
...
+                )
+            except OrganisationDivision.DoesNotExist:
+                raise CommandError(
+                    f"Division with official_identifier {e.division.official_identifier} not found in divisionset {divset_id}."
+                )
+
+    def get_elections(self, election_groups):
+        election_groups = Election.public_objects.all().filter(
+            election_id__in=election_groups
+        )
```
chris48s, line 69: If I invoke this command with 3 election IDs and 2 were valid but I made a typo in the other one, then this will select 2 election objects and just carry on.

Can we validate here that the number of election objects we have got back from the DB is equal to the number of election IDs we invoked with, and throw an error if it is not.

every_election/apps/elections/management/commands/assign_election_to_divisionset.py
```diff
@@ -0,0 +1,74 @@
...
+                e.division_geography = new_div.geography
+                e.save()
+                self.stdout.write(
+                    f"Assigned {e.election_id} to "
+                    f"{new_div.name} ({new_div.official_identifier}) from divisionset {divset_id}"
+                )
+            except OrganisationDivision.DoesNotExist:
+                raise CommandError(
+                    f"Division with official_identifier {e.division.official_identifier} not found in divisionset {divset_id}."
+                )
```
chris48s, line 64: If something in here throws, this is going to leave the user in an inconsistent state. Also if any divisions don't match up by ID, we probably messed up badly.

I think I'd prefer to roll the whole thing back if that happens.

I've not tested this code, but how about something like this for a pattern?

```py
for e in elections:
    try:
        new_div = OrganisationDivision.objects.get(
            divisionset_id=divset_id,
            official_identifier=e.division.official_identifier,
        )
        e.division = new_div
        e.division_geography = new_div.geography
        self.stdout.write(
            f"Assigning {e.election_id} to "
            f"{new_div.name} ({new_div.official_identifier}) from divisionset {divset_id}"
        )
    except OrganisationDivision.DoesNotExist:
        raise CommandError(
            f"Division with official_identifier {e.division.official_identifier} not found in divisionset {divset_id}."
        )

self.stdout.write('saving..')
with transaction.atomic():
    for e in elections:
        e.save()
self.stdout.write('..done')
```

so basically the idea here is:

- We make sure everything matches up how we expect first, modifying the objects in memory as we go
- Once we've been though all that without throwing an exception, then we do a second pass to actually persist the objects to the DB
- Wrap all the writes in a `transaction.atomic`, just in case something fails when we write to the DB

chris48s (commented):
Thanks. Agree this looks like the lowest hanging fruit if we are going to make part of this process less cumbersome :+1:

every_election/apps/elections/management/commands/assign_election_to_divisionset.py
awdem: done in 027430705d1902ba1cf37adbb4adf4d13273915a

every_election/apps/elections/management/commands/assign_election_to_divisionset.py
awdem: Thanks I missed that. Done in c304a5edebe16f8d328c5f494cbc3324f60cedd9

every_election/apps/elections/management/commands/assign_election_to_divisionset.py
awdem: Done in c304a5edebe16f8d328c5f494cbc3324f60cedd9

every_election/apps/elections/management/commands/assign_election_to_divisionset.py
awdem: That makes sense. Done in faf211d9df1cdbdf20f2f4662f41160e06f87029.  I tested the change by using the command to assign an election to new divisionset, changing the gss code of one division, and then using the command again to try and return to the old divisionset.

awdem:
> Thanks. Agree this looks like the lowest hanging fruit if we are going to make part of this process less cumbersome 👍

Yeah I think the most tedious part was going through the reviews to extract which wards had changed and having to find their gss codes, both old and new, so I could update them in the admin. But that felt like a big job to automate.

every_election/apps/elections/management/commands/assign_election_to_divisionset.py
```diff
@@ -0,0 +1,83 @@
...
+        parser.add_argument(
+            "divset_id",
+            action="store",
+            type=int,
+            help="The divisionset_id of the divisionset",
+        )
+
+    def handle(self, *args, **options):
+        divset_id = options.get("divset_id")
+        election_group_ids = options.get("election_group_ids", [])
```
chris48s, line 39: Given these arguments are required, you should be able to do

```suggestion
        divset_id = options["divset_id"]
        election_group_ids = options["election_group_ids"]
```

In fact, it would be preferable to to throw `KeyError` if we are here and one of those args is not defined.

chris48s:
One more minor thing. Aside from that, it is good. I'll leave it with you to squash the fixups and merge.

Then one final job: Add a note on
https://github.com/DemocracyClub/EveryElection/issues/570
Saying you can now use this command so when us-from-the-future read that issue in a year's time we know it is here.

chris48s (approved):

PR #2396, opened 2025-01-20
design system/django-utils upgrade

chris48s:
Refs https://app.asana.com/0/1204880927741389/1208958646411291/f

This PR is quite similar to https://github.com/DemocracyClub/UK-Polling-Stations/pull/8210

This one covers

- Upgrade to design-system==0.7.0
- Account for breaking changes
- Fix the underlines on the map +/- buttons (which if we zoom all the way out was the original point of this work)
- Update the footer to account for latest design-system/dc_django_utils's opinions on that

The reason I'm not updating dc_django_utils in this PR is because I already upgraded it, but I missed that it broke the footer :disappointed: which is why it is all broken in production.

Also in this PR, I have disabled Safety CI. Maybe that is not the final position, but moving to `safety scan` requires more work than just "change the command" so lets remove it for now and cover next steps on that in another ticket.

### DIFF
--- a/every_election/settings/base.py
+++ b/every_election/settings/base.py
@@ -195,7 +195,6 @@ PIPELINE["STYLESHEETS"].update(
     {
         "map": {
             "source_filenames": [
-                "css/map.css",
                 "leaflet/dist/leaflet.css",
             ],
             "output_filename": "css/map.css",
--- a/every_election/templates/base.html
+++ b/every_election/templates/base.html
@@ -57,27 +57,42 @@
 {% endblock main_base %}
 {% block mailing_list %}{% endblock mailing_list %}
 {% block footer_menu %}
-    <div class="ds-cluster-center">
-        <ul>
-            <li><a href="{% url 'home' %}">Home</a></li>
-            <li><a href="https://democracyclub.org.uk/about/">About</a></li>
-            <li><a href="https://democracyclub.org.uk/privacy/">Privacy</a></li>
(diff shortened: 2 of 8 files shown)


every_election/templates/base.html
```diff
@@ -57,27 +57,42 @@
...
-            <li><a href="https://democracyclub.org.uk/blog/">Blog</a></li>
-            <li><a href="https://democracyclub.org.uk/contact/">Contact</a></li>
-            <li><a href="https://democracyclub.org.uk/donate/">Donate</a></li>
-            <li><a href="https://www.facebook.com/democracyclub/">Facebook</a></li>
-            <li><a href="https://twitter.com/democlub">Twitter</a></li>
-            <li><a href="https://github.com/DemocracyClub/EveryElection">GitHub</a></li>
-        </ul>
-    </div>
+    {% block footer_links %}
+        <div class="ds-footer-links">
```
chris48s, line 61: Same comment here as
https://github.com/DemocracyClub/UK-Polling-Stations/pull/8210#discussion_r1924009703
I've just copied over the links as they are. If we want to review the content, lets do that as another job.

every_election/settings/base.py
```diff
@@ -195,7 +195,6 @@ def get_ec2_ip():
     {
         "map": {
             "source_filenames": [
-                "css/map.css",
```
symroe, line 198: Do you want to delete this file too?

chris48s: Yes. Good catch :+1:

symroe (approved):

PR #2398, opened 2025-01-28
re-organise settings

chris48s:
Here is a WIP version of the restructured django settings we looked at in surgery.

There are some more things to do to finish this PR:

- ~Update the setup/install docs~
- ~Decide if we want to throw when `DJANGO_SETTINGS_MODULE` is not set, instead of defaulting to `every_election.settings.deploy` (need to think this one through)~
- ~Review exactly what goes into dev.py - might need to pull some of these out into env vars~
- ~Remove print statements~

but this should be ok to take for a quick spin locally. You'll need to move your DB settings to `.env` if you currently have them in `local.py`


manage.py
```diff
@@ -3,10 +3,14 @@
...
+from django.core.exceptions import ImproperlyConfigured
 
 if __name__ == "__main__":
-    dotenv.read_dotenv()
-    os.environ.setdefault("DJANGO_SETTINGS_MODULE", "every_election.settings")
+    dotenv.load_dotenv()
+    if not os.environ.get("DJANGO_SETTINGS_MODULE"):
+        raise ImproperlyConfigured(
+            "You must explicitly set DJANGO_SETTINGS_MODULE"
+        )
```
chris48s, line 13: In order for this to work, we now have to set
`DJANGO_SETTINGS_MODULE: every_election.settings.deploy`
in SSM.
I've set it in dev (hence this PR deploys) but not staging or prod yet. We need to do those before we deploy it.

chris48s:
OK. Next steps: Lets have a quick chat about https://github.com/DemocracyClub/EveryElection/pull/2398/commits/2b20d66dc92d03b62bf5f17773d2676e52433b8d and confirm we're happy with what goes where.
I haven't done anything towards failing loudly if required settings are not present in this PR. I think I'm going to say that issue is a follow up PR.

every_election/settings/local.example.py
```diff
@@ -1,34 +0,0 @@
...
-GCS_API_KEY = ""
-
-# don't use the real url in dev or staging
-# there is a webhook url for each environment, so
-# we can test the slack integration
-SLACK_WEBHOOK_URL = ""
-
-# AWS credentials
-AWS_ACCESS_KEY_ID = ""
-AWS_SECRET_ACCESS_KEY = ""
```
chris48s, line 24: I didn't reproduce these anywhere. I think we're all-in on role based auth now (i.e: `AWS_PROFILE=`). We don't explicitly provide keys.

chris48s:
OK. This is done. The build will fail deploying to deb because `DJANGO_SETTINGS_MODULE` is blank (I unset it again as it was breaking other PRs). Final move before merging this will be to set that var in all 3 envs.

symroe (approved):
I've read this, and we've talked about it. 

I think we're only _really_ going to know how it works once we've been using it for a while. I'd say as long as nothing breaks here, let's ship it and see how we get on in future and with other projects. 

At some point once we've run with this a while, we should write up some docs on what we've done (in the handbook, I expect)

chris48s:
Well the good news is we've got a pretty decent starting point for docs
https://docs.google.com/document/d/1QXit4Y_8kaV6CSe5MDKD50o2HQafiSUaBTZ6QGfEcKQ

PR #2399, opened 2025-01-29
adapt boundaryline_import_boundaries to import CED boundaries

awdem:
Adapts `boundaryline_import_boundaries` to import CED boundaries by matching against name and county rather than GSS code.

### Testing:

Set GSS code for a CED in admin ([source](https://geoportal.statistics.gov.uk/datasets/2dd3a997bd87443fa098deb797375610_0/explore)) and run the command using that code (you could also use the command described in https://github.com/DemocracyClub/EveryElection/wiki/Backporting-CED-codes to backport the codes for a county).

---
  - https://app.asana.com/0/0/1209157822204376

### DIFF
--- a/every_election/apps/organisations/boundaries/boundaryline.py
+++ b/every_election/apps/organisations/boundaries/boundaryline.py
@@ -19,6 +19,19 @@ class BoundaryLine:
         self.layer = ds[0]
 
     def merge_features(self, features):
+        """
+        When we have geographies with 'Detached parts', these are
+        represented in BoundaryLine as multiple records with the same code.
+
+        For example, an area like this
+        https://mapit.mysociety.org/area/12701.html
+
+        is represented as 2 Polygons.
+
+        We just want to represent this as a single MultiPolygon feature,
+        so if we've matched >1 records with the same code
+        merge them into a single MultiPolygon feature.
+        """
         polygons = []
         for feat in features:
             if isinstance(feat.geom.geos, MultiPolygon):
@@ -41,20 +54,27 @@ class BoundaryLine:
             )
         if len(matches) == 1:
             return matches[0].geom.geos
+        # handle 'Detached Parts'
+        return self.merge_features(matches)
 
-        """
-        When we have geographies with 'Detached parts', these are
-        represented in BoundaryLine as multiple records with the same code.
+    def get_feature_by_name_and_county(self, div_slug, org_slug):
+        matches = []
+        slug_tuple = (div_slug, org_slug)
+        for feat in self.layer:
+            feat_name = normalize_name_for_matching(feat.get("name"))
+            feat_county = normalize_name_for_matching(
+                feat.get("file_name").replace("_", "-")
+            )
+            if (feat_name, feat_county) == slug_tuple:
+                matches.append(feat)
 
-        For example, an area like this
-        https://mapit.mysociety.org/area/12701.html
-
-        is represented as 2 Polygons.
-
-        We just want to represent this as a single MultiPolygon feature,
-        so if we've matched >1 records with the same code
-        merge them into a single MultiPolygon feature.
-        """
+        if len(matches) == 0:
+            raise ObjectDoesNotExist(
+                f"Expected one match for {slug_tuple}, found 0"
+            )
+        if len(matches) == 1:
+            return matches[0].geom.geos
+        # handle 'Detached Parts'
         return self.merge_features(matches)
 
     def get_code_from_feature(self, feature):
--- a/every_election/apps/organisations/boundaries/helpers.py
+++ b/every_election/apps/organisations/boundaries/helpers.py
@@ -20,6 +20,8 @@ def normalize_name_for_matching(name):
         return slug[:-3]
     if slug.endswith("-ward"):
         return slug[:-5]
+    if slug.endswith("-county"):
+        return slug[:-7]
     return slug
 
 
--- a/every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
+++ b/every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
@@ -42,7 +42,7 @@ class Command(BaseBoundaryLineCommand):
         group.add_argument(
             "--code",
             action="store",
-            help="code in the form gss:X01000001 or unit_id:12345",
+            help="code in the form gss:X01000001",
         )
         group.add_argument(
             "--codes",
@@ -70,19 +70,16 @@ class Command(BaseBoundaryLineCommand):
         code_type, code = split_code(identifier)
 
         if code_type == "unit_id" and re.match(r"^\d+$", code):
-            # FIXME before 2021
             error = (
-                "Importing boundaries from BoundaryLine against CEDs is "
-                "not yet available because unit_id is not a stable identifier"
+                "Importing boundaries from BoundaryLine against unit_ids is "
+                "not available because unit_id is not a stable identifier"
             )
             raise ValueError(error)
 
         if code_type == "gss" and re.match(r"^[A-Z][0-9]{8}$", code):
             return True
 
-        raise ValueError(
-            "Unknown code type. Expected 'gss:X01000001' or 'unit_id:12345'"
-        )
+        raise ValueError("Unknown code type. Expected 'gss:X01000001'")
 
     def filter_records(self, identifier):
         # use code to find matching OrganisationGeography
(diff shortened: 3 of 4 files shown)


every_election/apps/organisations/boundaries/helpers.py
```diff
@@ -20,6 +20,8 @@ def normalize_name_for_matching(name):
         return slug[:-3]
     if slug.endswith("-ward"):
         return slug[:-5]
+    if slug.endswith("_county"):
```
awdem, line 23: All the records in `file_name` column in the boundary CED shapefile that corresponds to the name of the county end in _COUNTY so I trim it off here.

every_election/apps/organisations/boundaries/boundaryline.py
```diff
@@ -57,6 +57,23 @@ def get_feature_by_field(self, fieldname, code):
         """
         return self.merge_features(matches)
 
+    def get_feature_by_name_and_county(self, tuple):
+        matches = []
+        for feat in self.layer:
+            feat_name = normalize_name_for_matching(feat.get("name"))
+            feat_county = normalize_name_for_matching(
+                feat.get("file_name")
+            ).replace("_", "-")
```
awdem, line 66: I use `.replace("_", "-")` here because the values of the `file_name` column in the BoundaryLine CED shapefile that corresponds to the name of the county are in the format:  COUNTYNAME_COUNTY. This was causing problems when the county name was two words, e.g. WEST_SUSSEX_COUNTY and EAST_SUSSEX_COUNTY. `normalize_name_for_matching` uses `slugify` which doesn't replace underscores, but the slugs for east and west sussex in our db are separated by hypens (e.g. `east-sussex`), so they weren't matching the results of normalize until I replaced the underscores with hypens.

every_election/apps/organisations/boundaries/boundaryline.py
```diff
@@ -57,6 +57,23 @@ def get_feature_by_field(self, fieldname, code):
...
+            if (feat_name, feat_county) == tuple:
+                matches.append(feat)
+
+        if len(matches) == 0:
+            raise ObjectDoesNotExist(f"Expected one match for {tuple}, found 0")
+        if len(matches) == 1:
+            return matches[0].geom.geos
+
```
awdem, line 74: This function is similar to the one above it (`get_feature_by_field`), which has a long explanatory comment. I chose not to copy the comment because of it felt uneccessary to include the same explanation twice, but I'm not sure if that was the right thing to do.

every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
```diff
@@ -70,9 +70,8 @@ def validate_identifier(self, identifier):
         code_type, code = split_code(identifier)
 
         if code_type == "unit_id" and re.match(r"^\d+$", code):
-            # FIXME before 2021
```
awdem, line 73: so hopeful

every_election/apps/organisations/boundaries/helpers.py
chris48s: This is a really minor thing, but just for consistency with the other cases here, I'd probably do this the other way round: Replace `"_"` with `" "` on the way in, and then test for `.endswith("-county")` here.

every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
chris48s: Oh Chris-from-the-past. You poor sweet child :smile: https://github.com/DemocracyClub/EveryElection/commit/2e4f560c502e9a2843e94fb585606ed2ff8912fe

every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
```diff
@@ -70,9 +70,8 @@ def validate_identifier(self, identifier):
...
             error = (
-                "Importing boundaries from BoundaryLine against CEDs is "
+                "Importing boundaries from BoundaryLine against unit_ids is "
                 "not yet available because unit_id is not a stable identifier"
```
chris48s, line 75: A few minor bits here.
1. Lets remove the word "yet" here. I don't think CED codes will ever be in BoundaryLine at this point
2. A few lines down, can we make the final exception at the end of the function `raise ValueError("Unknown code type. Expected 'gss:X01000001'")`. A unit_id is not an expected/valid input any more.
3. In the help for the input parameter, can you change it to `help="code in the form gss:X01000001"` (i.e: remove the unit_id example)

every_election/apps/organisations/boundaries/boundaryline.py
```diff
@@ -57,6 +57,23 @@ def get_feature_by_field(self, fieldname, code):
         """
         return self.merge_features(matches)
 
+    def get_feature_by_name_and_county(self, tuple):
```
chris48s, line 60: Just taking a parameter called "tuple" here doesn't really tell me how to call this function when I read this code.

A tuple of what? I'd rather we take params `ced_name` and `county_name` here or `division_slug` and `organisation_slug` or whatever. Then I can read the function signature and know what I am supposed to pass it.

every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
```diff
@@ -178,6 +177,21 @@ def import_div_geography(self, div):
             dg.save()
         self.stdout.write("..saved {}".format(str(div)))
 
+    def match_code_to_geom(self, div, bl):
+        code_type, code = split_code(div.official_identifier)
+        fieldname = "code" if code_type == "gss" else code_type
```
chris48s, line 182: This line here `fieldname = "code" if code_type == "gss" else code_type` is to handle the fact we might have a GSS code or a unit_id. But now we only accept a GSS as a valid identifier, we can just hard-code the field. So you should be able to make this function

```py
_, code = split_code(div.official_identifier)
return self.get_geography_from_feature(
    bl.get_feature_by_field("code", code)
)
```

every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
```diff
@@ -161,11 +160,11 @@ def import_org_geography(self, org_geo):
...
-        code_type, code = split_code(div.official_identifier)
-        fieldname = "code" if code_type == "gss" else code_type
-        geom = self.get_geography_from_feature(
-            bl.get_feature_by_field(fieldname, code)
-        )
+
+        if div.division_type == "CED":
+            geom = self.match_ced_to_geom(div, bl)
+        else:
+            geom = self.match_code_to_geom(div, bl)
```
chris48s, line 167: Today on naming things is hard.. these functions return a geometry, so I think I'd be inclined to call them:

- `get_geom_by_gss` and
- `get_geom_for_ced`

every_election/apps/organisations/boundaries/boundaryline.py
```diff
@@ -57,6 +57,23 @@ def get_feature_by_field(self, fieldname, code):
...
+        if len(matches) == 0:
+            raise ObjectDoesNotExist(f"Expected one match for {tuple}, found 0")
+        if len(matches) == 1:
+            return matches[0].geom.geos
+
+        return self.merge_features(matches)
+
     def get_code_from_feature(self, feature):
         if feature.get("area_code") == "CED":
             return "unit_id:" + str(feature.get("unit_id"))
```
chris48s, line 79: Lets not do anything about it in this PR, but while I'm seeing this code.. lets :round_pushpin: put a pin it it to come back to when we look at https://app.asana.com/0/1209090479296269/1209157822204377/f

awdem: I've started working on that ticket here: #2400. But wasn't sure what to do about these lines.

chris48s: I think we can just remove them and let it throw (so we follow the `if feature.get("code") == "999999999":` branch)

The only place we call this is from `boundaryline_backport_codes` (which isn't the right command to use for CEDs). So if we're trying to use that command to attach codes to CEDs then its good if we throw here.

every_election/apps/organisations/boundaries/boundaryline.py
awdem: Done in f6d1ff96ff8682f4a3733fb827f1381a18f8656f

every_election/apps/organisations/boundaries/boundaryline.py
awdem: Makes sense. Done in 076c6d059d37be25e85e437a5948054ec1580da3

every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
awdem: Done in d4ab62604d3b6fac8c9af28fc48f794042a322b6.

every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
awdem: Done in d40ffd7b0a6ab83abe6cf69458c7408740c5054e

every_election/apps/organisations/boundaries/management/commands/boundaryline_import_boundaries.py
awdem: Done in a4d21a49bc0d90fdbf6659d32db0dc61fa4b24ae

chris48s (approved):
Latest changes look good :+1:  Will leave it with you to squash the fixups + merge

PR #2401, opened 2025-01-30
Limit the number of results it's possible to return in API requests

symroe:
This prevents someone calling ?limit=100000 and DoSing the server

### DIFF
--- a/every_election/apps/core/helpers.py
+++ b/every_election/apps/core/helpers.py
@@ -1,4 +1,6 @@
+from django.conf import settings
 from django.contrib.auth.models import Group, User
+from rest_framework.pagination import LimitOffsetPagination
 
 
 def user_is_moderator(user: User):
@@ -6,3 +8,7 @@ def user_is_moderator(user: User):
         return False
     group = Group.objects.get(name="moderators")
     return group in user.groups.all()
+
+
+class MaxSizeLimitOffsetPagination(LimitOffsetPagination):
+    max_limit = getattr(settings, "API_MAX_LIMIT", 100)
--- a/every_election/settings/base.py
+++ b/every_election/settings/base.py
@@ -261,7 +261,7 @@ LOGIN_REDIRECT_URL = "home"
 LOGOUT_REDIRECT_URL = "home"
 
 REST_FRAMEWORK = {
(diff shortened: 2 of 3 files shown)


every_election/settings/base.py
```diff
@@ -261,7 +261,7 @@ def get_ec2_ip():
 LOGOUT_REDIRECT_URL = "home"
 
 REST_FRAMEWORK = {
-    "DEFAULT_PAGINATION_CLASS": "rest_framework.pagination.LimitOffsetPagination",
+    "DEFAULT_PAGINATION_CLASS": "core.helpers.MaxSizeLimitOffsetPagination",
     "PAGE_SIZE": 100,
```
chris48s, line 265: does this do anything if we're setting `page_size` on the `MaxSizeLimitOffsetPagination` class?

every_election/apps/core/helpers.py
```diff
@@ -1,8 +1,15 @@
...
 def user_is_moderator(user: User):
     if not user.is_authenticated:
         return False
     group = Group.objects.get(name="moderators")
     return group in user.groups.all()
+
+
+class MaxSizeLimitOffsetPagination(LimitOffsetPagination):
+    page_size = 100
+    max_limit = getattr(settings, "API_MAX_LIMIT", 100)
```
chris48s, line 15: Given that if we want to change this value, we have to edit code and `API_MAX_LIMIT` is only used in one place, would it be better to just hard-code the value here? Same as page_size.

every_election/settings/base.py
symroe: Removed it from the class, so this works as documented in DRF

every_election/apps/core/helpers.py
chris48s: If we're going back to setting the default page size in settings now, lets keep it there for locality of behaviour.

chris48s (approved):

symroe:
Note that this isn't going to update the onboard instances on WDIV until we do an image update (or give up on onboard installs). This isn't a problem in reality as WDIV is the only user of those installs, and the devs.DC API doesn't proxy through WDIV.

PR #2405, opened 2025-02-04
remove duplicated org problems in admin

awdem:
remove duplicated org problems in admin by add `.distinct()` to the end of the query.


chris48s (approved):

chris48s:
Just having a look at the build failure.. this failed because I set
`DJANGO_SETTINGS_MODULE: every_election.settings.deploy` on the dev environment to test
https://github.com/DemocracyClub/EveryElection/pull/2398
I've deleted that env var for now as it will also block the prod deploy, but now I've done that you should be fine to merge this :+1:

PR #2407, opened 2025-02-05
add bulk cancel command

chris48s:
This command allows us to call something like

```
python manage.py bulk_cancel local.essex.2025-05-01
```

to cancel a group and all of its child ballots.

We can also optionally pass a `metadata-id` param

```
python manage.py bulk_cancel local.essex.2025-05-01 --metadata-id 50
```

This allows us to efficiently process all the cancellations we need to do today. Ideally I'd have done this via the admin, but a management command was quick and we need it this afternoon.

### DIFF
--- a/every_election/apps/elections/management/commands/bulk_cancel.py
+++ b/every_election/apps/elections/management/commands/bulk_cancel.py
@@ -1,0 +1,45 @@
+from django.core.management import BaseCommand
+from elections.models import Election, MetaData
+
+
+class Command(BaseCommand):
+    help = """
+    This command cancels a given election and any descendents.
+
+    Example usage:
+    python manage.py bulk_cancel local.essex.2025-05-01
+    python manage.py bulk_cancel local.essex.2025-05-01 --metadata-id 50
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "election",
+            type=str,
+            help="Election to cancel. Specifying an election group or organisation group will cancel all child ballots",
+        )
+        parser.add_argument(
+            "--metadata-id",
+            action="store",
+            type=int,
+            help="ID of a metadata object to attach to the elections we are cancelling",
+            required=False,
+        )
+
+    def handle(self, *args, **options):
+        election_id = options["election"]
+
+        elections = (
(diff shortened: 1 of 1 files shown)


symroe:
Can you ensure that the parent is also saved? Like, get the parent first, then get the descendents, loop over them as you currently do and save the parent at the end if it all? Or is the parent saved automatically already?

every_election/apps/elections/management/commands/bulk_cancel.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+        self.stdout.write(f"updating {elections.count()} elections..")
+        for election in elections:
+            self.stdout.write(f"updating {election.election_id}..")
+            election.cancelled = True
+            if options.get("metadata_id"):
+                election.metadata = MetaData.objects.get(
+                    pk=options["metadata_id"]
+                )
+            election.save()
```
awdem, line 44: Since it's a bulk command, would it safer do all the saving at the end as a transaction?

https://github.com/DemocracyClub/EveryElection/blob/86788ee49bc974a9c3e1f14eda9a17c1a89e7002/every_election/apps/elections/management/commands/assign_election_to_divisionset.py#L65-L68

awdem (commented):
Have tested locally and it is working - just wanted to check - should we feed in a cancellation reason as well or is that not necessary (for this afternoon, anyway)?

every_election/apps/elections/management/commands/bulk_cancel.py
chris48s: done in https://github.com/DemocracyClub/EveryElection/pull/2407/commits/fbb030aa65e4be254ba4920c674abdbd2fe64d60

chris48s:
The elections we're dealing with here don't fit into any of the standard cancellation reasons so for these ones we want `cancellation_reason` to be null. I think I'm going to leave this one out given it is a quick sticking plaster for a feature that should ideally exist in the admin

awdem (approved):
Looks good!

chris48s:
@symroe `.get_descendents(inclusive=True)` means we get the parent object in the qeuryset as well as the children

PR #2417, opened 2025-02-18
Check if identified type passed before prefetching

symroe:
This saves a query when requesting a single ballot


chris48s:
quick ruff fix needed (unused var), but aside from that LGTM

chris48s (approved):

PR #2433, opened 2025-03-04
serve a Last-Modified header on election detail endpoint

chris48s:
Refs
https://app.asana.com/0/1204880927741389/1209455180573671
https://app.asana.com/0/1204880927741389/1209455180573673

In this PR, I have worked on an improved version of the quick code in
https://app.asana.com/0/0/1209455180573673/1209508193329969/f
It is the same concept, but I've fixed some things.

A couple of things I have learned from testing this on CloudFront:

We don't need to change any CloudFront config to tell it to send a `If-Modified-Since request` header. It will send one if the cached object had a `Last-Modified` header on it, so sending that header on responses (which happens if `single_ballot_cache()` returns anything other than `None`) automatically configures the behaviour we want from the CDN.
It is also worth noting that browsers also do this.

If the origin returns a 304, we get a `x-cache: RefreshHit from cloudfront` header on the response instead of a `x-cache: Hit from cloudfront` which is useful for debugging purposes.

### DIFF
--- a/every_election/apps/api/tests/test_api_election_endpoint.py
+++ b/every_election/apps/api/tests/test_api_election_endpoint.py
@@ -214,7 +214,7 @@ class TestElectionAPIQueries(APITestCase):
 
     def test_detail_num_queries(self):
         id_ = ElectionWithStatusFactory(group=None).election_id
-        with self.assertNumQueries(2):
+        with self.assertNumQueries(3):
             self.client.get(f"/api/elections/{id_}/")
 
     def test_identifier_type_filter(self):
--- a/every_election/apps/api/views.py
+++ b/every_election/apps/api/views.py
@@ -1,9 +1,14 @@
 from collections import OrderedDict
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from api import filters
+from django.conf import settings
 from django.db.models import Prefetch
 from django.http import Http404
+from django.utils import timezone
+from django.utils.decorators import method_decorator
+from django.utils.http import parse_http_date
+from django.views.decorators.http import condition
 from elections.models import (
     Election,
     ElectionSubType,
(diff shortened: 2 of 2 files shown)


every_election/apps/api/tests/test_api_election_endpoint.py
```diff
@@ -214,7 +214,7 @@ def test_detail_invalid_id(self):
 
     def test_detail_num_queries(self):
         id_ = ElectionWithStatusFactory(group=None).election_id
-        with self.assertNumQueries(2):
+        with self.assertNumQueries(3):
```
chris48s, line 217: See the commit message for
https://github.com/DemocracyClub/EveryElection/pull/2433/commits/e99ff9357e170e161e5ef7174d75a92ac1040aca

The approach of putting a `@method_decorator(condition(last_modified_func=some_function))` decorator on `retrieve()` adds another round-trip to the DB in the cache miss case
because we have to get the Election object once in the function which gets the modified date, but then having got that out of the DB we just to have to return a last modified date from that function. We can't pass the election object on so we have to fetch it from the DB again in the `retrieve()` function. Even if we could it would kind of negate the point because we would have to do the "expensive" version of the query up-front to avoid more round-trips.

I wonder if the juice is worth the squeeze here?

every_election/apps/api/views.py
```diff
@@ -47,6 +52,42 @@ class APIInvalidElectionIdException(APIException):
...
+        return None
+
+    if request.query_params.get("deleted", None) is not None:
+        qs = Election.private_objects.all()
+    else:
+        qs = Election.public_objects.all()
+
+    try:
+        election = qs.only("modified").get(election_id=kwargs["election_id"])
+        actual_modified = election.modified.replace(microsecond=0)
```
chris48s, line 66: There was an off-by-one in your cache invalidation code :D

every_election/apps/api/views.py
```diff
@@ -47,6 +52,42 @@ class APIInvalidElectionIdException(APIException):
...
+        )
+        if client_timestamp is None:
+            return actual_modified
+
+        client_dt = datetime.fromtimestamp(client_timestamp, tz=timezone.utc)
+
+        if actual_modified > client_dt:
+            return actual_modified
+
+        if timezone.now() - client_dt > timedelta(hours=4):
```
chris48s, line 85: I made this a seperate PR, but this is intended to pair with the code in https://github.com/DemocracyClub/EveryElection/pull/2432

Again, I decided to just hard-code the value here for the same reason as https://github.com/DemocracyClub/EveryElection/pull/2432#discussion_r1979410607

Combined these 2 PRs will cache an object at the CDN for 2 hours but only force a refresh every 4. I'm not sure those are the right numbers tbh. Lets review these next week.

chris48s:
:warning: Also worth noting that this probably isn't sensible to actually merge without doing some work on updating the `Election.modified` property in response to events, which I have not done here.

PR #2438, opened 2025-03-19
Fix sync_map_layers_to_s3.sh path

GeoWill:
The old `ee-maps` bucket is in the old AWS account. 

This one is in the new account, so probably makes sense to use it.


symroe (approved):

symroe:
Is this linked to anywhere?

PR #2440, opened 2025-03-31
WIP Re-trigger pipeline on election approved/soft deleted events

chris48s:
WIP


every_election/apps/elections/baker.py
```diff
@@ -0,0 +1,2 @@
+def push_event_to_queue():
+    print("trigger_rebuild")
```
chris48s, line 2: I think there are 4 things to consider when moving on to this:

1. TODO: actually write the code
2. Permissions: We need to assign permissions both on the EE side and the data-baker side to allow everything to talk to each other
3. There are some places where we want this function to be a no-op. For example in local dev and when running unit tests. I also would want this to be a no-op for an embedded install (accepting that we want to get rid of those).
5. We want to push to different queues in different environments. Although we want dev/staging WDIV to be consuming the same ballot data as prod WDIV, if I create an election in EE staging/dev I don't want to push an event onto the production queue. Nothing in the prod DB changed.

every_election/apps/elections/models.py
```diff
@@ -531,16 +532,16 @@ def clean(self):
             )
 
     @transaction.atomic
-    def save(self, *args, **kwargs):
+    def save(
+        self, *, push_event=True, status=None, user=None, notes="", **kwargs
```
chris48s, line 536: The way I've done this (both here and `ModerationHistory.save()`) is I've said the default is we push an event onto the queue and we have to opt-in to disabling it.

The upside of this is it means we are less likely to do some ad-hoc thing like a management command and forget about pushing an event onto the queue.

The downside of this is we are more likely to accidentally fill the queue with lots of events when we only wanted one if we forget to disable it.

I think we agreed this was the right way round though, and if we want to move to pushing ballot-level events to the queue on every create/delete then this is going to be the right way round.

every_election/apps/elections/sync_helper.py
```diff
@@ -196,7 +196,9 @@ def add_single_election(self, result: dict):
             setattr(election_model, key, value)
 
         election_model.modified = result["modified"]
-        election_model.save(update_modified=False, status="Approved")
+        election_model.save(
+            push_event=False, update_modified=False, status="Approved"
```
chris48s, line 200: I think this would only ever be used on "client" installs, so would not make sense to push an event in this context

every_election/apps/elections/tests/test_election_model.py
```diff
@@ -263,6 +264,34 @@ def test_requires_voter_id_empty(self):
...
         assert self.ballot.requires_voter_id is None
 
+    def test_ballot_push_event_true(self):
+        test_cases = [
+            [ModerationStatuses.approved.value, 1],
+            [ModerationStatuses.deleted.value, 1],
+            [ModerationStatuses.suggested.value, 0],
+            [ModerationStatuses.deleted.value, 0],
+        ]
+        for status, expected_calls in test_cases:
```
chris48s, line 274: I would have normally liked to use `@pytest.mark.parametrize` here, but it is not compatible with a UnitTest class
I considered splitting it out, but I wanted to use the data we are setting up in `setUp`
The path of least resistance here was to just use a for loop :/

every_election/apps/elections/tests/test_id_creator_wizard.py
```diff
@@ -102,36 +105,118 @@ def test_date_warning_shown(page, live_server, settings):
...
+    user.groups.add(group)
+    user.save()
+
+    # Log in
+    page.goto(f"{live_server.url}/admin/")
+    page.locator("#id_username").fill("fred")
+    page.locator("#id_password").fill("password")
+    page.locator("input[type=submit]").click()
+    # Ensure login worked correctly
+    expect(page).to_have_title("Site administration | Django site admin")
```
chris48s, line 187: Crude but effective.
Any thoughts on a slightly better proxy for "assert the login actually worked" in the context of a playwright test?

every_election/apps/elections/models.py
```diff
@@ -593,20 +600,34 @@ class ModerationHistory(TimeStampedModel):
...
     notes = models.CharField(blank=True, max_length=255)
 
-    def save(self, **kwargs):
-        # if this is the initial status no need to update the related election
-        # so return early. This is because the default status is identical on
-        # both this model and the Election model
-        if kwargs.pop("initial_status", False):
-            return super().save(**kwargs)
+    @transaction.atomic()
+    def save(self, *, push_event=True, initial_status=False, **kwargs):
```
chris48s, line 604: This isn't a new problem but conceptually this should be in a transaction. If `self.election.save()` fails, I also want to roll back the `super().save()`.
but doing so does add a few more round-trips to the DB to start/end. Note the `assertNumQueries()` changes in 29d2125

every_election/apps/elections/models.py
```diff
@@ -531,16 +532,16 @@ def clean(self):
...
         if self.requires_voter_id == "":
             self.requires_voter_id = None
 
         # used later to determine if we should look for ballots
         created = not self.pk
 
-        status = kwargs.pop("status", None)
-        user = kwargs.pop("user", None)
-        notes = kwargs.pop("notes", "")[:255]
+        notes = notes[:255]
```
chris48s, line 544: A little bit of leaving things better than I found them here in bef0228

Django's `Model.save()` takes no positional arguments
https://github.com/django/django/blob/35c58a7924aa07e87e925f6eea978c278f143f36/django/db/models/base.py#L767-L769
and in Django 6 support for positional arguments to `Model.save()` will be completely removed. See
https://docs.djangoproject.com/en/dev/releases/6.0/#:~:text=Support%20for%20passing%20positional%20arguments%20to%20Model.save()%20and%20Model.asave()%20is%20removed
https://github.com/django/django/blob/a0fb35eb726f1a04eaa1b47b8de191fafe55a0ab/django/db/models/base.py#L806-L813
So moving these to keyword-only arguments only is a safe thing to do here (and desirable for forwards-compatibility).

Moving these from the `**kwargs` dict to named arguments makes it clearer what params this function takes and gives you better editor hinting and stuff.

every_election/apps/elections/tests/test_id_creator_wizard.py
GeoWill: You could add the `browser` fixture to the test, and then inspect `browser.contexts[0].cookies()` and look for if `sessionid` has been set. However I don't think it really gains anything over what you're doing. 

Interesting aside I just logged into find out what the cookie was called, and the other difference is a successful login is a 302 and unsuccessful 200. So I guess you could capture the response and check the status too.

chris48s: I like your suggestion of checking the session cookie better. It feels like it is based on an assumption that is more likely to hold across future django upgrades. I've implemented that suggestion in https://github.com/DemocracyClub/EveryElection/pull/2440/commits/d30d312c356488f2931f260d88bbede048cb51ac

symroe:
This looks mostly good to go, with two comments:

1. I'm going to continue to bike shed the name I'm afraid. If the function is going to be generically 'push a thing to a queue' then I think it needs a payload that tells the queue what sort of thing it is. I think, however, this would be better implemented as `push_event_change_notification` (or something) that makes it clear what's going on. e.g there are two patterns: 1. generic name with specified payload `push_event(events.ElectionChanged)` or specific name `push_event_change_notification`. We have a generic name at the moment with little indication of when we should call it.
2. As a feature request: I wonder if we should have a global feature flag to stop events being pushed ever. This is to catch the situation where client installs end up saving events in ways we miss here. We could disable pushing events in client install, and maybe dev installs? Deff local development installs.

chris48s:
1. I think keeping it generic and adding params is best. In https://github.com/DemocracyClub/EveryElection/pull/2445 where we flesh out the function body, @geowill changes the function signature to

```py
def push_event_to_queue(
    message_body: Optional[str] = None, message_group_id: Optional[str] = None
):
    ...
```

but in that PR we're allowing them to be nullable. How I add the params in this PR but make them required, so the signature is

```py
def push_event_to_queue(message_body: str, message_group_id: str):
    ...
```

and then I explicitly pass in a `message_body` and `message_group_id` at each call site?

2. You mean like this, right? https://github.com/DemocracyClub/EveryElection/pull/2445#discussion_r2030976511

chris48s:
Do you reckon it makes most sense to make the `message_body` and `message_group_id` params required or keep them optional?

symroe:
Yeah, I don't think calling "push an event" with no data makes sense to me

chris48s:
OK. Lets make them required :+1: I'll quickly follow up with Will in standup tomorrow on who is going to do that on which branch.

PR #2445, opened 2025-04-03
Pipeline trigger triggering

GeoWill:
this is based off of #2440.

Maybe I should just push to that branch...

This gives the EC2 instances permission to push to the `CurrentElectionsEventQueue`
All three environments need a parameter adding called `CURRENT_ELECTIONS_PARQUET_EVENT_QUEUE_URL` that corresponds to the corresponding DataBaker stack. i.e. the dev ee account should have the url for the dev data baker account.


every_election/apps/elections/baker.py
```diff
@@ -0,0 +1,52 @@
...
+from typing import Optional
+
+import boto3
+
+logger = logging.getLogger(__name__)
+
+
+CURRENT_ELECTIONS_PARQUET_EVENT_QUEUE_URL = os.environ.get(
+    "CURRENT_ELECTIONS_PARQUET_EVENT_QUEUE_URL"
+)
```
chris48s, line 12: I think we should read this in via django settings instead of direct from the env here. Also, when we do that we should throw a `KeyError` error if `DC_ENVIRONMENT` is one of `["development", "staging", "production"]` and `CURRENT_ELECTIONS_PARQUET_EVENT_QUEUE_URL` is not set. Lets not silently pass it.

every_election/apps/elections/baker.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def push_event_to_queue(
+    message_body: Optional[str] = None, message_group_id: Optional[str] = None
+):
```
chris48s, line 17: By default this function should be a no-op under test and in local dev unless we're explicitly enabling it. I reckon we should set a `SEND_EVENTS` flags (or `PUSH_MESSAGES` or something - naming things is hard) in settings which defaults to

```py
SEND_EVENTS = settings.DC_ENVIRONMENT in ["development", "staging", "production"]
```

but we can override it in `local.py` (for local dev) or with `@override_settings` under test

..and then this function can be

```py
def push_event_to_queue(message_body, message_group_id):
    if not setings.SEND_EVENTS:
        # don't attempt to push anything in local dev/under test
        return

    # do the things...
```

chris48s: We'd also want to make sure this setting is off for client/embedded installs, while those still exist.

every_election/apps/elections/baker.py
GeoWill: Agree we should error if `DC_ENVIRONMENT` is set and  `CURRENT_ELECTIONS_PARQUET_EVENT_QUEUE_URL` isn't. The reason I was getting it directly from the env, is because I assumed it would be stored in parameter store, the contents of which get written to the env.

chris48s: It will be stored in parameter store, but as with other settings we read from the env, lets read then into django settings and then access them from `django.conf.settings` when we need them.

One benefit of this over accessing the env at runtime is we can perform all our consistency checks on the settings we read from the env at the point we bootstrap the app. Then if the settings are bad, we fail the deploy.

every_election/apps/elections/baker.py
```diff
@@ -1,2 +1,52 @@
...
+
+
+def send_event(detail: Dict, detail_type: str, source: Optional[str] = None):
+    if not settings.SEND_EVENTS:
+        logger.info(f"Skipping {detail_type} because SEND_EVENTS is disabled.")
+        # don't attempt to push anything in local dev/under test
+        return
+
+    if source is None:
+        source = f"everyelection-{settings.DC_ENVIRONMENT}-{settings.EC2_IP}"
```
chris48s, line 19: You're using `settings.DC_ENVIRONMENT` here, but I think we never actually assign this a value in settings. It is always referenced as `os.environ.get("DC_ENVIRONMENT")` in `base.py`. The tests pass because you're explicitly mocking it.

Lets set

```
DC_ENVIRONMENT = os.environ.get("DC_ENVIRONMENT")
```

or

```
DC_ENVIRONMENT = os.environ.get("DC_ENVIRONMENT", "development")
```

(if that doesn't cause other problems)

early doors in `base.py` and then reference it with `settings.DC_ENVIRONMENT` from the frowards in the config file and elsewhere.

every_election/apps/elections/baker.py
```diff
@@ -1,2 +1,52 @@
...
+                    "Detail": json.dumps(detail),
+                    "EventBusName": settings.DC_EVENTBUS_ARN,
+                }
+            ]
+        )
+        if response["FailedEntryCount"] > 0:
+            logger.error(
+                "Event was unsuccessfully ingested.",
+                extra={"response": response},
+            )
```
chris48s, line 41: What does this mean if the event was "unsuccessfully ingested"? Did we ingest it, or not?
If not, I suggest we say "Event was not ingested." or similar. Lets make it really clear.

---


Side issue: We don't need to deal with this right now, but if we become more dependent on events for propogating changes elsewhere then we probably need to either:

- Have a way store events in EE so we can re-send the event if bad things happen here (or in the the `InternalException` case).
- If we can't push the event to the queue, throw so we rollback the transaction (i.e: we prevent the user from making an edit if we can't also propagate the edit to other systems)

Lets not get bogged down in this issue now though.

GeoWill: Yep right now I've taken the approach that it's better to throw an error at sentry, but not block whatever was happening which meant an event was sent. Hopefully the error in [3b2b49f](https://github.com/DemocracyClub/EveryElection/pull/2445/commits/3b2b49f31f99018157dd3a443999a84d7f60d8a5) is clearer. I think 'ingested' came from reading some (tangentially related) docs about cloudwatch logs when I was doing this.

every_election/apps/elections/baker.py
GeoWill: [86a978a](https://github.com/DemocracyClub/EveryElection/pull/2445/commits/86a978a84a3cfe672eed5262868ef1c53616d1b7)

chris48s:
Cool. I've had a look over the latest changes and this looks good.
This PR is still marked as draft. Do you have additional changes you're hoping to make, or is it really finished?
Also, I think it would probably be useful for @symroe to have a look over this one - mainly for oversight of what is changing.

every_election/apps/elections/baker.py
```diff
@@ -1,2 +1,89 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def event_bus_exists(event_bus_arn: str) -> bool:
+    """
+    Check if the specified event bus ARN exists.
+    """
+    sts_client = boto3.client("sts")
+    role_arn = settings.DC_CHECK_EVENTBUS_ROLE
```
chris48s, line 17: Can you give a quick explanation of why we need a different role for this? i.e: Why do we need to create a new role, rather than assigning the existing role that the EC2 instances run with the ability to perform this action?

I'm not saying this is wrong. I just don't know.

GeoWill: This is a role in the account where the event bus lives. That role will trust any account in the org and allow it to assume it via `sts.AssumeRole`. This is necessary because we can't give cross account permissions to list event buses.

every_election/apps/elections/baker.py
```diff
@@ -1,2 +1,89 @@
...
+import os
+from typing import Dict, Optional
+
+import boto3
+from django.conf import settings
+
+logger = logging.getLogger(__name__)
+
+
+def event_bus_exists(event_bus_arn: str) -> bool:
```
GeoWill, line 12: This is all to guard against us typo-ing the evenbus we're trying to send events to. We check it exists before we send an event. This is because events will fail silently if an event bus doesn't exist.

I don't really like this solution, because it feels quite brittle, having the specific `DC_CHECK_EVENTBUS_ROLE` in the account with the eventbus we're targetting, but I haven't come up with anything better. The alternative would be just to delete this check.

every_election/apps/elections/tests/test_send_event.py
```diff
@@ -0,0 +1,320 @@
...
+from elections.baker import event_bus_exists, send_event
+from moto import mock_aws
+
+DEVS_DC_API_ACCOUNT_ID = "111111111111"
+EE_ACCOUNT_ID = "123456789012"  # This is the moto default
+
+
+@pytest.fixture
+def devs_dc_moto_account_env(monkeypatch):
+    monkeypatch.setenv("MOTO_ACCOUNT_ID", DEVS_DC_API_ACCOUNT_ID)
```
GeoWill, line 18: I think this is all for nothing in the end. This is because afaict moto just gives you all the permissions when you assume a role in another account. So it's definitely not testing the org trust relationship. I've pushed it up, because it might provoke some ideas, but I think it also over complicates the tests. OTOH it does make it clear we expect the target event bus to be in a different account, so that's something. 

Relevant moto docs: 
This is what I'm using: https://docs.getmoto.org/en/latest/docs/multi_account.html
I haven't tried to implement this: https://docs.getmoto.org/en/latest/docs/iam.html

every_election/apps/elections/baker.py
GeoWill: Here's the code that makes the corresponding role: https://github.com/DemocracyClub/dc-data-baker/pull/24/commits/5872895bb68401acfa20660381f590231c5bbaeb

chris48s: Got it. Ta

every_election/settings/base.py
```diff
@@ -278,6 +285,15 @@ def get_ec2_ip():
 CORS_ALLOW_METHODS = ("GET", "OPTIONS")
 
 UPSTREAM_SYNC_URL = "https://elections.democracyclub.org.uk/api/elections/"
+
+# DC Eventbus.
+SEND_EVENTS = False
+if DC_ENVIRONMENT in ("development", "staging", "production"):
+    # If we've set DC_ENVIRONMENT then we should raise a KeyError if queue url isn't set
```
chris48s, line 292: this doesn't seem like the right comment any more

every_election/apps/elections/baker.py
chris48s: ..and then, assuming the issue is that pushing an event fails silently (I assume that must be the issue, otherwise we'd just catch the error) is there any other operation we can try that would fail. For example, could we try a `describe_event_bus` before we `put_events`? Or does that also require a assuming a different Iam role? We're assigning `DescribeEventBus` in https://github.com/DemocracyClub/EveryElection/blob/a2c4b9b7a4427d1fdde66e8ae412f190a5552b57/cdk_stacks/stacks/code_deploy_policies.py#L109-L119

chris48s: I've just realised the last commit you pushed was switching from `list_event_buses` to `describe_event_bus` :facepalm:

chris48s: Yeah no worries.

Just had a quick test with `put_events` and the value of `Entries` doesn't contain anything helpful if the ARN was bad, which is quite annoying. So its basically we do this check, or we don't.

I think I'd be fine with saying it is on us to set the ARN correctly just to simplify the code here. Might be a good one to get a second opinion from @symroe on.

I think if we _are_ going to do this check, we should probably move it though. I reckon we should check the bus exists one on app startup in a system check, rather than every time we push an event to it (because of performance impact).

Lets decide if we actually want to do it first though.

GeoWill: I don't think the response does help us. 
This is the response to an event bus that doesn't exist: 
```
{'FailedEntryCount': 0,
 'Entries': [{'EventId': '0dcb7c51-2c36-87ee-39e2-fa56a0855f41'}],
 'ResponseMetadata': {'RequestId': '9a1f38a7-16ed-443e-a81f-e093e1118b21',
  'HTTPStatusCode': 200,
  'HTTPHeaders': {'x-amzn-requestid': '9a1f38a7-16ed-443e-a81f-e093e1118b21',
   'content-type': 'application/x-amz-json-1.1',
   'content-length': '85',
   'date': 'Thu, 24 Apr 2025 15:12:16 GMT'},
  'RetryAttempts': 0}}
```

Versus to one that does

```
{'FailedEntryCount': 0,
 'Entries': [{'EventId': '7ce793b7-7038-c618-52ef-b220308646ba'}],
 'ResponseMetadata': {'RequestId': '5397ef8c-492e-4233-af01-e59f6b52579d',
  'HTTPStatusCode': 200,
  'HTTPHeaders': {'x-amzn-requestid': '5397ef8c-492e-4233-af01-e59f6b52579d',
   'content-type': 'application/x-amz-json-1.1',
   'content-length': '85',
   'date': 'Thu, 24 Apr 2025 15:14:04 GMT'},
  'RetryAttempts': 0}}
```

The event that worked I can see in the console. The first just gets swallowed as best I can tell.

GeoWill: Sorry didn't see you're comment :) I was doing the same. 

Good point re. doing the check once.

symroe: I wrote the above comment before reading all of this, and now I'm making everything more confusing :laughing: Sorry!

So, broadly we agree this is a bit of a smell, and I think we might agree that it's better implemented as a one-time check.

Another approach, that we might want to consider as part of a wider "how to even event bus" conversation, is that we could host a 'gateway' lambda in the event bus hosting account. Lambda _does_ have good cross-org permissions and will fail if it doesn't exist. Calling Lambda in async mode is non-blocking, and the Lambda will deal with retrying enqueuing failures, too.  It's also a place for us to add additional logging / whatever, if that's useful. 

An added bonus to the approach is that we'd have a single gateway that all apps could send events to, and at this single point we can add some validation of the event payload.

I'm not really suggesting we do this now (I'd like to get this merged before the elections to see how it works under load), but it might save pushing this sort of logic to the clients. 

```python
lambda_client.invoke(
    FunctionName=function_arn,
    InvocationType="Event",
    Payload={},
)
```

Is a really easy thing to call, and is basically non-blocking (we use this approach for the DC Logging system, as Firehose doesn't have good cross-org permissions either)

PR #2452, opened 2025-04-24
specify min generation on mapit links if division is not current

chris48s:
Refs https://github.com/mysociety/mapit/issues/433

It used to be that you could call `/area/<gss>` with any GSS code and it would search for that GSS code, even if that GSS code was not active in the current generation.

Mapit made a change upstream which means this is no longer true. This means a lot of our mapit links for old geographies are broken.

Adding `min_generation=1` effectively restores the previous behaviour of searching across genenerations with the caveat that there may still be some GSS codes that will give a 404 because mapit has duplicates for some GSS codes. Annoyingly current Westminster Constituencies are one of those cases so we also don't want to just _always_ pass `min_generation=1`. Doing that causes other problems.

The change in this PR should fix _most_ of our links out to mapit. Maybe not 100% of them. It should also not break any that currently work. That said, I think there is a good case for just making our own maps and providing good maps for orgs/divisions/division sets like we do for election objects. We have the polygons to do it, and this would cover other cases (e.g: where we hold divisions that aren't in BoundaryLine yet). Even if we do that, linking out to mapit where we can is still a useful feature to retain IMO.

Refs https://app.asana.com/1/1204880536137786/project/1207880534533137/task/1210054691908446?focus=true

### DIFF
--- a/every_election/apps/organisations/models/organisations.py
+++ b/every_election/apps/organisations/models/organisations.py
@@ -103,7 +103,7 @@ class Organisation(UpdateElectionsTimestampedModel, DateDisplayMixin):
             return None
         if not self.geographies.latest().gss:
             return None
-        return "https://mapit.mysociety.org/area/{}".format(
+        return "https://mapit.mysociety.org/area/{}.html".format(
             self.geographies.latest().gss
         )
 
--- a/every_election/apps/organisations/models/divisions.py
+++ b/every_election/apps/organisations/models/divisions.py
@@ -162,7 +162,11 @@ class OrganisationDivision(UpdateElectionsTimestampedModel):
         if code_type.lower() != "gss":
             return None
 
-        return "https://mapit.mysociety.org/area/{}".format(code)
+        link = f"https://mapit.mysociety.org/area/{code}.html"
+        if self.divisionset.end_date:
+            link += "?min_generation=1"
+
+        return link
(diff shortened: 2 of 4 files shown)


every_election/apps/organisations/models/organisations.py
```diff
@@ -103,7 +103,7 @@ def format_geography_link(self):
             return None
         if not self.geographies.latest().gss:
             return None
-        return "https://mapit.mysociety.org/area/{}".format(
+        return "https://mapit.mysociety.org/area/{}.html".format(
```
GeoWill, line 106: Is there a reason for not including `"?min_generation=1"` here? 

Looks like we don't use it on the frontend, but if we're going to keep it around does it make sense to keep it the same as on the divisions?

GeoWill:
Question more for my own interest. I think this looks fine, I also agree with doing better org and division maps!

every_election/apps/organisations/models/organisations.py
chris48s: Yes. For the organisation geographies we're choosing the latest geography object ( `self.geographies.latest().gss` ), which should be current generation on mapit :crossed_fingers: . For divisions, we might be looking at a current division or a historic one.

GeoWill (approved):

PR #2453, opened 2025-04-29
Pipeline trigger triggering mk2

GeoWill:
Simpler version of #2445. 
This doesn't test for whether the event bus exists or not.


chris48s (approved):
:shipit:

PR #2456, opened 2025-05-14
fix debug toolbar settings

chris48s:
I messed this up in the settings refactor PR

### DIFF
--- a/every_election/settings/base.py
+++ b/every_election/settings/base.py
@@ -317,14 +317,6 @@ BASIC_AUTH_ALLOWLIST = [
 CURRENT_PAST_DAYS = 20
 CURRENT_FUTURE_DAYS = 90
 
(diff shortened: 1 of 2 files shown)


GeoWill (approved):
`runserver` now fails on a dev machine if debug toolbar isn't installed (i.e. you can't specify `DEBUG_TOOLBAR = False` anymore). However given `uv sync` installs dev dependencies automatically, I don't think it matters. If for whatever reason one wants to run without the toolbar, it's easy enough to commend out the relevant lines.

chris48s:
I think it is reasonable to say you need to have the dev packages installed for local dev (which includes pytest, ruff, etc). If you want I can put it behind an env var flag.

GeoWill:
Good as is I think. :+1:

PR #2457, opened 2025-05-14
improve performance of election types page in id creator

chris48s:
This resolves a long-standing issue that has been annoying me

Steps to reproduce:

- Go to https://elections.democracyclub.org.uk/
- Add a new election
- Put in a date
- Press Submit

The next page will take 10-15 seconds to load.

Conceptually this should not be an expensive query but for some reason postgres is making a complete hash of the query plan so I've rewritten it in a way that generates the same output but nudges postgres to make a more efficient query plan.

I also shoved an order by clause on there so they always come out in the same order.

### DIFF
--- a/every_election/apps/elections/forms.py
+++ b/every_election/apps/elections/forms.py
@@ -3,7 +3,7 @@ from typing import Union
 from dc_utils import forms as dc_forms
 from django import forms
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


GeoWill (approved):
Checked it out locally and gives ~10x speed up. Can't say I really understand why, but does what we want.

PR #2460, opened 2025-05-28
improve name mapping

awdem:
This PR makes it so that the `DivisionSetGeographyImporter` strips common suffixes off mapping file division names before trying to match them with the legislation names.

It should reduce the amount of division name mismatches we encounter when importing LGBCE boundary reviews and therefore reduce the potential for human error when generating name maps.

Tested by running through importing the Shropshire 2024 ECO locally. This was a good test because:
1) I made the North/North East mistake on this one before and now that isn't possible because those divisions no longer appear on the mismatch list.
2) You still need to go through the name_map.json generation process, but it's a much smaller list and they're all due to spelling mistakes in the mapping file.

There are likely more suffixes that will need to be added as we encounter them.

### DIFF
--- a/every_election/apps/organisations/importers.py
+++ b/every_election/apps/organisations/importers.py
@@ -71,24 +71,38 @@ class DivisionSetGeographyImporter:
         if settings.IN_TESTING:
             return None
 
-        legislation_names = set(legislation_names)
-        boundary_names = set(boundary_names)
+        # We want to remove common suffixes from the boundary names before comparing to legislation names,
+        # but we want to use the original names in the name map so we build a stripped-orig mapping
+        stripped_to_original = self.make_stripped_suffixes_map(boundary_names)
 
-        missing_from_leg = sorted(boundary_names - legislation_names)
-        map = {}
-        for name in sorted(legislation_names):
-            if name not in boundary_names:
+        # We start the name_map by adding all the names that had a suffix stripped successfully
+        name_map = {
+            orig: stripped
+            for stripped, orig in stripped_to_original.items()
+            if stripped != orig
+        }
+
+        # Then we compare the legislation names to stripped names, if any don't match we prompt the user
+        legislation_names = set(legislation_names)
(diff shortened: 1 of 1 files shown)


every_election/apps/organisations/importers.py
```diff
@@ -102,11 +116,29 @@ def make_name_map(self, legislation_names, boundary_names):
...
+        stripped_to_original = {}
+        for name in boundary_names:
+            stripped = self.remove_common_suffixes(name)
+            stripped_to_original[stripped] = name
+        return stripped_to_original
+
+    def remove_common_suffixes(self, name):
+        suffixes = [" Electoral Division", " ED", " Division", " Ward"]
+        for suffix in suffixes:
+            name = name.replace(suffix, "")
```
GeoWill, line 139: I think we should check that the suffix is a suffix. I.e. that we're not going to delete both instances of ` Ward` from `Danvers and Wardoun Ward`. Python has a [str.removesuffix](https://docs.python.org/3/library/stdtypes.html#str.removesuffix) method which I think would be more appropriate. 

Using that might make it safe to then extend suffixes as:

```
suffixes += [s.lower() for s in suffixes]
```

Which would catch the lowercase versions.

awdem: I thought you'd made up `Danvers and Wardoun` until I google it :). I've changed to using removesuffix and added lowercase versions as suggestd in [d86ce80](https://github.com/DemocracyClub/EveryElection/pull/2460/commits/d86ce8023ab3a7335f2d4eec254fde7abcb7fa49)

GeoWill: Did a quick `WHERE name like '% Ward%'` :).

awdem:
> Also have you checked that it works alright for generating a name map that isn't due to suffixes?

Just tested it for Cannock Chase 2023, which has a spelling error in one of the mapping file names but doesn't have suffixes. The command generates the same `name_map` as the existing one and imports the geographies with no issue.

PR #2465, opened 2025-06-05
don't cache moderation/review pages

chris48s:
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1210436446762455?focus=true

This fixes a long-standing bug: The lists of elections to be moderated are cached at the CDN. This means you approve one, go back to the list, and the one you just approved is still there.

In this PR I'm using the [never_cache](https://docs.djangoproject.com/en/5.2/topics/http/decorators/#django.views.decorators.cache.never_cache) decorator to send a cache-control header instructing browsers and proxies not to cache these pages. Given these are only shown to logged in users (so about 10 people) the performance impact will be negligible. I don't see a reason to do anything more complicated.

### DIFF
--- a/every_election/apps/election_snooper/views/bot_review.py
+++ b/every_election/apps/election_snooper/views/bot_review.py
@@ -5,6 +5,8 @@ from django.contrib.auth.mixins import UserPassesTestMixin
 from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from django.http import HttpResponseRedirect
 from django.urls import reverse
+from django.utils.decorators import method_decorator
+from django.views.decorators.cache import never_cache
 from django.views.generic import TemplateView
 from election_snooper.forms import ReviewElectionForm
 from election_snooper.models import SnoopedElection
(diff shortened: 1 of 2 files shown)


GeoWill (approved):
This looks fine, I don't know for sure whether it will work, and would need to deploy to try it out.
The rest of the caching behaviour is defined here: https://github.com/DemocracyClub/EveryElection/blob/d05372041659a958baf5361d5aa2342c7e97151c/cdk_stacks/stacks/code_deploy.py#L314-L389
The default header behaviour includes `Cache-Control` in the allow list, so I guess that means it will pay attention when `@never_cache` alters it. I don't know this though, so worth checking it's worked. If it doesn't work, then setting a specific cache policy in the cdk above is probably the way to go.

chris48s:
Yeah I have not tested it behind a CDN but I am pretty sure it will respect the cache-control headers because the /admin UI, uses `never_cache` and we haven't had to set any custom behaviour for that at the CDN level.

We've got some backlog of suggested elections I need to review so I will deploy this and double-check it.

Issue #2179, opened 2024-06-23
Add and display reason for by-elections being called.


pmk01:
This would work a bit like cancellation reason, I think. When creating the by-election, choose from a selection of options (resignation, death, disqualification I think are the three). We can then display this in places/the API.

pmk01:
Option to include a source.

PR #2477, opened 2025-06-20
Byelection reasons

GeoWill:
Addresses #2179

This adds a by election reason to Election model, admin and elections endpoint.

This won't be exposed in the devs.dc api, so it won't be in WCIVF, and therefore won't be in the ballot cache. Figuring out how we want to string that all together should probably be outside this.


every_election/apps/elections/models.py
```diff
@@ -577,6 +577,10 @@ def clean(self):
             raise ValidationError(
                 "Only a cancelled election can have a cancellation reason"
             )
+        if ".by." not in self.election_id and self.by_election_reason:
+            raise ValidationError(
+                "Only a by election can have a by_election_reason"
+            )
```
GeoWill, line 583: The `clean` method doesn't get called on `save`, rather when used through a form. Given I've excluded `by_election_reason` field from the admin form I don't think there's any forms where this check will be hit. So maybe this is a bit unnecessary.

GeoWill:
I think adding it to the id creator wizard is a good idea. For me at least it felt like it was going to be at least as much work again. In particular getting it to show only for by-elections seemed like we'd probably need to use js, which isn't a stopper, just makes it all a bit more complicated.
Happy to have a bash if you think it's worth it though.

GeoWill:
I'm not sure I understand:  "add a filter to the admin to show all by-election reasons, including `None`". Do you mean all by-elections grouped by reason? (as opposed to a list of reasons).

symroe:
> I'm not sure I understand: "add a filter to the admin to show all by-election reasons, including `None`". Do you mean all by-elections grouped by reason? (as opposed to a list of reasons).

Sorry, so we're able to:

1. Generate a list of all elections with [reason] 
2. So we're able to generate a list of all elections with no reason. This also requires a `is_by_election` filter to be useful (I thought we had this already when making this comment)

PR #2478, opened 2025-07-09
update-historic-election-seats-totals-and-seats-contested

awdem:
This PR contains a one-off command to update seats_total and seats_contested values for elections base on this csv: https://docs.google.com/spreadsheets/d/1FKgInyCSq3YE-m5bnsGnx6KULYZyDsl9ovfDyzoYAP0/edit?gid=1508742918#gid=1508742918

It won't update all of the election_ids in that csv, but it will update ~7000 out of ~9100. The remaining records are more complicated to update because their division don't have seats_total values, something we normally get from legislation.

However, this command will knock out the bulk of the sheet.

Tested locally. Has a `dry-run` flag.

---
  - https://app.asana.com/0/0/1210672041308456

### DIFF
--- a/every_election/apps/elections/management/commands/update_historic_seats_values_from_csv.py
+++ b/every_election/apps/elections/management/commands/update_historic_seats_values_from_csv.py
@@ -1,0 +1,64 @@
+from core.mixins import ReadFromCSVMixin
+from django.core.management import BaseCommand
+from django.db import transaction
+from elections.models import Election
+
+
+class Command(ReadFromCSVMixin, BaseCommand):
+    help = """
+    one-off command to set the correct seats contested and total seats for historic elections.
+
+    update_historic_seats_values_from_csv -u "https://docs.google.com/spreadsheets/d/e/2PACX-1vSXaZ20SSUmtN65_wqz8Zytu71h8MPBnDEwtCovfxRQASkSl2izgmWV1hceUgW8oSp5nKApw67PFeGI/pub?gid=1508742918&single=true&output=csv"
+    update_historic_seats_values_from_csv -f path/to/local/file.csv
+    """
+
+    def add_arguments(self, parser):
+        super(Command, self).add_arguments(parser)
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Run the command without making any changes to the database.",
+        )
+
+    def handle(self, *args, **options):
+        data = self.load_data(options)
+        records_to_update = []
+
+        for line in data:
+            # skip referendums and gla.a.
+            if line["election_id"].startswith(("ref.", "gla.a.")):
+                continue
+            record = Election.public_objects.get(
+                election_id=line["election_id"]
+            )
+            # update parl, mayor, pcc
+            if record.election_id.startswith(("parl.", "mayor.", "pcc.")):
+                if not record.seats_contested:
+                    record.seats_contested = 1
+                record.seats_total = 1
+                records_to_update.append(record)
+                continue
+
+            updated = False
+            if line["division_seats_total"].isdigit():
+                record.seats_total = int(line["division_seats_total"])
+                updated = True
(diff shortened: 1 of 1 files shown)


every_election/apps/elections/management/commands/update_historic_seats_values_from_csv.py
```diff
@@ -0,0 +1,60 @@
...
+                f"Dry run: would update {len(records_to_update)} records."
+            )
+        else:
+            self.save_records(records_to_update)
+
+    @transaction.atomic()
+    def save_records(self, records):
+        Election.public_objects.bulk_update(
+            records, ["seats_total", "seats_contested"]
+        )
```
awdem, line 59: I'm using bulk_update() here since I'm updating ~7000 records. However, [bulk_update()](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#bulk-update) doesn't call the save on each model object, so I don't think it will update the `modified` field. 

This means that these updates won't propagate to YNR unless the models are saved for some other reason. 

I'm not sure that it matters though, because `seats_contested` is the value that YNR uses to set the `winner_count` on its ballots. And I'm actually backporting those values from YNR here via the `ynr_seats_contested`  column on the sheet (in the case of `parl.`, `mayor.`, and  `pcc.`, only 9 records on the sheet don't have a seats_contested value set). 

However. what I don't know if other downstream projects need to "know" if these values have changed or not.

chris48s: We could explicitly set the modified date if we want but I think it is probably good if we _don't_ touch 7,000 ballot objects at once tbh. That's a lot for YNR to try and sync. If YNR has the right winner counts already then WCIVF should too.

every_election/apps/elections/management/commands/update_historic_seats_values_from_csv.py
```diff
@@ -0,0 +1,60 @@
...
+            # update parl, mayor, pcc
+            if record.election_id.startswith(("parl.", "mayor.", "pcc.")):
+                if not record.seats_contested:
+                    record.seats_contested = 1
+                record.seats_total = 1
+                records_to_update.append(record)
+                continue
+            # update the rest if we have a division seats_total
+            if line["division_seats_total"].isdigit():
+                record.seats_total = int(line["division_seats_total"])
```
chris48s, line 43: Where is `division_seats_total` from in the CSV? Did we get it from EE in the first place? Is there any chance that it can conflict with seats_total value on the related EE division?

every_election/apps/elections/management/commands/update_historic_seats_values_from_csv.py
```diff
@@ -0,0 +1,60 @@
...
+                if line["ynr_seats_contested"].isdigit():
+                    record.seats_contested = int(line["ynr_seats_contested"])
```
chris48s, line 45: Am I right that the premise of this command is that we are back-filling missing data (i.e: the seats_total and seats_contested values here should be `None` to start with - we don't expect to be overwriting existing values)? If so, is it worth checking that the fields are definitely null here before we overwrite? This slightly depends on how we got to this CSV in the first place.. if it is all sourced from YNR I'd probably add a quick check here. If we generated the CSV in the first place by looking for ballots in EE with null values for this then we probabyl don't need to bother with this.

chris48s (commented):
A few questions but they might need no follow up

every_election/apps/elections/management/commands/update_historic_seats_values_from_csv.py
awdem: The only column from YNR in the CSV is `ynr_seats_contested`. That column is the values that I set on YNR as part of the historic data import that I now want to update on EE. 

`division_seats_total` is from the EE division directly.

PR #2480, opened 2025-07-17
Changes for new Senedd elections

chris48s:
This PR does several semi-related things, but I decided to lump them into one PR because there's not really enough going on here that it should cause a problem. In fact it might even be easier to review it as one PR.

### 1. Upgrade `uk-election-ids==0.10.0` and account for changes

Changelog: https://github.com/DemocracyClub/uk-election-ids/releases/tag/0.10.0

Commits:
- 13935da
- d2ba860 (this is because I changed the description on AMS)
- 715a802

### 2. Changes supporting new Senedd elections

- 473da66
- b0accda
- e137f81

Assuming we go with this solution, we don't need to create a new organisation. We just make a new `DivisionSet` for the new constituencies and off we go.

### 3. Tangentially related change

- 55c8ba0

When I was looking at whether we needed to do anything about the change I made to `ELECTION_TYPES` I realised that

a) The answer was no
b) I could clean up some legacy code

At some point we did this to account for uk-election-ids not knowing about referenda, but now it does so we can just bin this. However it is not strictly related to this PR. We could have done this before upgrading to `uk-election-ids==0.10.0`

### DIFF
--- a/every_election/apps/elections/views/general.py
+++ b/every_election/apps/elections/views/general.py
@@ -4,9 +4,9 @@ from core.helpers import user_is_moderator
 from django.contrib.auth.mixins import AccessMixin
 from django.utils.html import mark_safe
 from django.views.generic import DetailView, ListView, TemplateView
-from elections.constants import ELECTION_TYPES
 from elections.forms import NoticeOfElectionForm
 from elections.models import Document, Election, ElectionType
+from uk_election_ids.datapackage import ELECTION_TYPES
 
 
 class ElectionTypesView(ListView):
@@ -28,11 +28,6 @@ class ReferenceDefinitionView(TemplateView):
         for et_key, et_record in OrderedDict(
             sorted(ELECTION_TYPES.items())
         ).items():
-            # for the moment leave 'ref' out of the docs
-            # because our spec for these is incomplete
-            if et_key == "ref":
-                continue
-
             et_record["slug"] = et_key
             et_record["subtype"] = None
 
--- a/every_election/apps/elections/constants.py
+++ b/every_election/apps/elections/constants.py
@@ -1,8 +1,0 @@
-from uk_election_ids.datapackage import ELECTION_TYPES
-
-# These types exist in EE but don't have defined behaviour in the IdBuilder
-ELECTION_TYPES["ref"] = {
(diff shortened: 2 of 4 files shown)


every_election/apps/elections/views/id_creator.py
```diff
@@ -88,6 +88,11 @@ def select_subtype(wizard):
     election_type = wizard.get_election_type
     if not election_type:
         return False
+    if (
+        election_type.election_type == "senedd"
+        and wizard.get_election_date().date() >= datetime.date(2026, 5, 7)
+    ):
+        return False
```
chris48s, line 95: I debated whether this should be something stored on the organisation object, and then we would create 2 different Senedd orgs, but I think it can just live here in code tbh. I think it is conceptually nicer that Senedd stays as one org anyway.

every_election/apps/elections/views/general.py
```diff
@@ -28,19 +28,20 @@ def get_context_data(self, **kwargs):
...
+                table_rec = et_record.copy()
+                table_rec["name"] += " (2026-05-07 onwards)"
+                election_types_table.append(table_rec)
             if et_record["subtypes"]:
                 # if we've got subtypes, duplicate the
                 # election type data for each subtype
                 for s_record in et_record["subtypes"]:
                     table_rec = et_record.copy()
+                    if et_record["slug"] == "senedd":
+                        table_rec["name"] += " (before 2026-05-07)"
```
chris48s, line 44: I've has a bash at presenting this in the table, but open to suggestions on it.

chris48s:
Next steps:

- Set an end date on the existing Senedd DivisionSet, create the new one, make the new IDs :rocket:
- I'd quite like to write a management command to null out the voting_system on any historic elections where we set it on a group like local.date, but that can be another PR and also the API serializer kinda has our back on this https://github.com/DemocracyClub/EveryElection/blob/499ae536f8814eb97d6fb50b85ded23b3137ff96/every_election/apps/api/serializers.py#L204-L214

symroe (approved):

chris48s:
I've now set voting_system=None on all the relevant historic elections, so the new DivisionSet is the next job after this

PR #2488, opened 2025-08-04
ni boundary import script

awdem:
a short script to run in django shell that copies the makes a new divset for 2027 NI boundaries and copies the WMC divisions over to it.


chris48s (approved):
yeah - happy for to just run this and close the PR

PR #2492, opened 2025-08-12
Annon by elections only

symroe:
This ended up being easier than I thought. I started writing it up and realised I could just do it. 

There is still a problem that elections to organisations can't have by-elections, and annon users can suggest these still. I've not touched this path at all, rather I'm just focusing on the path that already allows by-election creation (all ballots with a division).

---
  - https://app.asana.com/0/0/1211034641118652

### DIFF
--- a/every_election/apps/elections/templates/id_creator/by_election_source.html
+++ b/every_election/apps/elections/templates/id_creator/by_election_source.html
@@ -20,7 +20,7 @@
             {% endif %}
         {% endfor %}
 
-        {% ifchanged form.group %}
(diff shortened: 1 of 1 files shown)


symroe:
One thing I forgot to add is that https://github.com/DemocracyClub/EveryElection/pull/2492/commits/a2c3eb264f3584d76dd62fdcc8349a187d7f9714 is needed to fix a bug in production. 

I'll add the text, but if we can't merge this for some reason, I'll cherry-pick that commit out and deploy it. Right now we're asking for a source for every division, even if no election has been selected :(

PR #2505, opened 2025-09-02
Feat/multi div type pmtiles

awdem:
This PR implements pmtiles creation for division sets with more than on division type (e.g. Scottish Parliament, Welsh Assembly, old Senedd). 

mapgif2

In order to handle multiple division types, this PR adapts PMTilesCreator to create one feature layer per division_type for a divisionset. Layers are now named using `<divisonset_id>_<division_type>` rather than just `divisonset_id`.

e.g.
```bash
python manage.py create_pmiles_for_divset 23
# creates a pmtiles file: sp-23.pmtiles 
# with two layers named: 23_SPE and 23_SPC
```

To handle multiple layers, the division set map now generates map layers by iterating over the pmtiles layer metadata directly, instead of using the `divisionset_id` as the source layer name . This PR also adds a layer toggle element for maps with multiple layers.

One note about the styling, I've chucked all the toggle styling directly in the django template, but I've added a note to this [asana card](https://app.asana.com/1/1204880536137786/project/1207880534533137/task/1211196557393178?focus=true) to move that styling to the design system when making the maplibre map component. I plan to generally make the map component's styling more in line with our design system as part of that ticket (for example, right now all the elements under the map component use the wrong font).

Finally,  I'll have to regenerate the pmtiles for every divisionset once this PR is merged, because I've changed how the layers are named. This means the maps will be broken on production until I do so.

### DIFF
--- a/every_election/apps/organisations/pmtiles_creator.py
+++ b/every_election/apps/organisations/pmtiles_creator.py
@@ -1,36 +1,39 @@
 import subprocess
 
+import psycopg2
 from django.db import connection
 from organisations.models import DivisionGeography
 
 
 class PMtilesCreator:
     def __init__(self, divset):
-        div_subtypes = divset.divisions.values("division_subtype").distinct()
-        if div_subtypes.count() > 1:
-            raise NotImplementedError(
-                "PMTiles creation is not implemented for multiple division subtypes."
-            )
         self.divset = divset
 
     def create_pmtile(self, dest_dir):
         pmtiles_fp = f"{dest_dir}/{self.divset.pmtiles_file_name}"
+        # Create a geojson file for each division subtype
+        div_types = set(
+            self.divset.divisions.values_list("division_type", flat=True)
+        )
+        geojson_files = []
+        for div_type in div_types:
+            geojson_fp = self.create_geojson(dest_dir, div_type)
+            geojson_files.append(geojson_fp)
 
-        geojson_fp = self.create_geojson(dest_dir)
-        tippecanoe_command = f"tippecanoe -o {pmtiles_fp} -zg --drop-rate=2 --drop-densest-as-needed {geojson_fp} -l {self.divset.id}"
+        tippecanoe_command = f"tippecanoe -o {pmtiles_fp} -zg --drop-rate=2 --drop-densest-as-needed {' '.join(geojson_files)}"
         subprocess.run(tippecanoe_command, shell=True, check=True)
 
         return pmtiles_fp
 
-    def create_geojson(self, dest_dir):
-        geojson_fp = f"{dest_dir}/{self.divset.id}.geojson"
-        ogr_command = self.construct_ogr_command(geojson_fp)
+    def create_geojson(self, dest_dir, div_type):
+        geojson_fp = f"{dest_dir}/{self.divset.id}_{div_type}.geojson"
+        ogr_command = self.construct_ogr_command(geojson_fp, div_type)
         subprocess.run(ogr_command, shell=True, check=True)
         return geojson_fp
 
-    def construct_ogr_command(self, geojson_fp):
+    def construct_ogr_command(self, geojson_fp, div_type):
         db_settings = connection.settings_dict
-        sql_query = self.construct_sql_query(self.divset.id)
+        sql_query = self.construct_sql_query(self.divset.id, div_type)
 
         db_connection_string = (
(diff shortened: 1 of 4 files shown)


every_election/apps/organisations/pmtiles_creator.py
```diff
@@ -62,5 +65,11 @@ def construct_sql_query(self, divisionset_id):
...
             .query.sql_with_params()
         )
+
         sql = sql.replace("::bytea", "")  # Remove bytea typecast
-        return sql % params  # Substitute parameters into the SQL query
+
+        # Quote string params
+        quoted_params = tuple(
+            f"'{p}'" if isinstance(p, str) else p for p in params
+        )
```
chris48s, line 74: This won't handle edge cases on strings that contain quote characters. I would recommend

```py
import psycopg2

quoted_params = tuple(
    psycopg2.extensions.adapt(p).getquoted().decode() if isinstance(p, str) else p
    for p in params
)
```

here

every_election/apps/organisations/templates/organisations/divisionset_map.html
```diff
@@ -12,114 +53,197 @@
...
+
+                        // determine TextContent
+                        function getDivisionTypeLinkContent(code) {
+                            if (code.endsWith("E")) {
+                                return "Regions";
+                            } else if (code.endsWith("C")) {
+                                return "Constituencies";
+                            }
+                            return "Unknown";
+                        };
```
chris48s, line 172: I feel like what we're doing here is a bit messy and relies on the code making assumptions that might break or require updating over time if other bodies adopt AMS.
Here's an alternative suggestion:
When we render the DivisonSet view, can we use the DivisonSet DB object to write a JSON object into the DOM (using `|json_script`) like:

```json
{ "SPC": "Scottish Parliament constituency", "SPE": "Scottish Parliament region" }
```

..and then look up `code` in that object to get the description.

Also in the fallback case (lets say `code` isn't in the object or description is blank) just return the code here. I'd rather have a non-meaningful distinct label than 2 or more layers called "Unknown".

every_election/apps/organisations/pmtiles_creator.py
awdem: Fixed in [8428170](https://github.com/DemocracyClub/EveryElection/pull/2505/commits/8428170e0c9cce8fb832bac1394d9668a59a7307)

every_election/apps/organisations/templates/organisations/divisionset_map.html
awdem: In https://github.com/DemocracyClub/EveryElection/pull/2505/commits/77e19d6c1e5734cbb6301e3c440c80f5362d116d I've added a mapping generated from the DB object to the template context. I opted to use `Constituencies` and `Regions` still because I think using the entire `division_subtype` is long and also not plural. I think it's still future-proof in the same way?

I've also made the fallback case `divison_type` in both the mapping and the javascript.

awdem:
> I've just noticed there is a bug with [5442065]

Have addressed this in https://github.com/DemocracyClub/EveryElection/pull/2505/commits/fa8608386133dd48450d2e39a876af5ecb3654fa.

every_election/apps/organisations/templates/organisations/divisionset_map.html
```diff
@@ -224,21 +238,48 @@ <h3>Map</h3>
...
+                            <details class="map-popup-details" >
+                                <summary style="color: ${feat.layerColor};">${feat.layerDisplayName}</summary><br />
+                                <strong>${feat.name}</strong><br />
+                                ${feat.officialIdentifier}<br />
+                            </details>
+                            `).join('');
+                    } else {
+                        const feat = featuresData[0]
+                        popUpHTML = `
+                            <div style="border-top: 2px solid  ${feat.layerColor};">
```
chris48s, line 280: > I don't particularly like the colored top-border on the single feature popup

I think we can just remove the `style=` attr here then.

chris48s:
I've left one comment about getting rid of that top border. Aside from that, I think this is good to go.

every_election/apps/organisations/templates/organisations/divisionset_map.html
awdem: Done in https://github.com/DemocracyClub/EveryElection/pull/2505/commits/afae0c8a89e9a49b14ce387e9c664116100f430e

PR #2510, opened 2025-09-15
fix tippecanoe caching

awdem:
Deploys to EE are broken because the tippecanoe cache expired, and the step that adds the cache directory to the PATH happens before the install step so it's failing. This PR fixes that by moving the `Add tippecanoe to PATH` step to after the install step.


chris48s (approved):
This is fine, but as mentioned in standup lets also have a look at
https://launchpad.net/ubuntu/noble/+package/tippecanoe
If we can get away with just installing the version in the ubuntu/debian repos then that would be nice.

PR #2533, opened 2025-10-06
Feat/replace leaflet js

awdem:
This PR replaces the leaflet js map in `detail_map.html` with a maplibre map with the same functionality:

image

It also uninstalls `leaflet.js` and bumps the design system to `0.9.0` in order to use the `ds-maplibre-map` component.

---
  - https://app.asana.com/0/0/1211196557393182

### DIFF
--- a/every_election/assets/scss/styles.scss
+++ b/every_election/assets/scss/styles.scss
@@ -26,7 +26,6 @@
 @import 'partials/_codeblock.scss';
 @import 'partials/_dark.scss';
 @import 'partials/_page.scss';
-@import 'partials/_map-leaflet.scss';
 @import 'partials/_map-maplibre.scss';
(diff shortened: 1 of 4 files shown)


every_election/apps/core/templates/core/detail_map.html
```diff
@@ -2,59 +2,114 @@
...
+
+                            startCoord = geojson.geometry.coordinates[0][0][0];
+                            const bounds = new maplibregl.LngLatBounds(startCoord, startCoord)
+                            geojson.geometry.coordinates.forEach(polygon => {
+                                polygon.forEach(ring => {
+                                    ring.forEach(coord => {
+                                        bounds.extend(coord);
+                                    });
+                                });
+                            });
```
GeoWill, line 89: I think this is fine.

PR #2536, opened 2025-10-14
upgrade to django 5

chris48s:
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1211454481125632?focus=true

Notes on this: 

1. I don't think we need to change anything related to media or static files settings on this repo. It is just taking them all from dc_django_utils

2. I've deployed to dev: https://every-election.womblelabs.co.uk/ (might need waking up if it has gone to sleep)

3. One of the things I noted in the Asana ticket was this deprecation warning:

```
RemovedInDjango50Warning: The "default.html" templates for forms and formsets will be removed.
These were proxies to the equivalent "table.html" templates, but the new "div.html" templates will be the default from Django 5.0.
Transitional renderers are provided to allow you to opt-in to the new output style now.
See https://docs.djangoproject.com/en/4.2/releases/4.1/ for more details
```

but this actually got solved in https://github.com/DemocracyClub/EveryElection/pull/2534 by upgrading formtools so we don't need to do anything about it here.

4. Same warning we had on WCIVF about URLField. Lets deal with it under the new ticket we created for that

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -252,14 +252,14 @@ workflows:
         requires:
         - "CDK Synth"
         context: [deployment-development-ee, slack-secrets]
-        filters: { branches: { only: [ development, "form-tools20251015" ] } }
+        filters: { branches: { only: [ development ] } }
         dc-environment: development
     - code_deploy:
         name: "Development: AWS CodeDeploy"
         requires:
         - "Development: CDK Deploy"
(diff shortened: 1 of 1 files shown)


awdem (approved):
LGTM. Just need to sort out the form tools stuff as you said. And remove the dev deploy commit.

PR #2537, opened 2025-10-15
upgrade django-formtools==2.5.1 (attempt 2)

chris48s:
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1211648460534610?focus=true

Quick re-cap on this one:

- We did this originally in https://github.com/DemocracyClub/EveryElection/pull/2534
- It "worked" locally, but we deployed it, and it caused catastrophic performance issues
- I reverted the changes to fix it https://github.com/DemocracyClub/EveryElection/commit/8d39d10e9e00515865b4b5f8973a362a2296fd4f

In this PR, I am doing the upgrade a second time but this time in a way that retains reasonable performance. I've broadly succeeded. This version makes a lot fewer round-trips to the DB than the previous version we deployed. That said, it is worth noting that this version of django-formtools does make more DB requests than the previous one we were on and I don't think that is completely fixable in userspace.

There is an open PR upstream that might help with this if it is ever merged https://github.com/jazzband/django-formtools/pull/291

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -252,14 +252,14 @@ workflows:
         requires:
         - "CDK Synth"
         context: [deployment-development-ee, slack-secrets]
-        filters: { branches: { only: [ development ] } }
+        filters: { branches: { only: [ development, "form-tools20251015" ] } }
         dc-environment: development
     - code_deploy:
         name: "Development: AWS CodeDeploy"
         requires:
         - "Development: CDK Deploy"
         context: [deployment-development-ee, slack-secrets]
-        filters: { branches: { only: [ development ] } }
+        filters: { branches: { only: [ development, "form-tools20251015" ] } }
         dc-environment: development
     - cdk_deploy:
         name: "Staging: CDK Deploy"


knyghty:
Hi @chris48s, I noticed this when working again on the upstream PR for formtools.

It would be great if you could test the branch of my PR on this project, even if only locally, and report back upstream if you run into any issues. It may help get the PR merged faster.

chris48s:
Hi @knyghty

Thanks for your upstream work on this.

I installed

```
django-formtools = { git = "https://github.com/knyghty/django-formtools", rev = "f6665ed0db1b7238872fa2751ce74b808e8194d1" }
```

locally and had a quick play with django debug toolbar open.

The good news is:

- Using your changes reduces the number of round-trips to the DB needed to render each step in the wizard
- I tried rolling back all the optimisations I applied in this PR and the number of round-trips to the DB stayed about the same without them. So if your PR is merged upstream, we could get rid of some messy workarounds and optimisations we've had to apply here
- I did not encounter any bugs or errors while testing.

That said, cache invalidation bugs can often be subtle and the testing I've done was not exhaustive.

Thanks again for your help on this.

PR #2560, opened 2025-11-25
show seats cancelled/seats contested on election pages

### LINKED ISSUE #1082: Count cancelled and non-cancelled seats seperately on /elections/<id>
At the moment, the "Seats Up" total includes cancelled seats. On a page like this one, this is unhelpful because the total includes several hundred seats where no election is actually taking place.

Screenshot at 2021-05-06 20-01-00

That said, excluding them completely would be odd on a page like this because you'd say "Seats Up: 0"

Screenshot at 2021-05-06 20-01-25

I think the solution is to present total/cancelled/non-cancelled as 3 different totals for clarity

chris48s:
Closes #1082
Refs https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1212087409258934?focus=true

This PR modifies the election detail pages to show:

- seats contested (which is now seats contested _in all non-cancelled ballots_ )
- seats cancelled (if >0 ballots are cancelled)

I've opted to now show "seats cancelled: 0" where it is not relevant as it probably just confuses the issue.

### DIFF
--- a/every_election/apps/elections/templates/elections/election_summary.html
+++ b/every_election/apps/elections/templates/elections/election_summary.html
@@ -77,23 +77,25 @@
                     </div>
                 {% endif %}
             {% else %}
-                <div>
-                    <dt
-                        title="The total number of seats in this division"
-                    >Total Seats</dt><dd>{{ object.seats_total }}</dd>
-                </div>
-                {% if object.cancelled %}
+                {% if object.seats_total is not None and object.seats_contested is not None %}
                     <div>
                         <dt
-                            title="Number of seats that would have been contested in this ballot, were it not cancelled"
-                        >Seats Cancelled</dt><dd>{{ object.group_seats_cancelled }}</dd>
+                            title="The total number of seats in this division"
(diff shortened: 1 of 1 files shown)


GeoWill (approved):
This looks good an seems to do it's job. Agree it's hard to explain!d Think what you've done is pretty good. I guess it's not so useful on mobile, but EE is probably one of our sites that has less mobile traffic. 

I'm pretty sure it's not allowed to cancel some of the seats contested on a ballot. At least we can't model that if it is possible. 

-------------------------------------

One thing I thought while reviewing it is how much better the styling of `cancelled` is in the pre-design system design. 

image
vs 
image

I think it's because we don't style the `"label alert"` classes that cancelled is wrapped in. 
([link](https://github.com/DemocracyClub/EveryElection/blob/5f87e8a07073bdc7ff6ad73785b7e2a6adb81379/every_election/apps/elections/templates/elections/elections.html#L10))

I'm not sure there's an equivalent in the design system atm.

PR #2573, opened 2026-01-21
Command for making Unitary Authorities

GeoWill:
At the moment this command only allows the creation of geographies from existing organisations. A possible future requirement will be that the geography is made up from unioning wards/CEDs together.

This doesn't get into the weeds of making the divisions. But maybe that will be a future enhancement...

### DIFF
--- a/every_election/apps/organisations/management/commands/create_unitary_authority.py
+++ b/every_election/apps/organisations/management/commands/create_unitary_authority.py
@@ -1,24 +1,11 @@
 from datetime import datetime
 
-from django.contrib.gis.geos import MultiPolygon, Polygon
 from django.core.management import BaseCommand, CommandError
 from django.db import transaction
 from django.utils.text import slugify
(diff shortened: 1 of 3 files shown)


every_election/apps/organisations/management/commands/create_unitary_authority.py
```diff
@@ -0,0 +1,205 @@
...
+            elected_title="Councillor",
+            election_type=ElectionType.objects.get(election_type="local"),
+            organisation=self.ua_organisation,
+            elected_role_name=f"Councillor for {self.org_common_name}",
+        )
+        self.stdout.write(
+            self.style.SUCCESS(
+                f"Created ElectedRole: Councillor for {self.org_common_name}"
+            )
+        )
```
GeoWill: This didn't happen for Somerset, Westmorland and Furness, nor Cumberland. So doing that is another ToDo.

PR #2590, opened 2026-05-12
Boundary bot fixes 20260511

awdem:
This PR fixes two issues with boundary bot.

e51bdfe19498739ccdebfbba72a97b61e32212e8 fixes a bug where we scraping upcoming events as the latest event as described [here](https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1214171430825032?focus=true).

9a749a4b252bcbc726099ca60583ad7d6fb61a56 updates the parser to handle changes to the website. Recent reviews don't list the legislation title and link in the 'Latest Information' section, but rather in 'Previous Stages':
image

Older reviews still do:
image

 I've updated the parser to check the  'Latest Information' section first and then try 'Previous Stages'. 
 
 Tested by running locally.

---
  - https://app.asana.com/0/0/1214597478910528

### DIFF
--- a/every_election/apps/organisations/boundaries/boundary_bot/spider.py
+++ b/every_election/apps/organisations/boundaries/boundary_bot/spider.py
@@ -79,6 +79,21 @@ class LgbceSpider(scrapy.Spider):
             return latest_review_card_title.xpath("text()")[0].extract().strip()
         return None
 
+    def get_eco_containers(self, response):
+        latest_info_containers = response.xpath(
+            '//div[@class="latest-information"]//div[@class="link-name-and-view-container"]'
+        )
+        # if there's a link container in latest info, use that
+        if latest_info_containers:
+            return latest_info_containers
+        # otherwise, look for a made order div in previous stages
+        previous_stages_containers = response.css("div.previous-steps").xpath(
+            '//div[h5[contains(text(), "Made Order")]]//div[@class="link-name-and-view-container"]'
+        )
+        if previous_stages_containers:
+            return previous_stages_containers
+        return []
+
     def get_eco_title_and_link(self, response, latest_event):
         def get_link_title(selector):
             return selector.xpath(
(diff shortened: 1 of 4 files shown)


every_election/apps/organisations/boundaries/boundary_bot/spider.py
```diff
@@ -68,11 +68,32 @@ def get_shapefiles(self, response):
...
+        # get review cards but exclude Upcoming
+        review_card_divs = response.css("div.vits-review-card").xpath(
+            "self::div[not(div[contains(., 'Upcoming')])]"
+        )
+        if review_card_divs:
+            latest_review_card_title = review_card_divs.css(
+                "h3.vits-review-card__title"
+            )[-1]
+            return latest_review_card_title.xpath("text()")[0].extract().strip()
         return None
```
GeoWill: I think this works, but I'm worried they're going to put 'Upcoming' in the text somewhere. 

At the moment it looks to me like we can grab the 'in progress' card by getting the `div` with the classes `vits-review-card vits-review-card--secondary`. This seems a bit more precise to me, and then we can grab the `"h3.vits-review-card__title"` title out of that.

GeoWill (commented):
One way I think we could be more targetted since they give the 'in progress' review card it's own class. I've looked at a few reviews and this seems consistent. If it doesn't work for some reason, then lets just stick with looking for (and excluding) `'Upcoming'`.

awdem:
> One way I think we could be more targetted since they give the 'in progress' review card it's own class. I've looked at a few reviews and this seems consistent. If it doesn't work for some reason, then lets just stick with looking for (and excluding) `'Upcoming'`.

We also want to target completed review cards as well, though, because pages without current reviews don't have in progress cards ([example](https://www.lgbce.org.uk/all-reviews/cambridgeshire)).

every_election/apps/organisations/boundaries/boundary_bot/spider.py
awdem: Rather than grab the 'in progress' card, I added another condition so it now specifically excludes review cards if they have a  `vits-review-card__status` div that says `Upcoming`  (in https://github.com/DemocracyClub/EveryElection/pull/2590/commits/38fe7eb042168c11156962c4a8dc46efc1669d19).

GeoWill (approved):

PR #2591, opened 2026-05-19
allow creating mayor/PCC by-elections through the wizard

chris48s:
Does what it says on the tin really: At the moment you can't create a by-election for a mayor or PCC election in the wizard. This PR adds that functionality.

There is about to be a super high-profile mayoral by-election in Greater Manchester, so I'd like to be able to just easily make the ID once a date is announced.

### DIFF
--- a/every_election/apps/elections/views/id_creator.py
+++ b/every_election/apps/elections/views/id_creator.py
@@ -14,6 +14,7 @@ from elections.forms import (
     ElectionOrganisationForm,
     ElectionSubTypeForm,
     ElectionTypeForm,
+    MayorPCCBallotTypeFormSet,
 )
 from elections.models import (
     Document,
@@ -36,6 +37,7 @@ FORMS = [
     ("election_subtype", ElectionSubTypeForm),
     ("election_organisation", ElectionOrganisationForm),
     ("election_organisation_division", DivFormSet),
+    ("mayor_pcc_ballot_type", MayorPCCBallotTypeFormSet),
     ("by_elections_source", ByElectionSourceFormSet),
     ("review", forms.Form),
 ]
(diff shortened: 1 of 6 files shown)


every_election/apps/elections/views/id_creator.py
```diff
@@ -103,10 +106,27 @@ def select_organisation_division(wizard):
...
     """
-    We never ask for a source when creating elections for users that are
-    logged in
+    We never ask for a source when creating elections
+    for users that are logged in
+
+    Also skip by-election source for Mayor/PCC by-election
+    to keep things simpler
     """
     return not wizard.request.user.is_authenticated and wizard.get_divisions()
```
chris48s, line 131: Quick note here:
At the moment, the by-election source stuff is quite long and hard-coded to the assumption of one source per division
https://github.com/DemocracyClub/EveryElection/blob/f22a255ad561c934cb4f8b656e1e6a04d798577a/every_election/apps/elections/forms.py#L274-L327

I did consider whether to make a second version of this that does one form per org or generalise the existing code.

Given that:
- PCCs are soon to be abolished completely
- Mayor by elections are rare and high-profile: We'll probably be creating them ourselves rather than receiving an anonymous suggestion about a mayoral by-election we didn't already know about.

I decided to just keep things simpler here by skipping the source step for suggested mayoral by-election. I don't think the juice is worth the squeeze.

GeoWill (approved):
This looks good. I've had a play locally and it all seems to work. 

As an aside I do think it would be nice to have some generalised special casing of 'at large' elections. However, PCCs are about to be abolished, and then our only at large election type is 'mayor'. So it makes sense that the special case is on `election_type == 'mayor'`.

every_election/apps/elections/views/id_creator.py
GeoWill: This seems like the right trade off.

chris48s:
Even when PCCs are abolished, we still have to keep all the code around because we already have loads of historic PCC elections. It might still make sense to consolidate this concept somewhere as we do have a bunch of special cases for "mayor or PCC"

PR #2601, opened 2026-06-03
Incompatible role triggers by-election

GeoWill:
Adds ByElectionReason choices for when an elected member wins another election and can't do both jobs at once.

Sort of a special case of 'DISQUALIFIED' but without sounding like they've done something bad.

https://app.asana.com/1/1204880536137786/project/1204880927741389/task/1214801755278138?focus=true


chris48s:
This is fine, but there is a ruff formatting error that needs to be fixed before merging. Also there are some corresponding PRs to make on other repos

chris48s (approved):

PR #2604, opened 2026-06-16
make example postcodes better

chris48s:
Quick papercut bug fix. I just saw this code out of the corner of my eye and realised why our example postcodes are sometimes not very helpful


symroe (approved):

GeoWill (approved):
Nice!

PR #2609, opened 2026-06-29
Update copy divisions

awdem:
This PR adds an optional boolean flag to the `copy_divisions` command that lets you explicitly exclude geographies from the copy. The default copy includes geographies.

I've added this to make processing Welsh Community Reviews easier. Community Reviews can affect electoral ward boundaries, but not names or # of councillors. I just want to copy the latter two things from the previous division set and exclude the old boundaries.

### DIFF
--- a/every_election/apps/organisations/management/commands/copy_divisions.py
+++ b/every_election/apps/organisations/management/commands/copy_divisions.py
@@ -25,7 +25,7 @@ class Command(BaseCommand):
         )
 
     @transaction.atomic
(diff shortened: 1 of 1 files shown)


every_election/apps/organisations/management/commands/copy_divisions.py
```diff
@@ -1,3 +1,5 @@
+from argparse import BooleanOptionalAction
```
chris48s, line 1: Nice. TIL `BooleanOptionalAction` 🎓

chris48s (approved):

PR #2613, opened 2026-07-08
upgrade uk-election-timetables, add notice_of_election_deadline and sopn_publish_deadline fields

chris48s:
PR notes

This PR upgrades EE to use `uk-election-timetables==5.0.0`
Release notes: https://github.com/DemocracyClub/uk-election-timetables/releases/tag/5.0.0

Most of the churn here is related to

- Making use of the new `notice_of_election_deadline` and `sopn_publish_deadline` properties
- Adding unit tests

I'll leave some comments inline calling out specific things.

Once this is in, https://github.com/DemocracyClub/aggregator-api/pull/642 makes sense

So

- https://github.com/DemocracyClub/EveryElection/pull/2613
- https://github.com/DemocracyClub/UK-Polling-Stations/pull/9506
- https://github.com/DemocracyClub/aggregator-api/pull/642

Make sense to all merge as a block

Then we're on to moving downstream apps to remove use of uk-election-timetables + consume the API instead


every_election/apps/elections/migrations/0090_backfill_new_timetable_fields.py
chris48s, line 1: This is another migration that is going to touch the modified date on basically every ballot object in EE
I do at least know this time round that it will break the EE --> YNR sync and what to do about it when it does

every_election/apps/elections/tests/test_election_model.py
```diff
@@ -438,6 +446,64 @@ def test_no_id_required(self):
         self.assert_timetable_fields_all_none(election_group)
         self.assert_timetable_fields_all_none(org_group)
 
+    def test_nothern_ireland_ballot(self):
```
chris48s, line 449: This is a test I should have really added in https://github.com/DemocracyClub/EveryElection/pull/2608

Figured I might as well do it here

every_election/apps/elections/tests/test_election_model.py
```diff
@@ -536,3 +605,201 @@ def test_provisional_ballot(self):
         self.assert_timetable_fields_all_none(ballot)
         self.assert_timetable_fields_all_none(election_group)
         self.assert_timetable_fields_all_none(org_group)
+
+
+class TestCleanMethod(TestCase):
```
chris48s, line 610: We didn't have tests for the clean() method at all and this PR adds complexity to it
I told a LLM to generate tests for it and then did a review/tidy up on it. So this block of ~200 lines is mostly LLM output with a few edits from me.

every_election/apps/elections/migrations/0090_backfill_new_timetable_fields.py
```diff
@@ -0,0 +1,83 @@
...
+    "ENG": Country.ENGLAND,
+    "NIR": Country.NORTHERN_IRELAND,
+    "SCT": Country.SCOTLAND,
+    "GBN": None,
+}
+
+
+def backfill_new_timetable_fields(apps, schema_editor):
+    Election = apps.get_model("elections", "Election")
+    for election in Election.private_objects.using(
```
symroe: Maybe this doesn't matter for a one off query, but I reckon you can use `exclude()` here to catch some of the `continue` cases below? I also suspect it would help to use `annotate` to attach the `territory_code`(s), or at least use `select_related` with `only()` to get the fields, rather than making n+1 queries in the migration. If this times out, it could be a bit of a pain to unpick.

symroe: I wrote the above before seeing the later commit, but I think some of these points stand still? Feel free to ignore if you think it's ok

symroe (approved):
Approving even though I had a comment about the migration. Everything else looks fine and that migration doesn't really matter if you think it'll run ok.

every_election/apps/elections/migrations/0090_backfill_new_timetable_fields.py
chris48s: These are useful suggestions. Anything I can do to reduce the chances of the migration hanging on deploy is good. I've implemented these suggestions in https://github.com/DemocracyClub/EveryElection/pull/2613/commits/9eb9f00ce7d6bff59879ae463adb22755d8a3a54 It took about 5-10 seconds off the run time in dev, but that will translate to more when we run it on deploy and every round-trip to the DB is network request.

PR #2614, opened 2026-07-16
Add public_visibility field

awdem:
this PR adds a `dc_api_stage` field to the boundary review model that is an enum with three values. This field will be used by data baker to filter what boundary changes to include in the parquet ([here](https://github.com/DemocracyClub/dc-data-baker/blob/fb8c073f6e5ca70d345605a1213e0d10b88ef660/cdk/shared_components/lambdas/create_boundary_changes_csv/create_current_boundary_reviews_csv.py#L58)).

The enum values are:
- INITIAL - we've scraped/created a BR but it's not ready to show users yet. These will be excluded from the parquet.
- CONSULTATION - we have enough info to show at least block of text with links to the consultation.
- MAP - we have enough info + boundaries to show maps to users

`dc_api_stage` will be have to be managed manually as part of the regular boundary change work, e.g. once we've processed an ECO and created a new division set, we can set the boundary review to MAP.

I'm not thrilled about adding more manual overhead, but I can't think of a good way to automate it.

Also, I'm not confident about the field name or value so suggestions are welcome.


awdem:
I've change the field name to `public_visibility`, as I think it's clearer, and I've pared back the possible values to two:
- `HIDDEN`
- `MAP`

We can add more fine-grained fields like `CONSULTATION` once we've got a pipeline for connecting boundary changes to uprns in areas without elections.

every_election/apps/organisations/models/divisions.py
```diff
@@ -328,6 +333,11 @@ class OrganisationBoundaryReview(TimeStampedModel):
     consultation_url = models.URLField(blank=True, default="")
     boundaries_url = models.CharField(blank=True, default="")
     status = models.CharField(choices=ReviewStatus.choices)
+    public_visibility = models.CharField(
+        choices=PublicVisibility.choices,
+        blank=True,
```
chris48s: If its an enum with a default we don't need to allow blank

chris48s (approved):

every_election/apps/organisations/models/divisions.py
awdem: Fixed in https://github.com/DemocracyClub/EveryElection/pull/2614/commits/46c25f0514fbe1b3fd10503bc5e90fba6f2ed289

Is that only true in the case of enums? Because a lot of the other fields on this model have `blank=True, default=""`.

PR #2616, opened 2026-07-23
move uk-election-timetables pkg to workspace in this repo

chris48s:
The point of this PR is to move the uk-election-timetables package from being a standlone repo to a UV workspace under EE.

My rationale for this is:

- I've removed the uk-election-timetables package as a dependency from most of our projects already
- Removing it from the EC site is [WIP](https://github.com/DemocracyClub/ec-postcode-lookup-pages/pull/245)
- Ultimately, the package should have one consumer under the DC org: Every Election
- In principle, it is still useful that we can publish it as a package if anyone else wants to use it

Given we expect that moving forwards this package should have only one consumer it makes sense for those to just live in the same repo. It will make it easier to evolve them together. UV workspaces makes that possible.

So that's why I think this has advantages, but there are also some possible downsides too, which we will get into.

I have tried to structure the PR in a way that makes this relatively easy to review.

First off, there is commit 8d340585
This commit is big, but all it does is copy files from
https://github.com/DemocracyClub/uk-election-timetables
into this repo. So that commit shouldn't need review really.

Then everything I've done that changes anything is its own commit on top of that, so you can start reviewing from f7b894cf and that makes it pretty manageable.

While the diff of 8d340585 should not need review, there are a couple of points I want to make about it:

1. I've not copied the entire repo - just the bits we need. So to give an example: I didn't copy the `.circleci` or `.github` directories over. I left them out of the initial copy and then just created new CI stuff from scratch in the repo root.
2. This gives us one of the possible disadvantages of this move. https://github.com/DemocracyClub/uk-election-timetables has a lot of history. There are commit messages, pull requests etc that we can refer back to. Just copying it all in one commit essentially squashes all that down into one commit. I'm not sure that there is a way round this :(


scripts/test.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -euxo pipefail
+
+pytest
+cd packages/uk-election-timetables
+pytest
```
chris48s, line 6: I thought about whether to make it so that running pytest in the root should run the package tests as well.
A property that we want to preserve here is good "hygiene" between EE and the timetables package so we can't accidentally make uk-election-timetables depend on EE and we retain good developer experience for just working on the package.
Keeping the test suites separate helps with this because it means we can run the package tests in an isolated build that doesn't install the root dependencies (e.g: django).
It also means if you just want to work on the package, you can work in that subdirectory and just call `pytest` in there to run the package tests only.
So the two test suites are independent. Doing this also allows us to easily run the package builds in parallel with a build matrix, which is good.
But I did make a script you can use to run both of them. That seems kinda useful.
The other thing I haven't done is merged the coverage reports. At the moment, only the results of the application test suite go to coveralls. I can have a look at merging them once we are sure this direction of travel makes sense

packages/uk-election-timetables/uk_election_timetables/__init__.py
```diff
@@ -0,0 +1,3 @@
+from importlib.metadata import version
+
+__version__ = version("uk-election-timetables")
```
chris48s, line 3: I've made a bit of a change here that I think makes sense more widely as we go to using UV for packaging.
One way you can manage your version is to set `__version__` in a python file and use `dynamic = ["version"]`
You can also do it the other way round: Manage the version number in `pyproject.toml` and then import it when we need it in python. That's what I've done here.
For packages that support python >=3.10 this is the way to go IMO. You get to keep single source of truth for the version number and `__version__` etc, but we provide static metadata and you can bump it with `uv version` if you want, which is good IMO.

.github/workflows/check-for-bank-holiday-updates.yml
chris48s, line 1: I've checked this runs.
One of the issues we have with this is that GitHub disables scheduled actions on public repos if there has not been a commit in 60 days
This means sometimes GitHub disables the action because it is not unusual for https://github.com/DemocracyClub/uk-election-timetables to go 60+ days without a commit, and the holiday calendar doesn't change that often.
I'm hoping that EE is active enough that this should never be disabled. I would think it is very rare this repo goes for 60+ days without any commit

.github/workflows/python-publish.yml
chris48s, line 1: I have tested this as far as making sure that the "build" step builds the package, writes the artifact, and that the wheel has the right files in it. I've not tested the "publish" step.

There is a **TODO** here: Before we publish the first version, we need to register the DemocracyClub/EveryElection repo as a trusted publisher for this package over on PyPI. I have not done this yet.

This also points to another small consideration. This is set to trigger on a release being published. We don't currently do releases for the EE Web App itself (and I think it makes sense to treat it as "rolling") so I'm going to suggest we include the package name in the release titles when we make them to make it clear what the release represents.

packages/uk-election-timetables/README.md
chris48s, line 1: Here I have copied some content over from
https://github.com/DemocracyClub/uk-election-timetables/blob/main/README.md
but I've seperated the docs into 

- `README.md` - user-facing (how to install/use the package)
- `DEVELOPMENT.md` - dev-facing (how to run the tests, do a release, etc)

Do you reckon there's any other docs stuff I need to cover?

packages/uk-election-timetables/DEVELOPMENT.md
```diff
@@ -0,0 +1,43 @@
...
+This is set to run automatically on a schedule
+https://github.com/DemocracyClub/EveryElection/actions/workflows/check-for-bank-holiday-updates.yml
+
+## Documentation
+
+```
+uv sync --package uk-election-timetables --group docs
+```
+
+There are sphinx docs, and it is possible to build them but they need a bit of work.
```
chris48s, line 39: The docs site for this application is in a bit of a sad state.
It used to deploy to readthedocs.io but that hasn't worked in ages and we can't get access to the project.
It relies on some old deprecated stuff. For example https://github.com/readthedocs/recommonmark has been dead for years.
If you build the docs, there's a bunch of warnings and errors.
I think there is a piece of work to do here around improving this and deploying it somewhere from CI.
For the moment, I've put the packages you need to build the docs in here and made sure sphinx-build can run and spit out some HTML. Then I've just left it there.

Those might be problems we want to solve at some point, but I've decided for now the scope of this PR is just to "lift and shift" what we already have. Don't make anything worse than it is now. Don't shave any additional yaks we don't have to in this PR.

packages/uk-election-timetables/pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+name = "uk-election-timetables"
+version = "5.0.0"
+description = "Derives significant dates for UK elections"
+requires-python = ">=3.10"
+dependencies = []
+
+[dependency-groups]
+dev = [
+    "pytest==9.0.3",
+    "requests==2.33.0",
```
chris48s, line 11: While I was working on this, I realised a file in this package has an undeclared dependency on `requests`

This needs to be installed to load `bank_holidays.py`
https://github.com/DemocracyClub/uk-election-timetables/blob/1f16ca5f1264abb52702de616e8a1e56531080a0/uk_election_timetables/bank_holidays.py#L6

Naughty naughty
https://github.com/DemocracyClub/uk-election-timetables/blob/1f16ca5f1264abb52702de616e8a1e56531080a0/.github/workflows/check-for-bank-holiday-updates.yml#L21

I think I want to fiddle with this a bit at some point to get the bit that depends on requests out of the files we ship with the package. That stuff only needs to be in the maintenance script.

That's another yak for another day though.

chris48s:
Assuming we go forward with this, there's a little bit of cleanup TODO to decommission the old repo:

- Archive the repo
- Un-register trusted publisher

etc

symroe:
When I was moving some things around for LGSF I found `git-filter-repo` and some commands that let me move every file to a new location while maintaining the git commit history for it. 

I expect we can maintain the history of the package by moving the commits inside the repo, adding that as a remote and merging with `--allow-unrelated-histories`.

chris48s:
Thanks. I've done a bit of reading on this, but it is not something I've done before.

In general, I think it probably is worthwhile to preserve the commit history if we can. However I think I will have to go back to the start, do that, replay my new commits on that and submit a new PR.

Also, I wonder what reviewing a PR that copies all that history is going to look like?

Anyway, I will have a go at it this week. If the history ends up being really messy maybe that is a good candidate to try out stacked PRs: One PR for the old repo history and then the new stuff on top of it.. I will have an experiment.

chris48s:
I've replaced this with:

- https://github.com/DemocracyClub/EveryElection/pull/2617
- https://github.com/DemocracyClub/EveryElection/pull/2618

and taken the opportunity to try out a stack

PR #2617, opened 2026-08-05
move uk-election-timetables pkg to workspace in this repo (part 1 of 2)

chris48s:
This PR is part of a stack: https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

This is the first time we've used this feature, so its a learning experience for all of us.

---

This pull request should be equivalent to
https://github.com/DemocracyClub/EveryElection/pull/2616/changes/8d340585a7fbbd13916b6ed55d6d461c823dcae6
except that instead of squashing everything down into one commit, I've preserved the history from the uk-election-timetables repo using the `filter-repo`/`--allow-unrelated-histories` technique. So this PR has a lot of commits, but it shouldn't really require review.

While this does preserve commit history, all the merge commits now reference the wrong PRs because "merge pull request from #63" now points at https://github.com/DemocracyClub/EveryElection/pull/63 and not https://github.com/DemocracyClub/uk-election-timetables/pull/63
So this is not perfect, but probably still better than a kick in the teeth I guess.

The next PR will build on top of this.

I've not copied the entire repo - just the bits we need. So to give an example: I didn't copy the `.circleci` or `.github` directories over. I left them out of the initial copy and then just created new CI stuff from scratch in the repo root.


chris48s:
Just for completeness, I did this with:

```bash
git filter-repo --path docs --path tests --path LICENSE.md --path manage_bank_holidays.py --path uk_election_timetables --to-subdirectory-filter packages/uk-election-timetables
```

symroe (approved):
There's a CI fail and a merge conflict. Also, it's not like I've read all of this or anything (that's not the point). 

I think this works nicely to keep the history. It's a bit of a shame about the PR links, but, I don't think we look at them much anyway.

chris48s:
So the CI fail is expected because I don't update the ruff formatting or pytest config in this PR. That is done in https://github.com/DemocracyClub/EveryElection/pull/2618

I think its good to keep

- history I've copied from A to B and
- new changes I have made

in 2 different PRs, even if that means the first one has a failing build.

I'm not seeing a merge conflict though.
There is a conflict on `uv.lock` in https://github.com/DemocracyClub/EveryElection/pull/2618 which I will resolve before we merge the stack :+1:

PR #2630, opened 2026-09-14
organisation change legislation (1 of 2)

awdem:
This PR is part 1 of a stack

---

This PR attempts to model UK legislation that affects organisations boundary changes. These include legislation like Structural Change Orders, or Boundary Amendment Orders -- anything that can:
- Create or end organisations
- Update organisation boundaries

Some examples include:
[The Surrey (Structural Changes) Order 2026 ](https://www.legislation.gov.uk/uksi/2026/264/made
) - Ends 12 orgs and create 2 new ones
[The Glasgow City Council Area and North Lanarkshire Council Area (Cardowan by Stepps) Boundaries Amendment Order 2018](https://www.legislation.gov.uk/ssi/2018/308/made) - updates boundaries for 2 orgs
[The Hampshire and the Solent Combined County Authority Regulations 2026](https://www.legislation.gov.uk/uksi/2026/595) - creates a new org

The new `OrganisationChangeLegislation` model has fields for legislation information, a public visibility field for use in future data pipelines, a information URL field, and explanation text field for custom explanation text. It also has a M2M relationship with `Organisation` through the `OrganisationChange`  model. 

The  `OrganisationChange` model has its own information URL field, in case there is an org-specific info site, and start/end date validation comparing to the OCL effective date.

This PR also includes an admin interface for managing `OrganisationChangeLegislation` and `OrganisationChange`  objects.

Follow Up Work:

- Add an explanation in the admin for what `OCLPublicVisibility`  values actually correspond to on WCIVF. 
---

---
  - https://app.asana.com/0/0/1218266317355784


every_election/apps/organisations/models/organisations.py
```diff
@@ -214,3 +219,118 @@ class OrganisationGeographySubdivided(models.Model):
...
+
+
+class OCLPublicVisibility(models.TextChoices):
+    HIDDEN = "HIDDEN", "Hidden"
+    INFORM = "INFORM", "Inform"
+    MAP = "MAP", "Map"
+
+
+class OrganisationChangeLegislation(TimeStampedModel):
+    name = models.CharField(max_length=255)
```
chris48s: Can you explain the difference between `OrganisationChangeLegislation.name` and `OrganisationChangeLegislation.legislation_title`. When would these need to be different?

every_election/apps/organisations/models/organisations.py
```diff
@@ -214,3 +219,118 @@ class OrganisationGeographySubdivided(models.Model):
...
+            raise ValidationError(**error_kwargs)
+
+
+class OCLPublicVisibility(models.TextChoices):
+    HIDDEN = "HIDDEN", "Hidden"
+    INFORM = "INFORM", "Inform"
+    MAP = "MAP", "Map"
+
+
+class OrganisationChangeLegislation(TimeStampedModel):
```
chris48s, line 313: Minor: Can we use `verbose_name_plural` to set the plural name to "Organisation change legislation" instead of "Organisation change legislations"

every_election/apps/organisations/models/organisations.py
```diff
@@ -1,7 +1,12 @@
+from datetime import timedelta
```
chris48s: Minor: If we are adding new imports from the datetime module, can we use this import style please https://adamj.eu/tech/2019/09/12/how-i-import-pythons-datetime-module/

every_election/apps/organisations/tests/test_organisation_models.py
```diff
@@ -1,17 +1,19 @@
-from datetime import date
+from datetime import date, timedelta
```
chris48s: ..and here

every_election/apps/organisations/models/organisations.py
```diff
@@ -214,3 +219,118 @@ class OrganisationGeographySubdivided(models.Model):
...
+            models.UniqueConstraint(
+                fields=["organisation_change_legislation", "organisation"],
+                name="unique_ocl_organisation",
+            )
+        ]
+
+    def __str__(self):
+        return f"Org Change for {self.organisation.common_name} ({self.change_type})"
+
+    def clean(self):
```
chris48s, line 264: I'm kind of in two minds about these validation checks in that sometimes the right way to "fix" them is to go and change the Organisation model (which isn't on this form). I think we probably just roll with this to start with, but I can see we might end up revising this once we're using it in anger.

every_election/apps/organisations/models/organisations.py
```diff
@@ -214,3 +219,118 @@ class OrganisationGeographySubdivided(models.Model):
...
+        org_end_date = self.organisation.end_date
+
+        error_kwargs = {"message": {}, "code": "invalid_date"}
+
+        if self.change_type == OrganisationChangeType.END and (
+            org_end_date is None or org_end_date > ocl_effective_date
+        ):
+            error_kwargs["message"]["organisation"] = (
+                f"The Organisation's end date ({org_end_date}) must be before the OCL's effective_date {ocl_effective_date}"
+            )
```
chris48s, line 283: The error message doesn't match the check here.
If the check is right, the error text should be "on or before"
In reality though, I think actually the error message is what we want and the check is off by one
i.e: the effective date of the new legislation should be the start date of the new org, and the end date for the old one should be strictly before the effective date

every_election/apps/organisations/models/organisations.py
```diff
@@ -214,3 +219,118 @@ class OrganisationGeographySubdivided(models.Model):
...
+        on_delete=models.CASCADE,
+        limit_choices_to=~Q(organisation_type__in=["police-area", "europarl"]),
+    )
+    change_type = models.CharField(
+        max_length=10, choices=OrganisationChangeType.choices
+    )
+    information_url = models.URLField(
+        blank=True,
+        default="",
+        help_text="Link to council-specific information site",
```
chris48s: Organisation could be something other than a council (e.g: a Combined Authority). Can we generalise this and also explain when we want to use this as well as or instead of `OrganisationChangeLegislation.information_url`

every_election/apps/organisations/models/organisations.py
```diff
@@ -214,3 +219,118 @@ class OrganisationGeographySubdivided(models.Model):
...
+        through=OrganisationChange,
+        blank=True,
+    )
+    public_visibility = models.CharField(
+        choices=OCLPublicVisibility.choices, default=OCLPublicVisibility.HIDDEN
+    )
+    information_url = models.URLField(
+        blank=True,
+        default="",
+        help_text="Link to area-wide LGR information site",
```
chris48s: This is also a bit specific. For something like a Boundaries Amendment Order, "LGR" isn't really the right term.

every_election/apps/organisations/models/organisations.py
chris48s, line 313: I think it would be useful to include the text from your top post

> These include legislation like Structural Change Orders, or Boundary Amendment Orders -- anything that can:
> - Create or end organisations
> - Update organisation boundaries
>
> Some examples include:
> [The Surrey (Structural Changes) Order 2026 ](https://www.legislation.gov.uk/uksi/2026/264/made
) - Ends 12 orgs and create 2 new ones
> [The Glasgow City Council Area and North Lanarkshire Council Area (Cardowan by Stepps) Boundaries Amendment Order 2018](https://www.legislation.gov.uk/ssi/2018/308/made) - updates boundaries for 2 orgs
> [The Hampshire and the Solent Combined County Authority Regulations 2026](https://www.legislation.gov.uk/uksi/2026/595) - creates a new org

in a docstring for the class here.

every_election/apps/organisations/models/organisations.py
awdem: So the reason I added `name` was because I wanted to a name to use for cases where the legislation doesn't exist yet, so there is no `legislation_title`. 
 
 So for example, to create a record for the Devon LGR proposal, I set `name = "Devon LGR"`  and left `legislation_title` blank.
 
I did it this way because we want to be able tell people earlier than the actual written legislation that something is going on. `name` is a pretty vague, ironically. Maybe it could be called something else?

chris48s: OK. Its a bit weird that `name` is a required field and `legislation_title` is optional but `legislation_title` is the thing we really want, and we still have to populate `name` even when we have a `legislation_title`. I think I'd be inclined to:

- Just have a single name/title field
- Make it required
- Explain in the notes "if we don't have a title yet, use a provisional one"

or we could:

- Have 2 fields: `legislation_title` and `provisional_name`
- Make them both optional
- Have a constraint/validation that says one of them must be non-blank

Tbh, I'd be inclined to go with the first one. Once we have a legislation title, we don't really need the provisional name any more so we don't really lose anything by overwriting it.

awdem: Given that `legislation_title` will likely end up in the aggregator API, do you think it's ok to have provisional and real values in the same field or would that be confusing to an API consumer?

every_election/apps/organisations/models/organisations.py
awdem: Done in [ed5a641](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/ed5a64169197f8602531c18e9519de1030550559)

every_election/apps/organisations/models/organisations.py
awdem: I've elaborated in [093cd5a](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/093cd5aa53eea864e950cd0d7deb3fd295db7397)

every_election/apps/organisations/models/organisations.py
awdem: Tried to make it more generic in [093cd5a](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/093cd5aa53eea864e950cd0d7deb3fd295db7397)

every_election/apps/organisations/models/organisations.py
awdem: Added in [d56fb68](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/d56fb6807e677032845708ff3c526dcd5c0d9387)

every_election/apps/organisations/models/organisations.py
awdem: Yes you're right about the message being correct and the check being wrong. I've corrected it in: [fd46b18](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/fd46b1881b578caee89a92323d0d118d584e3fcc)

every_election/apps/organisations/models/organisations.py
awdem: Fixed in [000299f](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/000299f3df104c1f9806d3d1729e15897ffcc04f)

every_election/apps/organisations/tests/test_organisation_models.py
awdem: Fixed in same commit [as above](https://github.com/DemocracyClub/EveryElection/pull/2630#discussion_r4018698922): [000299f](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/000299f3df104c1f9806d3d1729e15897ffcc04f)

every_election/apps/organisations/models/organisations.py
chris48s: Yes. I see what you mean. Maybe option two then :+1:

every_election/apps/organisations/migrations/0076_organisationchange_organisationchangelegislation_and_more.py
```diff
@@ -123,8 +123,7 @@ class Migration(migrations.Migration):
                 ),
             ],
             options={
-                "get_latest_by": "modified",
-                "abstract": False,
+                "verbose_name_plural": "Organisation Change Legislation",
```
chris48s, line 129: Here we are losing properties because our `class Meta` is overriding `TimeStampedModel`'s rather than appending to it.

I think the solution to this is:

```
class Meta(TimeStampedModel.Meta):
```

I bet this bug already exists in other places in our code. Maybe lets not pull that thread now, but lets at least not introduce it here.

awdem: Fixed in [d561881](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/d561881ba94d3fd320420d6e9f76fce5863045e0)

every_election/apps/organisations/models/organisations.py
awdem: [001b87c](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/001b87cb02471700d87b7f31e5c08192e3cdfef3) renames `name` to `provisional_name`, makes it optional and adds a check constraint.

[7978e8e](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/7978e8e7e1562f702b0e0c78dc33ac85ba4f802e) renames it on the factory model

[2cff731](https://github.com/DemocracyClub/EveryElection/pull/2630/commits/2cff7317c7c3f726895b88bbf7152796ac88b35d) adds a test for the check constraint

every_election/apps/organisations/models/organisations.py
awdem: At least this can be changed without a data migration.

chris48s (approved):
I think this one is good to go. I'll leave it with you to squash the fixups and merge

PR #2631, opened 2026-09-14
organisation change legislation (2 of 2)

awdem:
This PR is part 2 of a stack

---

This PR adds a provisional flag to the `Organisation` model. An org marked as provisional is meant to represent an organisation that is proposed (for example, by an LGR proposal), but has not been created by legislation yet.

This means we can create provisional organisations in areas undergoing LGR (or any org change) and attach them to  `OrganisationChangeLegislation` objects so that we can eventually communicate about them downstream.

This PR also splits the `Organisation` model class manager into `private_objects` and `public_objects`, with the latter returning a queryset that excludes provisional orgs. We use this same manager pattern for  the `Election` model and it allows us to more explicit about where we are exposing provisional orgs. For the reasons explained [here](https://github.com/DemocracyClub/EveryElection/blob/8a267432a9bdb6466e5b53b284b74c4ebe900236/every_election/apps/elections/models.py#L434-L452), `private_objects` is the default manager. Ater updating all the existing organisation manager calls in [03f2d79](https://github.com/DemocracyClub/EveryElection/pull/2631/commits/03f2d79e1d6fe3e25920005a21ceba2aa85e13d3), I've made it so that provisional organisations are not exposed by the API or show up in the public frontend, and should only appear in admin.

Future Work:

- One thing I've found in testing this is that we don't have an easy way to import an arbitrary organisation geography into the database. If we want to attach geographies to provisional organisation, I think it would be useful to have a tool to do so. Our existing tooling is only for importing from sources like boundaryline or for unioning existing geographies together. I've made a ticket about improving tooling for org creation and noted this down.

([03f2d79](https://github.com/DemocracyClub/EveryElection/pull/2631/commits/03f2d79e1d6fe3e25920005a21ceba2aa85e13d3) is included as drive-by fix.)

---


every_election/apps/organisations/models/organisations.py
```diff
@@ -65,8 +88,11 @@ class Organisation(UpdateElectionsTimestampedModel, DateDisplayMixin):
     start_date = models.DateField(null=False)
     end_date = models.DateField(blank=True, null=True)
     legislation_url = models.CharField(blank=True, max_length=500, null=True)
+    provisional = models.BooleanField(default=False)
     ValidationError = ValueError
-    objects = OrganisationManager().as_manager()
+
+    private_objects = PrivateOrganisationManager()
```
chris48s, line 97: I don't think we need to duplicate the whole thing, but maybe link to
https://github.com/DemocracyClub/EveryElection/blob/8a267432a9bdb6466e5b53b284b74c4ebe900236/every_election/apps/elections/models.py#L434-L452
here.

chris48s:
Some things that don't really fit at a specific line on the diff:

---

There's 2 concepts we need to think about here.

1. Is the Organisation provisional
2. Is this [other type of object] related to a provisional Organisation

So for example, if/when we add a DivisionSet API endpoint, we would need to make sure to only expose DivisionSets related to non-provisional Organisations.
My instinct is that I can't think of anywhere we need to worry about this for now, but can you think of anything?

---

> One thing I've found in testing this is that we don't have an easy way to import an arbitrary organisation geography into the database. If we want to attach geographies to provisional organisation, I think it would be useful to have a tool to do so. Our existing tooling is only for importing from sources like boundaryline or for unioning existing geographies together. I've made a ticket about improving tooling for org creation and noted this down.

I agree, but I think its hard to do completely generically. Lets have a conversation about this one.

---

There's a few management commands where we are using `.public_objects`. The ones that instantly stand out to me are:

- `create_divisions_for_new_org`
- `import_divisionsets_from_csv`
- `import_lgbce`

where I can see that we might want to use/re-purpose these while an Org is still provisional and we're trying to cobble together boundaries. I'd be inclined to make these use `.private_objects`. We're not really trying to "hide" anything in this dev-facing context. Can you think of any others that are similar?

---

This code here could theoretically pick a provisional org for the example:

https://github.com/DemocracyClub/EveryElection/blob/8a267432a9bdb6466e5b53b284b74c4ebe900236/every_election/apps/elections/templates/elections/election_types.html#L15

We need to fix this in this PR

every_election/apps/organisations/models/organisations.py
```diff
@@ -65,8 +88,11 @@ class Organisation(UpdateElectionsTimestampedModel, DateDisplayMixin):
     start_date = models.DateField(null=False)
     end_date = models.DateField(blank=True, null=True)
     legislation_url = models.CharField(blank=True, max_length=500, null=True)
+    provisional = models.BooleanField(default=False)
```
chris48s, line 91: Do you think we need to make any more fields optional on the Org model to support this, or are we good as things stand?

every_election/apps/organisations/migrations/0078_alter_organisation_managers.py
```diff
@@ -14,7 +13,6 @@ class Migration(migrations.Migration):
         migrations.AlterModelManagers(
             name="organisation",
             managers=[
-                ("public_objects", django.db.models.manager.Manager()),
```
chris48s: I haven't really dug into this other than noticing the diff looks unexpected. What's going on with `every_election/apps/organisations/migrations/0078_alter_organisation_managers.py`
Why do we remove `public_objects` in 03f2d79e1 ?

awdem: Sorry I think I've made a mess of this migration when I was trying to clean some things up locally before creating the PR. I've also unintentionally changed all the commit hashes with a `gh stack rebase` to further complicate things :(.

The migration file that gets added in [afbcf4e](https://github.com/DemocracyClub/EveryElection/pull/2631/commits/afbcf4e819192df1de3c833c697f76614040d927) couldn't have actually been generated at that stage because `makemigrations` can't run until changes in [8a757d2](https://github.com/DemocracyClub/EveryElection/pull/2631/commits/8a757d24bdff4be5bee616de82fd763ae16b06c5) are applied ([8a757d2](https://github.com/DemocracyClub/EveryElection/pull/2631/commits/8a757d24bdff4be5bee616de82fd763ae16b06c5) was formerly [03f2d79](https://github.com/DemocracyClub/EveryElection/commit/03f2d79e1d6fe3e25920005a21ceba2aa85e13d3)). 

If I remove the migration when checking out 8a757d2 and then run `makemigrations`, the generated migration doesn't have `public_objects` in it - I think because the public manager does not have `use_in_migrations=True` on it. So I think the version of the migration in afbcf4e is out of date as well.

I think the best way to fix this is to squash 8a757d2 and afbcf4e together and re-generate the migration file at that stage during the rebase.

every_election/apps/organisations/models/organisations.py
awdem: I think we're good. The only required fields are `official_identifier` and `start_date`, which I think are ok to require for even a provisional organisation since we can update them when necessary and provisional orgs won't be exposed to anyone else unless:
- we turn off the provisional flag 
- we change the public visibility of any associated organisation change legislation to a public state

In both cases we're making a (hopefully) informed decision about what data we're showing to people.

every_election/apps/organisations/models/organisations.py
awdem: Done in [92cd4f9](https://github.com/DemocracyClub/EveryElection/pull/2631/commits/92cd4f9ce6fb333ff8eb2c71e64177ffd94d4916)

awdem:
> This code here could theoretically pick a provisional org for the example:
> 
> https://github.com/DemocracyClub/EveryElection/blob/8a267432a9bdb6466e5b53b284b74c4ebe900236/every_election/apps/elections/templates/elections/election_types.html#L15
> 
> We need to fix this in this PR

I've done this in [fac7363](https://github.com/DemocracyClub/EveryElection/pull/2631/commits/fac73638d3b4f52f2fbabc47e00d148262a8a6c8). I also chose to move the logic that constructs the example IDs to the view rather than leaving it on the template as that feels cleaner.

awdem:
> There's a few management commands where we are using `.public_objects`. The ones that instantly stand out to me are:
> 
> * `create_divisions_for_new_org`
> * `import_divisionsets_from_csv`
> * `import_lgbce`
> 
> where I can see that we might want to use/re-purpose these while an Org is still provisional and we're trying to cobble together boundaries. I'd be inclined to make these use `.private_objects`. We're not really trying to "hide" anything in this dev-facing context. Can you think of any others that are similar?

So I did think about this as I was going through [8a757d2](https://github.com/DemocracyClub/EveryElection/pull/2631/commits/8a757d24bdff4be5bee616de82fd763ae16b06c5), and I decided to not use `.private_objects`, because I thought that by the time we're using commands like bottom two you've listed, the organisation should no longer be provisional. And not being able to use these commands for a provisional org would remind to us unmark the field. 

But, actually, you're right about `create_divisions_for_new_org`, because it would allow us to create division sets for provisional orgs that are just existing orgs stitched together. And it looks like Will wrote it to create East and West Surrey. I'll change that one. (edit: done in [daf2427](https://github.com/DemocracyClub/EveryElection/pull/2631/commits/daf2427340dba6c0a1dfb2b447b348f507e5df5e))

I've had a look through the other commands that call `Organisation.public_objects` and I *think* that they're all correct in doing so.

awdem:
> So for example, if/when we add a DivisionSet API endpoint, we would need to make sure to only expose DivisionSets related to non-provisional Organisations. My instinct is that I can't think of anywhere we need to worry about this for now, but can you think of anything?

Yeah I don't think there is anywhere that could happen currently, but you're right we'll need to think about it for a DivisionSet API endpoint.

every_election/apps/elections/views/general.py
```diff
@@ -18,6 +18,29 @@ class ElectionTypesView(ListView):
...
+        return context
+
+    def build_example_id(self, obj):
+        id_elements = [obj.election_type]
+
+        if obj.subtype.exists():
+            ex_subtype = obj.subtype.first().election_subtype
+            id_elements.append(ex_subtype)
+
+        if obj.organisation_set.count() > 1:
```
chris48s:
```suggestion
        if obj.organisation_set.exclude(provisional=True).count() > 1:
```

Agree with moving this to be done at the view layer 👍

every_election/apps/organisations/migrations/0078_alter_organisation_managers.py
chris48s: Right, yes I see

chris48s (approved):
I reckon:
- Squash the fixups on https://github.com/DemocracyClub/EveryElection/pull/2630
- Merge that
- Rebase this one
- Tidy up the commit history on this one (taking into account the wierd migration thing - squash it to make sense)
- Then merge this one

PR #2634, opened 2026-09-18
Public election radar

symroe:
This is an iteration on a not very perfect feature. I think it's more useful than it was, but there's still some thought needed in future as to how we manage this,. 

---
  - https://app.asana.com/0/0/1216919182048824


every_election/apps/election_snooper/templates/election_snooper/snooped_election_list.html
chris48s, line 11: I think I'd be inclined to exclude this menu if you're logged out since we're not linking this on the main menu and we don't really use any of the statuses anymore

chris48s:
If we remove the `is_authenticated` check here

https://github.com/DemocracyClub/EveryElection/blob/0303821d39cdee49ee7f12765fa9267f4ca3f552/every_election/apps/elections/templates/id_creator/id_creator_base.html#L14

then you'll get the banner if you're logged out too.

